PR #50, opened 2014-03-11
WIP: constrained optimization

timholy:
This is not done yet, but I thought I should put it out there so people can comment sooner rather than later, if it looks like this is going in the wrong direction.

When finished, this should implement all the standard inequality constraints; equality constraints may come, but likely will be a little later (I have a pressing need for the inequalities, but not the equalities). The only algorithm I've implemented is a central-path-following interior-point solver---but I'd venture to say that if you could only pick one, interior-point is the way to go. 

Using this, a version of `nnls` (just two lines, `objective = linlsq(A, b); result = interior(objective, initial_x, bounds, method=:newton)`) is 100x faster on a 50x50 problem compared to the existing implementation of `nnls` in `Optim`.

Left to do:
- [ ] Add support for constraints to other linesearch algorithms
- [ ] Implement value, gradient, and hessian for barrier functions of linear and nonlinear constraints
- [ ] Add more tests

In addition to adding support for constrained optimization, this makes several relevant improvements to `hz_linesearch`, in particular ones that are smarter about some of the odd behaviors of floating-point numbers, and others that will make constrained optimization perform significantly better. Thanks to @JeffBezanson for his quick work on https://github.com/JuliaLang/julia/issues/6097, making it feasible to debug some of the darker corners of `hz_linesearch`.

One final point: note that this changes how the `problems` directory is handled. I changed it because it wasn't obvious to me that we wanted this to be a regular part of loading `Optim`. Perhaps we should go one step further, and only load it when running tests (and not even have `testpaths` in `Optim.jl`).


johnmyleswhite:
@timholy: Any interest in getting this branch ready to merge? It seems like the desire for consistency with JuMP had the unfortunate effect of delaying all progress on making your work widely available. If we're going to consider redesigning the outward facing API, we should make sure we have basic support for constrained optimization in place first.

src/interior.jl
```diff
@@ -0,0 +1,367 @@
...
+                                    0)
+        end
+        if gc == 0
+            t = one(T)  # FIXME: bounds constraints, starting at exact center. This is probably not the right guess.
+        else
+            t = gc/(convert(T,0.1)*go)
+        end
+    end
+    if method == :newton
+        return iterior_newton(objective, initial_x, constraints; t=t, mu=mu, eps_gap=eps_gap) # FIXME
```
johnmyleswhite: Typo here?

timholy:
In just the last two weeks, I've abandoned this in my lab's internal code and switched to MathProgBase. Not because I don't think this is promising, but because the effort to rewrite/finish this to make it consistent with the MathProgBase architecture is not small. Optimization is really important to me, but between my lab's internal needs and the big revamps occurring in the world of Images I just don't have time.

timholy:
I should add that I have no reservations about anyone grabbing this and running with it, if it seems like a useful starting point.

mlubin:
Which solver have you been using? Ipopt?

timholy:
Yes. (I also have a fix to Ipopt I need to make---it sometimes finds a solution that's worse than the original warmstart solution, and we should at least check for that. The fix is to twiddle the barrier-penalty options; the bottleneck is finding a **simple** test case that reproduces this behavior.)

tkelman:
Are you warmstarting all of primal, dual, and the barrier parameter? It can be tricky to entirely satisfy all of the convergence requirements of an IP solver when warmstarting (there's quite a bit of internal algorithmic state, in the filter line search and elsewhere), and you may end up converging to local optima fairly far from your initial point (even when that initial point is or is close to a better solution) in non-convex problems.

It would be good to continue brainstorming APIs that enable matrix-free operator approaches to large scale constrained optimization, but several of the key operations for that are already standardized in MathProgBase (just not used yet by any existing solvers, to my knowledge).

timholy:
Ipopt's `warmstart` only sets the primal parameters, as far as I can tell. And yes, this is a case of multiple minima, where I'm able to give a better initial guess than the minimum found by Ipopt after the barrier pushes it out of its starting basin.

timholy:
Speaking of API, I'll start one aspect of discussion over [here](https://github.com/JuliaOpt/MathProgBase.jl/issues/87).

tkelman:
This is off topic for this repo, but without duals there's no way for Ipopt to know if the point you gave it satisfies KKT conditions. Dual warmstart might not be exposed in one or more of Ipopt.jl, MathProgBase.jl, or JuMP, but it should be available in Ipopt itself (if a bit buried).

Tweaking options or doing extra checks should usually be in user code, not solver wrappers, unless there's an existing julia API for them.

mlubin:
Yes, dual warm starts are possible in Ipopt but not currently exposed.

timholy:
I'm back to being interested in this branch again. Reference: https://groups.google.com/forum/#!topic/julia-opt/TVmuXFWfeBM. In contrast, here I see no overhead from the solver, and the progress towards the minimum **per iteration** also seems better. I haven't yet measured it **per function evaluation**; that's likely to look a little less favorable, because I noticed that our linesearch not infrequently uses more evaluations than Ipopt. (There's some evidence that might be a [good thing (p. 2164)](http://epubs.siam.org/doi/abs/10.1137/120898097), but it seems likely to reduce the speed of optimization.)

I kinda wanted to get out of the business of writing my own optimization code, and maybe I still will if someone comes up with another solution. But I thought I'd at least ping this issue to say it may not be dead yet. Another thing about Optim that I like is that I can use `Float32`s (my 100GB disk file would be a 200GB disk file had I used `Float64`s), and I worry a little about the minimizer detecting "edges" among adjacent `Float32` values if it's using `Float64`s as coordinates for the optimization.

Since I've just been playing with this, here are a couple of random API thoughts:
- With barrier methods, you have a combined objective. With `show_trace=true`, it's nice to show just the part that corresponds to the user's objective. The architecture here doesn't yet make that easy, because what the "outer" optimizer sees is the total function penalty (user-objective + barrier terms).
- The API here isn't yet as flexible as MathProgBase. However, I note that it takes substantially fewer "user" lines of code. (Making it more flexible might reduce that advantage, of course.)

mlubin:
@timholy, I'm open to suggestions like https://github.com/JuliaOpt/MathProgBase.jl/issues/87 on how to make the MPB API easier to both use and implement on the solver side.

timholy:
Other than what I suggested already, I can't think of any other ways to improve on what you've done---at the end of the day you have to be able to interact with C code, and the API you've already designed is quite good for that. I just wanted to throw in a tweak or two.

But of course new ideas may crop up in the course of implementation.

src/cg.jl
```diff
@@ -89,7 +89,7 @@ macro cgtrace()
                 dt["g(x)"] = copy(gr)
                 dt["Current step size"] = alpha
             end
-            grnorm = norm(gr, Inf)
+            grnorm = norm(gr[:], Inf)
```
johnmyleswhite: Can you use `vecnorm(gr, Inf)` here?

timholy: Yeah.

Thanks for the comment!

timholy:
I am under the illusion that I'll soon have a few free days to work on Optim, and might want to put this into shape for merger. Any API thoughts about this PR would be welcome; I think I'll need to make quite a few changes.

mlubin:
@timholy, I'd say choose the API that seems most natural to you, so long as it's possible to write a lightweight wrapper to take in MathProgBase input as well. There's no need for the native API to the solver to be exactly MathProgBase.

timholy:
Aside from the integration into MathProgBase, what I wonder most about is the proper API for specifying constraints. Ipopt, for example, does not appear to have a special category for linear constraints, but I assume that would be a useful thing to specialize on. Likewise, https://github.com/JuliaOpt/Optim.jl/pull/50#issuecomment-37338779 suggests two possible APIs for specifying equality or inequality constraints. As I think about it, I lean towards making everything an equality constraint and then specifying bounds on the slack variables, but any comments would be welcome.

From the user's perspective, the good news is that no matter how crappy an API I design, it will still look pretty when accessed via the remarkable JuMP :smile:. But that's probably not something to shoot for.

tkelman:
If you want the API to be appropriate for problems that may have general nonlinear equality constraints, then my personal favorite canonical form (equalities and bounds) hasn't changed over the last 2 years. Conversions between canonical forms are simple so the user-facing API can use something more general like the row-wise form even if the algorithm internally uses something different. For other classes of problems you may want the internal implementation to use a different form (and/or API) though.

In the Ipopt algorithm, linear constraints can at best save you a subset of Jacobian row evaluations, and those are typically not the bottleneck relative to the solution of the KKT system for the Newton step. You can set an option flag if all constraints happen to be linear then the Jacobian will only be evaluated once. Whether it's worth the API complication of allowing situations in between, either via a linearity bitmask or separate inputs for linear vs nonlinear constraints (Matlab fmincon style), depends mostly on whether you expect Jacobian evaluations to be expensive.

timholy:
Don't delete this branch just yet, though. People may be using it.

pkofod:
Well, we could let it stay forever. It could be used as a "masterclass in git-fu"... I know I've tried to rebased this quite a few times, only to give up halfway through!

timholy:
Probably a good idea. Certainly people in my lab are using it!

PR #92, opened 2015-01-01
Fix warnings from deprecated syntax and functions for Julia 0.4

dankolbman:
[nan and inf deprecated](https://github.com/JuliaLang/julia/pull/8776)
[iceil merged into ceil](https://github.com/JuliaLang/julia/pull/9133)


dankolbman:
Reverted. Will make a new request with v0.3 compatability

PR #98, opened 2015-01-18
Don't add special behavior to Base.dot

### LINKED ISSUE #97: Overloading Base.dot
Optim.jl overloads the Base.dot function with this method

    # Dot product of two "vectors", even if they don't have a vector shape
    function dot(x::Array, y::Array)
       d = x[1] * y[1]
        for i = 2:length(x)
            d += x[i] * y[i]
        end
        return d
    end

Therefore when using Optim the following might happen

    julia> dot([1,2], [3 4 5])
    11

This is a behavior that I definitely don't expect and really don't want to have. Would it maybe be possible to not overload Base.dot but just use it internally in Optim?

johnmyleswhite:
Resolves #97

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,0 +1,16 @@
+# Dot product of two "vectors", even if they don't have a vector shape
+_dot(x::Vector, y::Vector) = dot(x, y)
+function _dot(x::Array, y::Array)
+    # TODO: Check length here?
+    if length(x) != length(y)
+        throw(DomainError())
+    end
+    d = x[1] * y[1]
+    for i = 2:length(x)
+        d += x[i] * y[i]
+    end
+    return d
+end
+
+# Vector-norm-squared, even if it doesn't have a vector shape
+norm2(x::Array) = _dot(x, x)
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -2,8 +2,7 @@ module Optim
     using Calculus
     using Compat
 
-    import Base.dot,
-           Base.length,
+    import Base.length,
            Base.push!,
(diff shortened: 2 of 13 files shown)


src/utils.jl
```diff
@@ -0,0 +1,16 @@
+# Dot product of two "vectors", even if they don't have a vector shape
+_dot(x::Vector, y::Vector) = dot(x, y)
```
andreasnoack: I think that both these methods could be covered by `_dot(x::Array, y::Array)=dot(vec(x),vec(y))`.

johnmyleswhite: That's true, but I'm not sure there's a reason to allocate a new Array object for such a simple function.

andreasnoack: Usually, `vec` doesn't allocate a new `Array`.

johnmyleswhite: Doesn't `vec` always allocate a new `jl_array` object, even though that object aliases the "same" memory?

andreasnoack: Okay I see. There is of course an allocation of a new `Array` *object*. For small arrays your implementation is therefore slightly faster, but the reason I suggested the `vec` based solution was to avoid code duplication.

mlubin: I agree with @johnmyleswhite that triggering a new allocation (even if it's small) for every call to ``dot`` should be avoided.

andreasnoack: I have no problem with that.

pkofod:
Is there any reason why this is not merged?

johnmyleswhite:
No, I just moved on to other things.

PR #99, opened 2015-02-04
fixed deprecation warnings thrown by julia 0.4.  Did NOT use the Compat ...

Cody-G:
...package to ensure compatibility with earlier Julia versions.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.2-
 Calculus
 DualNumbers
+Compat
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -1,6 +1,7 @@
 module Optim
     using Calculus
-
+    using Compat
+    
     import Base.dot,
            Base.length,
            Base.push!,
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -150,7 +150,7 @@ function cg{T}(df::Union(DifferentiableFunction,
     # Store f(x) in f_x
     f_x = df.fg!(x, gr)
     @assert typeof(f_x) == T
-    f_x_previous = nan(T)
+    f_x_previous = convert(T,NaN)
     f_calls, g_calls = f_calls + 1, g_calls + 1
     copy!(gr_previous, gr)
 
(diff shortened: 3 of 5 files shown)


johnmyleswhite:
Thanks for doing this, Cody. Let's plan to merge this when 0.4 becomes a release candidate.

mlubin:
What's the reason for not using Compat? We won't be able to merge this for months.

Cody-G:
I'm new to the scene (I'm a student in Tim's lab, that was my first pull
request ever).  So if I'm not following the standard procedure, please do
tell me.  I didn't use Compat because I wasn't sure whether the lines were
incompatible with previous versions of Julia or not, and I had imagined
that a lot of unnecessary @compat statements would clutter the code.  Is it
better practice to use @compat even when I'm not sure?  Thanks for your
patience.

mlubin:
@Cody-G, I apologize for the unduly alarmist tone. In a publicly released package, it's a good idea to keep the ``master`` branch (or equivalent) compatible with the latest julia release unless there's a good reason (like a new feature that's needed). Otherwise it makes it difficult to provide updates to users on julia 0.3 and 0.4 at the same time. So the ``@compat`` statements are a bit ugly, but they make it much easier to maintain the code.
The little red X next to your commit means that it failed the Travis CI tests, and you can see that it worked on 0.4 but not 0.3. If you open a new PR (or update this one) you can use this to double check that your changes haven't broken anything.

timholy:
Hey, Cody, thanks for doing this!

Others, note this is not against `master`, it's against `teh/constrained`. Still, in case anyone besides us is using this, it's probably best not to break 0.3 support. Fortunately, I think you'll only need a `@compat` in front of the Dict statements, I think the rest are valid syntax on 0.3 or are provided by Compat as plain functions.

Cody-G:
I appreciate the feedback, and I'm sorry for the delay.  I've added @compat statements where appropriate, but the Travid CI tests now fail becuase the Compat package was not found.  Is it necessary to change a .yml file to have Travis install package dependencies before testing?  I've never used Travis before.  Thanks again!

timholy:
Nice. I think all you'd need to do to fix this is to add `Compat` to Optim's REQUIRE file, and the package manager will install it automatically.

Cody-G:
Thanks, I squashed and pushed and it's passing now.  Travis is nice!

timholy:
Yep, keeps us from making all sorts of mistakes. It's amazing how many bugs only happen on someone else's machine :smile:.

PR #100, opened 2015-02-08
Improve error messages in hz_linesearch

benkuhn:
If an incorrect or absent gradient or Hessian is given, hz_linesearch
will sometimes error out with "assertion failed: dphia < 0." This commit
supplies a more informative error message suggesting that the user's
gradient or Hessian may be wrong.

### DIFF
--- a/src/linesearch/hz_linesearch.jl
+++ b/src/linesearch/hz_linesearch.jl
@@ -420,8 +420,8 @@ function secant2!{T}(df::Union(DifferentiableFunction,
     b = lsr.alpha[ib]
     dphia = lsr.slope[ia]
     dphib = lsr.slope[ib]
-    @assert dphia < 0
-    @assert dphib >= 0
+    @assert dphia < 0 "dphia = $dphia < 0; gradient or Hessian may be incorrect"
+    @assert dphib >= 0 "dphib = $dphib >= 0; gradient or Hessian may be incorrect"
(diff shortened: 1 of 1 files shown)


mlubin:
I'll let @johnmyleswhite comment, but may as well be more explicit and say something like "this error may indicate that user-provided derivatives are inaccurate".

johnmyleswhite:
If you can, you should figure out where the raw value of this is defined. I believe it'll be a line involving `dot`, which would confirm my claim that this error is an assertion that the search direction is not a direction of descent.

As for idiomatic Julia, I'd replacing `@assert` with an explicit `if` that writes out an error using `throw(Error("FOO"))`.

benkuhn:
@johnmyleswhite: You're right, e.g.: https://github.com/JuliaOpt/Optim.jl/blob/master/src/newton.jl#L91. I'll update the error messages.

johnmyleswhite:
This is a lot closer to ready. One last thing I'd like to do is replace string interpolation with calls to `@sprintf`.

benkuhn:
Will do. Out of curiosity, why is that preferable?

johnmyleswhite:
It offers better control over printing and does more of the tedious computation at compile-time.

benkuhn:
@johnmyleswhite Sorry for the delay. That better?

johnmyleswhite:
Yeah, that's good.

PR #101, opened 2015-02-08
Update unconstrained.jl

pkofod:
Himmelbrau:

Added Hessian.

Hosaki:

Fixed an error (it should be x[1] in the exponential function), and added gradient and Hessian. The Hessian has simplified xx and yy-partial derivaties, or it gets messy (rest of the document doesn't seem to have simplified expressions, so I tried to avoid it elsewhere).

### DIFF
--- a/src/linesearch/hz_linesearch.jl
+++ b/src/linesearch/hz_linesearch.jl
@@ -420,8 +420,11 @@ function secant2!{T}(df::Union(DifferentiableFunction,
     b = lsr.alpha[ib]
     dphia = lsr.slope[ia]
     dphib = lsr.slope[ib]
-    @assert dphia < 0
-    @assert dphib >= 0
+    if !(dphia < 0 && dphib >= 0)
+        throw(Error(string("Search direction is not a direction of descent; ",
+                           "this error may indicate that user-provided derivatives are inaccurate. ",
+                           @sprintf "(dphia = %f; dphib = %f)" dphia dphib)))
+    end
     c = secant(a, b, dphia, dphib)
     if display & SECANT2 > 0
         println("secant2: a = ", a, ", b = ", b, ", c = ", c)


andreasnoack:
Could you also add a few tests of these functions?

src/problems/unconstrained.jl
```diff
@@ -120,16 +122,29 @@ examples["Himmelbrau"] = OptimizationProblem("Himmelbrau",
 ##########################################################################
 
 function hosaki(x::Vector)
-    a = (1 - 8 * x[1] + 7 * x[1]^2 - (7 / 3) * x[1]^3 + (1 / 4) * x[1]^4)
-    return a * x[2]^2 * exp(-x[2])
+    a = (1.0 - 8.0 * x[1] + 7.0 * x[1]^2 - (7.0 / 3.0) * x[1]^3 + (1.0 / 4.0) * x[1]^4)
+    return a * x[2]^2 * exp(-x[1])
```
johnmyleswhite: This disagrees with Problem 20 at http://www.researchgate.net/profile/Montaz_Ali/publication/226654862_A_Numerical_Evaluation_of_Several_Stochastic_Algorithms_on_Selected_Continuous_Global_Optimization_Test_Problems/links/00b4952bef133a1a6b000000.pdf

Is there a typo in that reference?

johnmyleswhite:
Thanks for working on this!

src/problems/unconstrained.jl
pkofod: I think there is a typo in the reference in the code, and the reference you provided is correct. I will change it to x[2], and change the reference.

pkofod:
Not sure if I can reply to specific comments, but no prob JMW, and I will add tests Noack.

pkofod:
I think I have done something "un-githubby", since now it says it is failing. However, I have a local version that is working now (fixed some mistakes). Now I just need to make tests.

@johnmyleswhite is it as intended, that I get the following when starting Newton at the true values?

ERROR: assertion failed: dphia < 0
 in secant2! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:423
 in hz_linesearch! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:333
 in hz_linesearch! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:188
 in newton at /home/cce-pkm/.julia/v0.3/Optim/src/newton.jl:96
 in optimize at /home/cce-pkm/.julia/v0.3/Optim/src/optimize.jl:225

johnmyleswhite:
We should probably raise a better error, but starting Newton's method from a local minimum is dangerous business since the method has no reason to leave any minimum once it's found one.

pkofod:
I do realize that, but I might have expected an error along the lines of the starting point being a local optimum. I didn't actually want to start it at a minimum, but by pure chance I hit all three roots of the polynomial three times in a row, so I thought my code might be wrong. It would have been nice to know that they were potentially local optima.

I'll add some tests later, and finish some of the other examples in unconstrained.jl.

johnmyleswhite:
I think the best solution to that concern is to improve the error messages. That error message means, in English, search direction is not a direction of descent. In practice, that means either:

* You're at a local optimum
* You've made a mistake in your gradient

From empirical evidence, the odds that a user is at an optimum are very low (you're the first person to ever report that). The vast majority of users have buggy gradients.

mlubin:
What about checking for convergence before taking the first step?

timholy:
I can't imagine that starting at the minimum is an especially rare problem. Perhaps there should be a special check?

johnmyleswhite:
How do we define convergence given the possibility of false gradients? Always use finite differencing to determine "proper" gradients and then assert that we're at a local minimum if the norm of the "proper" gradient is approximately 0?

mlubin:
Well, garbage in garbage out. The chance that there's a bad gradient *and* the bad gradient is equal to zero at the starting point seems relatively small. Given the cost, I wouldn't do any derivative checking by default, though it would certainly be a useful option.

johnmyleswhite:
Ok, I can live with some false positive convergence claims.

pkofod:
I'm not sure if there is a "gradient_check"-function in Calculus, but I guess the warning could tell people to run a function "gradient_check" on their (f,g!), to see if it was a buggy gradient.
```julia
using Optim
using Calculus

function gradient_check (df::DifferentiableFunction, point)
	temp = zeros(size(point))
	df.g!(point, temp)
	@assert norm(temp - derivative(df.f, point)) < 10-3
end

example = Optim.UnconstrainedProblems.examples["Himmelblau"]
df=DifferentiableFunction(example.f,example.g!)
point = example.initial_x

gradient_check(df, point)
```

johnmyleswhite:
There is such a focus in Calculus, although the interface is a little wonky.

andreasnoack:
I think the changes look fine, so this should be finalized. A few comments though

 - you have somehow changed the permissions of all the files.
 - in a few places, you are using tabulator instead of four spaces.
 - the commits should be squashed into one unless you are fixing separate things
 - I think the default is not to use spaces around `^`, but I'm not sure if we are consistent on that

At this point it might be easier to copy your changes to some new files, reset this branch to master, paste your changes into the fresh version and commit that and push it to the remote branch. Fixing up a branch like this can be quite cumbersome.

pkofod:
Thank you for the comments. I will do the changes and commmit again.

pkofod:
I seem to have closed this involuntarily, and I can't reopen, since github tells me I have no commits in my branch (not true: https://github.com/JuliaOpt/Optim.jl/compare/master...pkofod:patch-1 ). Sorry for the mess, but am I supposed to make a new pull request? 
problem

joehuchette:
Looks like the git history got messed up, did you do a force push? It's probably easiest to open a new PR.

pkofod:
Yes I did. Problem was, that my branch had gotten messed up, so I did:

git fetch origin
git reset --hard origin/master (with my fork checked out)
copied over my changes from backup + committed
git push --force fork patch-1

I'll open a new PR and link to this one.

PR #104, opened 2015-03-04
Added problem hessians and tests

pkofod:
PR as a follow-up to https://github.com/JuliaOpt/Optim.jl/pull/101 where I messed up the git history

### DIFF
--- a/src/problems/unconstrained.jl
+++ b/src/problems/unconstrained.jl
@@ -1,5 +1,12 @@
 module UnconstrainedProblems
 
+### Sources
+### 
+### [1] Ali, Khompatraporn, & Zabinsky: A Numerical Evaluation of Several Stochastic Algorithms on Selected Continuous Global Optimization Test
+### Link: www.researchgate.net/profile/Montaz_Ali/publication/226654862_A_Numerical_Evaluation_of_Several_Stochastic_Algorithms_on_Selected_Continuous_Global_Optimization_Test_Problems/links/00b4952bef133a1a6b000000.pdf
+###
+### [2] Fletcher & Powell: A rapidly convergent descent method for minimization,
+
 immutable OptimizationProblem
     name::ASCIIString
     f::Function
@@ -47,7 +54,10 @@ examples["Exponential"] = OptimizationProblem("Exponential",
 ##########################################################################
 ###
 ### Fletcher-Powell
-###
+### 
+### From [2]
+### Source: A rapidly convergent descent method for minimization
+###         Fletcher & Powell
 ##########################################################################
 
 function fletcher_powell(x::Vector)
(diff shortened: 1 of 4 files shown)


andreasnoack:
Hm. For some reason your git is messing with the file permissions. I haven't seen that issue before, but have a look [here](http://stackoverflow.com/questions/1257592/how-do-i-remove-files-saying-old-mode-100755-new-mode-100644-from-unstaged-cha). Maybe that can fix it.

pkofod:
Can I ask where you can see that there is a problem? 1) so I can see the error message and 2) so I can see if it is fixed if I try fix it

andreasnoack:
If you take look here https://github.com/JuliaOpt/Optim.jl/pull/104/files then after each file name it says 100644 → 100755 which is showing the change in file permissions.

pkofod:
I have no idea why that happened. I just chmod'ed and committed. Now it should be fine, right?

johnmyleswhite:
Looks much better.

src/problems/unconstrained.jl
```diff
@@ -77,70 +87,75 @@ examples["Fletcher-Powell"] = OptimizationProblem("Fletcher-Powell",
                                                   fletcher_powell,
                                                   fletcher_powell_gradient!,
                                                   fletcher_powell_hessian!,
-                                                  [0.0, 0.0, 0.0],
-                                                  [[0.0, 0.0, 0.0]], # TODO: Fix
+                                                  [-1.0, 0.0, 0.0], # Same as in source
```
johnmyleswhite: So this works now with Newton's method?

src/problems/unconstrained.jl
```diff
@@ -77,70 +87,75 @@ examples["Fletcher-Powell"] = OptimizationProblem("Fletcher-Powell",
...
-                                             himmelbrau,
-                                             himmelbrau_gradient!,
-                                             himmelbrau_hessian!,
-                                             [0.0, 0.0],
-                                             [[1.0, 0.0]], # TODO: Fix
+examples["Himmelblau"] = OptimizationProblem("Himmelblau",
+                                             himmelblau,
+                                             himmelblau_gradient!,
+                                             himmelblau_hessian!,
+                                             [2.8, 1.8],
```
johnmyleswhite: Why start so close to the solution? Because you want to always get this solution?

johnmyleswhite:
Minor nit: we don't put spaces around `^`. Would be great to go back to that style.

src/problems/unconstrained.jl
pkofod: Well, yes. I guess I didn't really think of the tests as "benchmarks" as much as tests that the newton (for example) can actually accept the gradients and hessians and find a nearby local minimum. I guess we/I chould change it such that we just see if the algorithm converged to a point which is one of the local minima. I just put [3, 2] because it is the global solution.

johnmyleswhite: Ah. The original design of these problems was to make sure the algorithms would work as black boxes, since that's how users will make use of them. They're not ideal for that, so I'd understand changing that, but I saw it as a virtue that you could confirm convergence to some solution given a plausible black-box starting point.

src/problems/unconstrained.jl
pkofod: No, because the Newton tests are only run for the problems with twicedifferentiablefunction == true.

src/problems/unconstrained.jl
pkofod: Okay. Originally my PR was just to include additional gradients and hessians for the problem, and then I just added some tests to show (and document) that they "worked". The "tests" I added should probably be revised if they confuse more than they help.

johnmyleswhite: Well, what happens if you just start again from the old initial starting points?

pkofod: Then I get the following. I am not quite sure what the error is about, but it is a concave region, so I guess the newton step is in the wrong direction.

ERROR: Error not defined
 in secant2! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:143
 in hz_linesearch! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:333
 in hz_linesearch! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:188 (repeats 10 times)
 in newton at /home/cce-pkm/.julia/v0.3/Optim/src/newton.jl:96

pkofod:
I've changed the ^ locally, and they'll be fixed when I push the next time.

src/problems/unconstrained.jl
andreasnoack: I think it is due to this bug https://github.com/JuliaOpt/Optim.jl/blob/master/src/linesearch/hz_linesearch.jl#L424. Try replacing `throw(Error(string` with just `error`

pkofod: Do you mean replace throw(Error(string(.... ))) with error(...) ? Then it just gives me the thing below. 

ERROR: Error not defined
 in secant2! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:143
 in hz_linesearch! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:333
 in hz_linesearch! at /home/cce-pkm/.julia/v0.3/Optim/src/linesearch/hz_linesearch.jl:188 (repeats 10 times)
 in newton at /home/cce-pkm/.julia/v0.3/Optim/src/newton.jl:96

src/problems/unconstrained.jl
pkofod: Optim.nelder_mead finds the optimum, but I'm not sure what would happen with a method based on the gradient or Hessian. The gradient isn't really defined at the optimum I guess, since datan/dx is 1/(x^2/y + y), and the solution is x=(x1,x2,x3)=(1,0,0), and you would also have a 1/y'ish term in the hessian.

pkofod:
So I changed the ^-spacing, and I also changed the initial value from [2.8,1.8] to [2.0,1.5] just to start a further away from [3,2]. To be honest I don't think it makes sense to think too much about what the starting values should be in these test problems, until the problem with starting in concave regions is resolved. Whenever I seem to start at a concave region, I get errors. I wouldn't expect this to happen, though, but I have to read a bit more into how Optim's Newton-algorithm is globalized.

johnmyleswhite:
I don't think there's anything special done to globalize Newton's method.

pkofod:
Okay, so in that case I guess it actually _should_ throw an error if I start it in a concave region, right?

andreasnoack:
@johnmyleswhite Do you have further comment here? If not, I think it can be merged when the commits have been squashed.

johnmyleswhite:
No more comments. Merge ready now for sure.

pkofod:
Squashed and pushed... Let's see what Travis has to say!

andreasnoack:
@johnmyleswhite Can you push the bottom? I'm not authorized to that.

PR #111, opened 2015-04-11
Update hz_linesearch.jl

AndyGreenwell:
A call to error had a seemingly unintended capitalization.

### DIFF
--- a/src/linesearch/hz_linesearch.jl
+++ b/src/linesearch/hz_linesearch.jl
@@ -421,7 +421,7 @@ function secant2!{T}(df::Union(DifferentiableFunction,
     dphia = lsr.slope[ia]
     dphib = lsr.slope[ib]
     if !(dphia < 0 && dphib >= 0)
-        throw(Error(string("Search direction is not a direction of descent; ",
+        throw(error(string("Search direction is not a direction of descent; ",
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
Thanks for working on this.

On 0.3.2, the lowercase version works whereas the uppercase version fails. Has this changed over time?

andreasnoack:
I think this has been wrong for while, but that the code path hasn't been tested. In general, we have moved away from `error` so maybe it would be better to define an `Exception` for this. The string can also be simplified to avoid the `string` function and `@sprintf` macro by just writing the message as a single string with some `$` signs. Something like
```julia
throw(OptimException("Search direction is not a direction of descent; this error may indicate that user-provided derivatives are inaccurate. (dphia = $dphia; dphib = $dphib)"))
```

mlubin:
I'll merge this, no reason to leave things broken.

timholy:
This came about from a renaming (julia 0.2 days?) `Error` -> `ErrorException`. It should either be `throw(ErrorException(str))` or `error(str)`; `throw(error(str))` triggers the error before `throw` ever gets executed.

mlubin:
@timholy, ah, good point. Just fixed this.

PR #116, opened 2015-04-29
Suppress deprecation warnings on 0.4-dev

yuyichao:
This get rid of the annoy warning when importing `Optim.jl` on `0.4-dev`.


yuyichao:
```
yuyichao% julia -f -e 'using Optim'         
WARNING: [a] concatenation is deprecated; use [a;] instead
 in vect at ./abstractarray.jl:29
WARNING: [a] concatenation is deprecated; use [a;] instead
 in vect at ./abstractarray.jl:29

yuyichao% JULIA_LOAD_PATH=${PWD}/.. julia -f -e 'using Optim'
```

yuyichao:
Not sure what was causing the failing just now. Update the patch and it should now pass on both 0.3 and 0.4...

mlubin:
Thanks!

PR #127, opened 2015-07-02
bump minimum julia version requirement

mlubin:
I don't think we can claim to support 0.2 and prereleases at this point.


johnmyleswhite:
Thanks!

Issue #102, opened 2015-02-17
Addressing closures


yakir12:
Related to [#53](https://github.com/JuliaOpt/Optim.jl/issues/53), maybe an additional method for `optimize` that accepts `DataType` types will help with the closures-speed-bump?
We could then use `FastAnonymous` and custom made types.

mlubin:
Do you have a particular example when the overhead of using closures is significant?

yakir12:
In my case, I'm remapping RGB values using a relationship between semi-random noisy colormaps and some quanta (angle of polarization, degree of polarization, etc.). It means finding what, say, degree of polarization would result in an RGB color that best matches that of a certain pixel. I have 500x500 pixels per image, and a few tera images. I wouldn't bother if the improvement is less than 10 orders of magnitude though.

johnmyleswhite:
Have you tried making it clear to Julia that the closed over variables are constant when defining your closure? In my experience, that's usually sufficient to remove some of the worst performance costs of closures.

yakir12:
Here is the version of the code I assume would be the fastest:
```julia
function main(a)
	function fun(x)
		# some code that involves a and x
	end
	optimize(fun,-π/2,π/2).minimum
end
```
Where would I set any constants?
Or in a more closure-like syntax:
```julia
function fun(x,a)
	# some code
end
function main(a)
	optimize(x -> fun(x,a),-π/2,π/2).minimum
end
```
All these examples are very minimal, but I want to understand what you meant with the constants.

johnmyleswhite:
I believe something like this should get you the perf that you need:

```
function main(y)
    const a = y
    function fun(x)
        # some code that involves a and x
    end
    optimize(fun,-π/2,π/2).minimum
end
```

It's worth noting that closures aren't anonymous functions. Your first example is a named function that's also a closure. Your second example is an anonymous function that's also a closure. Being anonymous doesn't make your code more "closure-like".

yakir12:
Awesome, thanks! I'll confess that I thought that closure and anonymous are the same :P
 
Didn't do much though, here's a concrete example, `RGB` is a `Color.jl` type which is immutable (so `const` doesn't change much, right?):
```julia
using  Optim, Color
function fAoP(b::RGB)
	const a = b
	function f(x)
		x = 6.*x/π + 3. # 0 <= x <= 6
		x < 1 ? 
			1. - a.r + abs(a.g - x) + a.b :
		x < 2 ?
[...]

Issue #94, opened 2015-01-03
Extra arguments to objective/gradient/hessian functions


bdeonovic:
I didn't see any examples in the documentation that showed how you can optimize an objective function (with corresponding gradient and hessian) that depends on some variables you are not trying to optimize on. 

What is the appropriate way to handle such a situation?

mlubin:
#90 #85 #53
We definitely need an FAQ on this.

johnmyleswhite:
Agreed.

johnmyleswhite:
Subsuming in broader documentation issue.

Issue #53, opened 2014-04-11
Allow passing in non-optimizable parameters?


johnmyleswhite:
Lately I've found myself constructing some pretty complicated closures to work around the current Optim interface.

As such, I'd like to propose a revised protocol for functions passed to Optim. Instead of working with a function `f(x::Array) -> Float64`, I'd like to work with functions of the form `f(x::Array, extra::Any)` where `extra` can be used to provide any additional information needed to evaluate `f` given `x`. This would make several things simpler:
- Fitting statistical models that need to be conditioned on observed data
- Implementing alternating minimization
- Working with functions that need to update several buffers during evaluation

What would people think of making this kind of breaking change? It would make Optim much more general, but might also make the interface less intuitive for newcomers since you'd want to do something like `f(x, nothing)` when your function doesn't depend on additional information.

PR #128, opened 2015-07-08
abstract DifferentiableFunction

alyst:
This is related to #53, #94, #102 etc.
With the current implementation of ```DifferentiableFunction``` it's not very convenient to implement functions that require external data for evaluation. It always ends up writing something like
```julia
df = DifferentiableFunction(x -> real_func(x, data) .... )
```
before every optimizer invocation.

This PR turns ```DifferentiableFunction``` into an abstract type, and calls like ```df.f(x)``` are supposed to be replaced by ```evalf(df, x)```.
The old ```DifferentiableFunction``` is renamed into ```SimpleDifferentiableFunction```. Since ```SimpleDifferentiableFunction``` is used when one wants to convert 3 functions into ```DifferentiableFunction``` object, the current Optim code does not require any modifications.

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -144,7 +144,7 @@ function cg{T}(df::Union(DifferentiableFunction,
     y = similar(x)
 
     # Store f(x) in f_x
-    f_x = df.fg!(x, gr)
+    f_x = evalfg!(df, x, gr)
     @assert typeof(f_x) == T
     f_x_previous = convert(T, NaN)
     f_calls, g_calls = f_calls + 1, g_calls + 1
@@ -230,7 +230,7 @@ function cg{T}(df::Union(DifferentiableFunction,
         copy!(gr_previous, gr)
 
         # Update the function value and gradient
-        f_x_previous, f_x = f_x, df.fg!(x, gr)
+        f_x_previous, f_x = f_x, evalfg!(df, x, gr)
         f_calls, g_calls = f_calls + 1, g_calls + 1
 
         x_converged,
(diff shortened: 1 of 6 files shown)


johnmyleswhite:
I'm not really sold on this change. In general, I think we should exploit attempts to optimize closures in Base Julia rather than try to work around it.

I also would like to see us move away from the `DifferentiableFunction` type someday and use normal function types.

alyst:
It might be regarded as a closure performance workaround, but the intention is to allow callable objects be used with Optim package.
I thought subclassing ```DifferentiableFunction``` would be also a natural way to specify how the gradient/hessian should be calculated, i.e. Autodiff, FiniteDifference (subtype allows specifying whether we want left, center or right) etc.

johnmyleswhite:
I'm hoping to move in the opposite direction as we redesign the API, so closing out.

alyst:
Is there some document that outlines the new design of `Optim`?
I would like to understand better how it fits the use cases I have in mind.

johnmyleswhite:
In the long term, I'd like to move towards an API that takes in arbitrary callables and uses only the number of arguments and the types of the trailing arguments to do dispatch.

For a concrete example, something like `optimize(f, x0::Vector, ::Type{LBFGS})` vs `optimize(f, g!, x0::Vector, ::Type{LBFGS})` would be used instead of the differentiable function machinery. We would have implicit contracts on `f` and `g!` rather than trying to force them via `DifferentiableFunction`.

We also adopt a more complex API like that used by NLopt. For now, I mostly want to remove the gap between the `optimize` wrapper and the inner methods that get called because the `optimize` wrapper gets tasked with too much moving around of arguments.

timholy:
> In the long term, I'd like to move towards an API that takes in arbitrary callables and uses only the number of arguments and the types of the trailing arguments to do dispatch.

I think this PR implements the very API you're talking about. `evalfg!(df, x, gr)` is a much more general interface (one that relies on dispatch) than `df.fg!(x, gr)`. #163 provides another example of where you need to be able to pass in (and dispatch on) something that's not one of the "core" numerical types.

Reopen?

PR #129, opened 2015-08-08
Allow precompilation

johnmyleswhite:
Title says it all.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6521" && __precompile__(true)
+
 module Optim
     using Calculus
     using Compat


src/Optim.jl
```diff
@@ -1,4 +1,6 @@
 module Optim
+    VERSION >= v"0.4.0-dev+6521" && __precompile__(true)
```
mlubin: Shouldn't this be outside ``module``?

johnmyleswhite: I think you're right. Since this passed tests I didn't think about the fact that this is a file level property.

johnmyleswhite:
As compilation is a file-level property, this was moved outside of the `module`.

pkofod:
Any idea about the speed up of `using Optim`?

johnmyleswhite:
It's there. The package loads fast enough for me now.

Issue #130, opened 2015-08-17
Is there interest in callback functions?


Evizero:
Hi! Awesome package! 

I wonder, was there a conscious decision to not support callback functions (performance reasons maybe)? If not, would there be interest? I know that there is the trace/history option and it might probably not be the most practical thing to do. However, I would really like them for what I want to implement for interface consistency among reasons 

I would of course be willing to spend the time to figure out a good way to allow for them (i.e. do the whole coding). I figured I ask first before coding away. Maybe that idea came up before and was rejected for some good reason that simpleton me is oblivious to

mlubin:
What kind of callbacks are you looking for? Just informational callbacks or also callbacks to interact with the solution method (e.g., early termination)?

Evizero:
Informational at the moment, but both in the long run. For example I want to be able to do a live plotting of the learning every k iterations. Also, if external code is given control I would be able to "step" through the optimization for demonstrations (to students). But It would be great if I could do early stopping after computing the validation error in the callback.

Evizero:
Not I am not asking for anyone to spend time on this. I really just want to get a feel of what people are interested here. I have no issue with doing it myself and refactor my changes according to feedback. I just want to avoid doing work where it was clear that its not of interest to the package maintainers

mlubin:
No objections here

Evizero:
I think I can do it without breaking the interface using multiple dispatch. Essentially I would make the first parameter the callback and create a default callback for the current trace option. This would allow me to do the following additionally to the current interface

```Julia
res = optimize(f, [0.0, 0.0], break_every=100) do status
  # do stuff with status
end
```

mlubin:
There are already multiple methods with a varying number of functions as inputs, not sure we can tack on another one without ambiguity. Why not a keyword argument to pass in the callback?

Evizero:
Now I see what you mean in optimize.jl. My bad. I was looking at the algorithms themselves which just have the `DifferentiableFunction` and such. silly me. Yes keyword argument then

PR #131, opened 2015-08-18
implemented callback functions

Evizero:
This is the pull request discussed in #130. 

It allows for having callback functions without breaking the existing interface

First of all it introduced the `show_every` param (defaults to 1) which controls for how regularly traces are shown and callbacks are called.

`show_trace` is unaffected except the `show_every` influence and one detail: If a callback is given then setting `extended_trace` to true will not automatically set `show_trace` to true as well. This is such that you can have a callback that has access to the extended stuff without having to have the console spam of `show_trace`. Using the interface as it was before this PR yields same results as before.

Callbacks are implemented for all models but `fminbox` and `levenberg_marquar` since they are special in the show_trace regard

Here is an example for using callbacks
```Julia
function cb(os::OptimizationState)
    # do something with state
end
res = optimize(f, [0.0, 0.0], method = :gradient_descent, callback=cb, show_every=50)
```

Or even with the history
```Julia
function cb(tr::OptimizationTrace)
    # do something with trace
end
res = optimize(f, [0.0, 0.0], method = :gradient_descent, callback=cb, show_every=50, store_trace=true)
```

I added tests as well. Tested on 0.3.11 and 0.4.0-dev+6698

Let me know what you think

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -12,7 +12,9 @@ module Optim
 
     export optimize,
            DifferentiableFunction,
-           TwiceDifferentiableFunction
+           TwiceDifferentiableFunction,
+           OptimizationState,
+           OptimizationTrace
 
     # Utils
     include("utils.jl")
--- a/src/accelerated_gradient_descent.jl
+++ b/src/accelerated_gradient_descent.jl
@@ -21,7 +21,9 @@ macro agdtrace()
                     grnorm,
                     dt,
                     store_trace,
-                    show_trace)
+                    show_trace,
+                    show_every,
+                    callback)
         end
     end
 end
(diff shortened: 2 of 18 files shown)


mlubin:
Is the ``show_every`` necessary? You should be able to just check the iteration count inside the callback.

Evizero:
well necessary is a strong word, I guess. However, for example in combination with the standard show_trace it has real performance differences (because printing is expensive). I find that I don't often need to know every iteration, especially if there are a couple of thousands in total.

Evizero:
Maybe I got carried away with the `show_every`. I guess I'm just used to such a parameter that I didn't question it to much. I can change it if you feel it doesn't fit in your design

mlubin:
I'm not too hostile to the idea. Would help to get some feedback from other current and potential users.

Evizero:
Can we please reach a decision on this? I am at a point where I have to decide what back-end to use

mlubin:
I'm okay with this as is. @timholy or @johnmyleswhite, could you also sign off?

timholy:
Just skimmed it very quickly (~90secs), didn't see anything to worry about and I trust your judgment (and haven't had time for Optim so don't feel I have a big stake).

mlubin:
Thanks @Evizero! Sorry for the delay in merging

Evizero:
thanks a lot

PR #134, opened 2015-08-24
Make show MultivariateOptimizationResult end with newline

KristofferC:
Just fixing a minor annoyance.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6521" && __precompile__(true)
+
 module Optim
     using Calculus
     using Compat


johnmyleswhite:
Don't most show methods intentionally avoid the trailing newline to prevent weird printing issues in the REPL?

KristofferC:
Maybe, but no matter there is an inconsistency because the univariate does print a new line: https://github.com/KristofferC/Optim.jl/blob/f88f4445e17be164eb32c8c0306d987688ac6876/src/types.jl#L131

johnmyleswhite:
It looks like the univariate result `show` method is wrong if you take the usage of `show` in Base as precedent. The REPL adds an extra newline to the output of show:

```
julia> show(STDOUT, x -> x + 1)
(anonymous function)
julia> x -> x + 1
(anonymous function)

julia>
```

PR #136, opened 2015-09-15
changes in levenberg_marquardt

matthieugomez:
Three changes
- sse is replaced by sumabs2
- sqrt(lambda) is replaced by lambda^2. I think it's a typo in the original implementation.
- lambda is diminished only if rho is close to one. This seems to be the standard for levenberg_marquardt

### DIFF
--- a/src/levenberg_marquardt.jl
+++ b/src/levenberg_marquardt.jl
@@ -1,14 +1,6 @@
-# sse(x) gives the L2 norm of x
-sse(x) = (x'*x)[1]
-
-function levenberg_marquardt(f::Function, g::Function, x0;
-                             tolX=1e-8,
-                             tolG=1e-12,
-                             maxIter=100,
-                             lambda=100.0,
-                             show_trace=false,
-                             callback = nothing,
-                             show_every = 1)
+function levenberg_marquardt(f::Function, g::Function, x0; 
+	tolX::Real = 1e-8, tolG::Real = 1e-12, maxIter::Integer = 100,
+	lambda::Real = 10.0, show_trace::Bool = false)
 	# finds argmin sum(f(x).^2) using the Levenberg-Marquardt algorithm
 	#          x
 	# The function f should take an input vector of length n and return an output vector of length m
@@ -29,7 +21,9 @@ function levenberg_marquardt(f::Function, g::Function, x0;
 	const MAX_LAMBDA = 1e16 # minimum trust region radius
 	const MIN_LAMBDA = 1e-16 # maximum trust region radius
 	const MIN_STEP_QUALITY = 1e-3
+	const GOOD_STEP_QUALITY = 0.75
 	const MIN_DIAGONAL = 1e-6 # lower bound on values of diagonal matrix used to regularize the trust region step
+
 
 	converged = false
 	x_converged = false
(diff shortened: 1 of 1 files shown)


src/levenberg_marquardt.jl
```diff
@@ -56,14 +55,19 @@ function levenberg_marquardt(f::Function, g::Function, x0; tolX=1e-8, tolG=1e-12
...
 		# It is additionally useful to bound the elements of DtD below to help
 		# prevent "parameter evaporation".
-		DtD = diagm(Float64[max(x, MIN_DIAGONAL) for x in sum(J.^2,1)])
-		delta_x = ( J'*J + sqrt(lambda)*DtD ) \ ( -J'*fcur )
+		DtD = sumabs2(J, 1)
+		for i in 1:length(DtD)
+			if DtD[i] < MIN_DIAGONAL
+				DtD[i] = MIN_DIAGONAL
+			end
+		end
```
blakejohnson: This block seems equivalent to the previous code on line 63. Is this faster, other than the substitution of `sumabs2` for `sum(J.^2,1)`?

matthieugomez: the existing code creates two vectors while this new version creates one vector (ie the max operation is done "in place"). But clearly it's not a bottleneck.

blakejohnson:
Looking at this again, I can't remember why I had `sqrt(lambda)`, but I think it should probably be just `lambda` with a default initial value of 10.

blakejohnson:
Well, looking again I think I got the `sqrt` by looking at what Ceres does. But, Ceres also increases/decreases lambda by factors of 2 rather than factors of 10. No `sqrt` and no `^2` is closer to netlib, though, so we should probably stick with that.

matthieugomez:
Ok I'll change it. What about the good enough step = 0.75?

matthieugomez:
My reference for both the lambda2 and the GOOD_STEP_QUALITY is *An Inexact Levenberg Marquardt Method for Large Sparse Nonlinear Least Squares* by SJ Wright and J.N. Holt

blakejohnson:
I also see the GOOD_STEP_QUALITY thing in Ceres and Netlib. I'll check out
your reference but it seems to disagree with at least 5 other references I
looked at last night.

matthieugomez:
I've just updated the pull request with your proposed change for lambda. It'd be great to add a reference to the Netlib code. Is it MINPACK?

blakejohnson:
I was looking at MINPACK, though admittedly it is rather difficult to unearth the structure from the algorithm in all that fortran code. Perhaps some easier to read sources:
http://mathworld.wolfram.com/Levenberg-MarquardtMethod.html
https://www.cs.nyu.edu/~roweis/notes/lm.pdf
http://people.duke.edu/~hpgavin/ce281/lm.pdf
https://en.wikipedia.org/wiki/Levenberg%E2%80%93Marquardt_algorithm

blakejohnson:
@matthieugomez Can you also change the default value of `lambda` to 10 for consistency with the prior code?

matthieugomez:
done. (when changing to f! and g! it would be good to also change tolX tolG and maxIter to xtol grtol and iterations).

blakejohnson:
Great, LGTM.

johnmyleswhite:
Can't be merged without a rebase it seems

matthieugomez:
sorry! rebased.

johnmyleswhite:
No worries. Thanks for doing this!

PR #139, opened 2015-10-09
Print something readable for problems with more than two parameters

johnmyleswhite:
tl;dr: This makes printing not suck when you have a large parameter space.

### DIFF
--- a/test/types.jl
+++ b/test/types.jl
@@ -1,0 +1,28 @@
+module TestTypes
+    using Base.Test
+    using Compat
+    using Optim
+
+    prob = Optim.UnconstrainedProblems.examples["Rosenbrock"]
+    f_prob = prob.f
+    res = Optim.nelder_mead(f_prob, prob.initial_x)
+
+    io = IOBuffer()
+    show(io, res)
+    s = takebuf_string(io)
+
+    lines = split(s, '\n')
+    @test lines[1] == "Results of Optimization Algorithm"
+    @test startswith(lines[2], " * Algorithm: ")
+    @test startswith(lines[3], " * Starting Point: ")
+    @test startswith(lines[4], " * Minimum: [")
+    @test startswith(lines[5], " * Value of Function at Minimum: ")
+    @test startswith(lines[6], " * Iterations: ")
+    @test startswith(lines[7], " * Convergence: ")
+    @test startswith(lines[8], "   * |x - x'| < ")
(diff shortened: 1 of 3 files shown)


mlubin:
Looks like a good improvement. Could you add some tests for this? I don't think the printing code has any coverage.

johnmyleswhite:
Sure, I can do that. I'm generally pretty cautious about adding rich tests for show methods because they're so brittle and tedious. I'll probably just check that there's a line for each expected entry in the method.

mlubin:
Works for me, thanks!

pkofod:
Any chance we can address minimum vs minimizer in the output as per #76 ?

johnmyleswhite:
Sure. I'd also be very open to a PR that adds a real and documented API. I keep dreaming making a better Optim someday with a better API, but we could make substantial progress if someone puts in a few days of work.

pkofod:
Yeah, I agree. A lot of great work has already been done, but Optim is still very unpolished. At some point I'd be more than willing to donate a some days worth of work - but as for everyone else time is not in surplus here at the moment. We have no active issue/wishlish to discuss the overall future of Optim, right?

johnmyleswhite:
Not to my knowledge. But Optim is actually remarkably reliable given how easily it could have bugs and how many people seem to use it. I'm always surprised that Optim works as well as it does.

My personal priorities would be:

* A real API based on functions, not accessing fields of returned objects
* Real documentation -- both on the web and in the REPL
* Better correctness tests
* Better perf tests

The really ambitious project would be a redesign of the API to handle some kinds of constrained optimization problems. I would also like to rethink the assumptions we make about the functions we take in as arguments. But I think we should avoid going down that path until other stuff has been done.

pkofod:
I agree that it is quite reliable, but re:unpolished I was thinking about more superficial things. As you mentioned, stuff like functions to extract information, documentation, etc. I agree with your list (order as well), although perf tests are so difficult to "do right". We should be able to show some minimal results on performance though.

I'll see if I can throw some time at this when I get home from Canada.

johnmyleswhite:
Added a test for the new `show` method. Once tests pass, I'll merge.

test/types.jl
```diff
@@ -0,0 +1,27 @@
+module TestTypes
+    using Base.Test
+    using Optim
```
tkelman: using `Compat`. or drop 0.3 if you're going to do that anyway

johnmyleswhite:
Just threw that in after seeing the failures.

PR #141, opened 2015-10-11
Removed _dot and norm2, as they are not (or versions of them) in Base

pkofod:
As in commit message, I've removed _dot and norm2, as they are now in Base (as vecdot and vecnorm), and norm2 is not even used .


mlubin:
Fails on 0.3. You can bump REQUIRE and update the travis script to run on 0.4 and nightly only.

pkofod:
I bumped to v0.4, but I might as well remove the unnecessary @compat macros in this PR as well then. I'll do it later.

tkelman:
If this gets tagged, it should be as 0.5.0.

pkofod:
I forgot to include a file in the last commit, and I also fixed a typo in the commit message. @tkelman ok.

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.4
 Calculus
 DualNumbers
 Compat 0.7.1
```
mlubin: You could remove the dependency on ``Compat`` while you're at it

johnmyleswhite:
I'm good with this. If we break things for 0.4, it would be good to make some of the other breaking changes at the same time.

johnmyleswhite:
I tagged a release that still works on 0.3, so let's get to breaking stuff

johnmyleswhite:
Sigh. This needs a rebase. Sorry for that.

pkofod:
No problem, I'll fix it asap. Makes sense to tag a "last" v0.3 compatible version before merging this.

johnmyleswhite:
Yeah, that's what I did.

pkofod:
Yes I know, my English was probably a bit unclear. Maybe I should have written:
"Tagging a "last" v0.3 compatible version before merging this was smart."

:)

johnmyleswhite:
Sorry for the confusion. I should have thought more carefully about what you said.

Issue #147, opened 2015-10-22
Trace output errors


slundberg:
It seems like show_trace is reporting the estimated gradient as the function value and not showing the function value at all. See the below example on master:

```julia
julia> optimize(x->norm(x+4)+4, [0.0,0.0], show_trace=true, iterations=20)
Iter     Function value   Gradient norm 
     0     3.517950e-01              NaN
     1     3.187842e-01              NaN
     2     6.838181e-01              NaN
     3     6.390792e-01              NaN
     4     1.360921e+00              NaN
     5     1.246015e+00              NaN
     6     2.464085e-01              NaN
     7     7.868933e-02              NaN
     8     3.687705e-01              NaN
     9     3.532858e-01              NaN
    10     1.366622e-01              NaN
    11     1.043260e-01              NaN
    12     4.241014e-02              NaN
    13     8.865064e-02              NaN
    14     8.766231e-02              NaN
    15     4.231915e-02              NaN
    16     1.743947e-02              NaN
    17     1.577730e-02              NaN
    18     1.009202e-02              NaN
    19     5.752459e-03              NaN
    20     3.368566e-03              NaN
Results of Optimization Algorithm
 * Algorithm: Nelder-Mead
 * Starting Point: [0.0,0.0]
 * Minimum: [-4.005367537339528,-3.998781104882558]
 * Value of Function at Minimum: 4.005504
 * Iterations: 20
 * Convergence: false
   * |x - x'| < NaN: false
   * |f(x) - f(x')| / |f(x)| < 1.0e-08: false
   * |g(x)| < NaN: false
   * Exceeded Maximum Number of Iterations: true
 * Objective Function Calls: 38
 * Gradient Call: 0
```

pkofod:
I will add, that you cannot use `extended_trace=true` with nelder mead for this reason. We probably have to look at the internals of `NelderMead()` to fix syntax between this and other methods.

```julia
# This _won't_ work 
optimize(f, [0., 0.];method = :nelder_mead, store_trace=true, extended_trace=true)
```

slundberg:
Okay. If there is some small change to the PR that would be helpful let me know. Otherwise we can just wait for a more thorough change.

pkofod:
fixed by #204

PR #148, opened 2015-10-22
Fix the trace printing (#147)

slundberg:
f_x is not the objective value in nelder_mead but closer to the
gradient. So this reports a sensible objective value and uses f_x for
the gradient printing.

### DIFF
--- a/src/nelder_mead.jl
+++ b/src/nelder_mead.jl
@@ -30,8 +30,8 @@ macro nmtrace()
             grnorm = NaN
             update!(tr,
                     iteration,
-                    f_x,
-                    grnorm,
+                    Base.minimum(y), # best point in the simplex
+                    f_x, # f_x is used more like a grnorm for nelder_mead
                     dt,
                     store_trace,
                     show_trace,


slundberg:
Addresses #147

pkofod:
I can see from #147 that something is indeed off. The last function value does not correspond to the output. When you say "closer to the gradient" what do you mean by that? I think we should be careful not to say that something is a gradient norm if it isn't.

slundberg:
nelder_mead is a simplex based method and does compute an explicit gradient. But it does compute a variance of the points in the simplex (which is similar in spirit to a gradient). We could not show the gradient norm, or change the label, but changing the label for nelder_mead might require more code changes.

pkofod:
@slundberg I am closing this, see #204 . I've essentially done what you proposed, but I also have some additional changes regarding the printing and extended trace. I hope you find the other PR to successfully address you concern. Thanks for the effort!

slundberg:
great, thanks for your work on this!

Issue #145, opened 2015-10-13
Tentative roadmap


johnmyleswhite:
- API cleanup
  - Formal API For interacting with optimization results by merging a successor to #87
  - Replace method symbols with abstract types to exploit type dispatch and simplify exposing custom keyword arguments for different methods
- Documentation
  - Document a formal API
  - Provide examples of basic problems one might want to solve
- Benchmarks
  - For every problem we test against, we should log the number of iterations, the number of function calls and the number of gradient calls. Shrinking any of these numbers is the best way to improve perf.

pkofod:
Should API discussions happen at #87  ?

johnmyleswhite:
Probably, but I'm not worried if discussion is split a little bit.

pkofod:
Constrained optimization should probably be added as well, even if @timholy might not have the time right now (or someone else).

Evizero:
I'll probably have some time in about a week or two to work on some aspects of this. If no one else is already doing it, that is.

> Replace method symbols with abstract types to exploit type dispatch and simplify exposing custom keyword arguments for different methods

I am assuming you mean adding a positional argument to `optimize` that defines the algorithm to use, right (e.g. something like `BFGS()`)? A lot of the verbose code in optimize.jl could be avoided that way. Plus it would be easier to extend the functionality.

> For every problem we test against, we should log the number of iterations, the number of function calls and the number of gradient calls

I do something along the lines for KSVM.jl to [compare the results of my implementation against scikit learn](https://github.com/Evizero/KSVM.jl/blob/master/test/tst_automated.jl) that I stored in [csv files](https://github.com/Evizero/KSVM.jl/tree/master/test/reference). Does this go in the direction you're thinking of?

---

One reason I would like to work on this is because I would kinda like to sneak-in early stopping to the callback functions. Does this at all fit into your design goals?

tkelman:
* think about removing `nnls` if https://github.com/ahwillia/NonNegLeastSquares.jl is strictly better?

davidanthoff:
Not sure this is within the scope for anything near term, but long term it would be really neat if the algorithms in Optim were exposed as MathProgBase solvers, and if the user facing API in Optim would just be a user friendly layer over MathProgBase.

mlubin:
[...]

PR #149, opened 2015-10-24
[WIP] Replace method symbols with abstract types

Evizero:
As discussed in #145 . 

- [x] replace method symbols with abstract types to exploit type dispatch and simplify exposing custom keyword arguments for different methods
- [x] remove nnls
- [x] clean up tests to use types instead of symbols to specify the method
- [x] deprecate old standalone methods (e.g. Optim.l_bfgs)
- [x] add tests for deprecated methods (?)
- [x] use hashtable for method symbols
- [x] adapt readme to new API (thanks to @pkofod)
- [x] review and then resolve merge-conflicts with master

~~This works right now although the tests still need some clean-up.~~

I thought this would be a good point to gather feedback


mlubin:
I like it. Should we deprecate choosing the algorithms via symbols?

gabrielgellner:
Super sweet! Should `getOptimizer` be `get_optimizer`? I thought camel case was a no no in Julia.

Evizero:
> Should we deprecate choosing the algorithms via symbols?

I don't know. I simply assumed so for some reason. Personally, I'm neutral to it.

> Should `getOptimizer` be `get_optimizer`? I thought camel case was a no no in Julia.

yes, good catch! old habits...

src/accelerated_gradient_descent.jl
```diff
@@ -28,18 +28,22 @@ macro agdtrace()
...
+                     ftol::Real = 1e-8,
+                     grtol::Real = 1e-8,
+                     iterations::Integer = 1_000,
+                     store_trace::Bool = false,
+                     show_trace::Bool = false,
+                     extended_trace::Bool = false,
+                     callback = nothing,
+                     show_every = 1,
+                     linesearch!::Function = hz_linesearch!,
+                     nargs...)
```
johnmyleswhite: What's this? Let's avoid passing around things that aren't strictly necessary.

src/fminbox.jl
```diff
@@ -171,7 +176,7 @@ function fminbox{T<:AbstractFloat}(df::DifferentiableFunction,
             println("#### Calling optimizer with mu = ", mu, " ####")
         end
         pcp = (P, x) -> precondprep(P, x, l, u, mu)
-        resultsnew = optimizer(dfbox, x; xtol=xtol, ftol=ftol, grtol=grtol, iterations=iterations,
+        resultsnew = optimize(dfbox, x; method=optimizer, xtol=xtol, ftol=ftol, grtol=grtol, iterations=iterations,
```
johnmyleswhite: Should this be `optimize(dfbox, s, optimizer)` to avoid unnecessary runtime dispatch?

src/optimize.jl
```diff
@@ -1,17 +1,78 @@
+function getOptimizer(method::Optimizer)
+    method
+end
+
+function getOptimizer(method::Symbol)
```
johnmyleswhite: Probably easier to use a hash table here.

src/accelerated_gradient_descent.jl
Evizero: what do you mean?

johnmyleswhite:
This looks really great. Two concerns:

* Passing around `nargs` is somewhat un-Julian. We shouldn't propagate things that aren't amenable to human-level static analysis.
* We should deprecate the old functions that are being removed, like `Optim.l_bfgs`. Some people are almost certainly calling those functions directly.

src/accelerated_gradient_descent.jl
mlubin: What's the purpose of ``nargs...``?

src/optimize.jl
```diff
@@ -1,17 +1,78 @@
...
+    end
+    warn("Specifying the method using symbols is deprecated. Use \"method = $(T)()\" instead")
+    T()
+end
+
+function optimize(f::Function,
+                  initial_x::Array,
+                  method::Optimizer;
+                  autodiff::Bool = false,
+                  nargs...)
```
Evizero: I pass around `nargs...` to not break this design choice using `autodiff` as a KW args. This is because now this method is a generic fall-back.

src/fminbox.jl
Evizero: good catch

src/optimize.jl
Evizero: what's the clean way to do this? I thought having global variables messes with pre-compilation.

Evizero:
> Passing around nargs is somewhat un-Julian

See my last comment. I'm not sure how to solve this without breaking the `autodiff` KW, but I'll look into it. (If you have an idea I'm all ears)

> We should deprecate the old functions that are being removed

ok will do

Evizero:
A question about `nargs...`: Do you want to get rid of those in general? even in `optimize.jl` where they greatly reduce redundant code? If so I will. But I got to say that this would make the whole code a lot uglier and almost all default values meaningless as they would always be overwritten by the initial method. Seems like bad coding practice to me

johnmyleswhite:
Yes, we can't merge code that uses `nargs`. It makes local reasoning excessively difficult because it pushes around an unknown number of arguments of unknown types.

When I have some time free, I will figure out how to avoid the use of `nargs`. Ideally we'll move towards a shim-layer API that translates keyword arguments into a type that has fields for all options. You can see what NLopt does for an example: https://github.com/JuliaOpt/NLopt.jl/blob/523987afbb32105ed78dad0a127298284113147d/src/NLoptSolverInterface.jl#L4

Evizero:
> When I have some time free, I will figure out how to avoid the use of nargs. Ideally we'll move towards a shim-layer API that translates keyword arguments into a type that has fields for all options

This makes sense to me. I'll give it a shot.

Evizero:
Let me know what you think of this. Basically I put the general optimization options in `OptimizationOptions` and the special arguments (such as the history size of LBFGS) into the Optimizer type itself

mlubin:
:+1: 
Embedding the options in the algorithm types themselves is essentially what we also do in MathProgBase.

mlubin:
@Evizero, any plans to finish this up? It would be a very welcome contribution.

Evizero:
I got sidetracked because the approach of optimizing Julia-functions directly turned out to be somewhat inconvenient for what I was working on at the time.

Be that as it may, I do think the main bulk of work is done already. So if you're still interested in this, then I'll see to it that I finish this up within the next couple weeks. I'm currently drowning in work for an overdue project, but I do try to get a little Julia coding done each day.

mlubin:
Definitely interested, I promise a quick review on my part

Evizero:
Some progress. Still need to add the old methods back and deprecate them (e.g. `Optim.l_bfgs(...)`)

Evizero:
BTW I wasn't able to rebase because of conflicting patches. So now `nnls` is reintroduced. If I remember it correctly I originally deleted nnls because I read somewhere that this was on the agenda anyway.

Should `nnls` stay or go?

johnmyleswhite:
Either is good by me if it helps you.

Evizero:
So I guess all that is left is add tests to check if the deprecated methods still work. where should these tests be put? in a separate file or to the corresponding method?

pkofod:
I'd say throw them in the "test" folder, but others may disagree; @mlubin @johnmyleswhite ?

mlubin:
``test/deprecated.jl``?

pkofod:
@Evizero do you have a plan for what to do with the readme?

Evizero:
I completely forgot about the readme. I'll adapt it as well

pkofod:
@Evizero Okay. I'm going to have a go at it in relation to #165, but if you could change it so the examples work, that would be great :)

pkofod:
What are the tests for the deprecated methods supposed to do? See that the API change has not messed anything up, or?

Evizero:
I will finish this up this weekend. I'm sorry that this endeavour took me so long. Life was very "unexpected" the past few months. Bear with me a few more days.

@pkofod was nice enough to help out with the readme adaption

Evizero:
@pkofod they will simply test that the old way to use to methods still works

pkofod:
Okay, so basically just copy it from the old tests I guess?

Evizero:
yes more or less. But as @mlubin suggested they will be put into `test/deprecated.jl`.

FYI: I undid the premature merge with master (to clean up the mess of a history that I made here) and force pushed this branch

pkofod:
I see I did a few of those,  will fix, sorry.

Evizero:
Ready for the final review round

README.md
```diff
@@ -52,36 +30,55 @@ optimize(f, [0.0, 0.0])
 Optim will default to using the Nelder-Mead method in this case. We can specify the Nelder-Mead method explicitly using the `method` keyword:
```
KristofferC: `method` is no longer used as a keyword in the example here and below.

README.md
```diff
@@ -115,24 +112,36 @@ end
...
-* `:simulated_annealing`
+Box constrained minimization:
+* `Fminbox()`
 
-In addition to the `method` keyword, you can alter the behavior of the Optim package by using the following keywords:
+Special methods for univariate optimization:
+* `Brent()`
+* `GoldenSection()`
 
+In addition to the `method` keyword, you can alter the behavior of the Optim package by using the following keywords in a `OptimizationOptions` constructor:
```
mlubin: Another reference to the ``method`` keyword here

pkofod:
`method` keyword references are addressed in https://github.com/Evizero/Optim.jl/pull/4 - better fix my own mistakes.

Evizero:
As it stand now the old way to call `optimize` is still valid aside from specifying the `method` using an `Optimizer` instead of a symbol. The idea behind it was for convenience and backwards compatibility.

Do we want to deprecate the method keyword and just have the new way to call optimize?

pkofod:
I think it makes sense to keep the `method` keyword only for symbols (deprecated) and then only allow the new `Optimizer` syntax without `method` kw.

Evizero:
How about now? Deprecating the `method` KW has the consequence that with the new API the user can no longer (i.e. gets deprecation warning) the various KW arguments (e.g. `show_trace`) to `optimize` directly, but has to pass it using the `OptimizationOptions` type

mlubin:
Why do we need to deprecate the other keyword arguments? It seems generally useful to use them instead of an explicit ``OptimizationOptions`` object.

Evizero:
> Why do we need to deprecate the other keyword arguments?

Basically because deprecating the `method` kw results in ambiguity then for calling `optimize`.

 > It seems generally useful to use them instead of an explicit OptimizationOptions object.

The point of the `OptimizationOptions` is to reduce the mess of defining the same default values in a 100 places. My initial intend was that the user only has to deal with `OptimizationOptions` if he cares about it, while still being able to call the `optimize` method as before with the only difference that the KW `method` should now be an `Optimizer` instead of a symbol. Forcing the method to instead always be a positional argument instead of a KW messes this up a bit.

Now that I think of it: What is the idea behind deprecating the `method` KW if the other KW should  remain?

mlubin:
> What is the idea behind deprecating the method KW if the other KW should remain?

Type stability perhaps?

If getting rid of the keyword arguments lets us greatly simplify the code, then that's a good argument for doing so.

pkofod:
My comment above was mainly motivated by the fact, that my understanding was, that consensus was to deprecate `method = ...`. If it messes with the possibility of people writing `optimize(f, x, optimizer; xtol = 1)` then maybe we should just leave it. I can see the point that if you just want to tweak one option, you may not want to write `OptimizationOptions(xtol=1)`.

On they other hand I rarely use optimization such that the time to write `OptimizationOptions` is a significant fraction of the total optimization time, so personally I don't have many feelings towards "keep all keywords" or "remove all keywords".

Evizero:
I may undo my last commit then, which means that as a high-level interface the user can still use the convenient keywords, while as a low-level interface the user could also use the positional and typed arguments directly.

That would imply that the `method` keyword is back on the table

Evizero:
basically now we are at this point that the code below shows (excerpt from the readme):

> Thus, one might construct a complex call to `optimize` like:

```jl
res = optimize(f, g!,
               [0.0, 0.0],
               GradientDescent(),
               OptimizationOptions(grtol = 1e-12,
                                   iterations = 10,
                                   store_trace = true,
                                   show_trace = false))
```

> It is also possible to call a more convenient version of the method that uses keyword arguments directly:

```jl
res = optimize(f, g!,
               [0.0, 0.0],
               method = GradientDescent(),
               grtol = 1e-12,
               iterations = 10,
               store_trace = true,
               show_trace = false)
```

pkofod:
Looks good to me.

Evizero:
All right then! Tomorrow morning I will go over the code again to see if I missed something (like a test here or there) and then I'll resolve the merge conflicts with master so we can merge this!

pkofod:
Just a quick question, do we need to export `Optimizer`? I'm not saying it's wrong, I'm just trying to understand what the benefit is.

Evizero:
no, good point. I guess there is no benefit for the end user in having `Optimizer` in the namespace. I will take a close look at the exports tomorrow as well.

Evizero:
So, I think this should be ok now as far as the API is concerned. I undid some commits here and there, since the `method` keyword is back into play.

The travis error on 0.5 looks unrelated to this patch.

test/callbacks.jl
```diff
@@ -1,6 +1,14 @@
 
-for method in (:nelder_mead,
-               :simulated_annealing)
+function cb(tr::OptimizationTrace)
```
KristofferC: This function and the one below should be removed. They got accidentally added in your rebase.

After removing those, tests pass on 0.5.

Evizero: Oh, I see. Well If I remove those then there is no test anymore that checks if the callback got called

Maybe an anonymous function will do the trick

KristofferC: I am only talking about the two most upper ones. They seem completely unused since they get overwritten immediately.

Evizero: You are right, I saw them too and removed them.

pkofod:
Essentially, lgtm. Looking down over the `README.md`again. Now, we still allow `method` to be specified, but I'm not sure there's a reason to introduce the usage of `optimize` as if `method` is required. On the other hand, the documentation might need a larger overhaul soon, so this can be merged as is (imho) as the syntax still works.

pkofod:
@johnmyleswhite sorry to ping, but do you have any comments ? Besides a pre-merge squash maybe.

johnmyleswhite:
I will review this tonight and make sure this gets merged unless there turns out to be an issue I don't expect -- which I think is extremely unlikely.

README.md
```diff
@@ -52,36 +30,55 @@ optimize(f, [0.0, 0.0])
 Optim will default to using the Nelder-Mead method in this case. We can specify the Nelder-Mead method explicitly using the `method` keyword:
 
 ```jl
-optimize(f, [0.0, 0.0], method = :nelder_mead)
+optimize(f, [0.0, 0.0], method = NelderMead())
```
johnmyleswhite: Should we encourage people to use keywords in the docs? I would think one of the main gains of the switch to types is the chance for completely static dispatch, which I would assume want to promote, although that's obviously a complex issue.

README.md
```diff
@@ -52,36 +30,55 @@ optimize(f, [0.0, 0.0])
...
 
 The `method` keyword also allows us to specify other methods as well. Below, we use L-BFGS, a quasi-Newton method that requires a gradient. If we pass `f` alone, Optim will construct an approximate gradient for us using central finite differencing:
 
 ```jl
-optimize(f, [0.0, 0.0], method = :l_bfgs)
+optimize(f, [0.0, 0.0], method = LBFGS())
 ```
 
-For greater precision, you should pass in the exact gradient function, `g!`:
+For greater precision, you should pass in the exact gradient function. This can be shown to be:
```
johnmyleswhite: Maybe "For better performance and greater precision"? In my experience, the analytic gradient is less valuable for precision and much more valuable for the often enormous performance gains it brings with it.

README.md
```diff
@@ -115,24 +112,36 @@ end
...
+* `ConjugateGradient()`
+* `GradientDescent()`
+* `MomentumGradientDescent()`
+* `AcceleratedGradientDescent()`
 
+Requires a function, a gradient, and a hessian (cannot be omitted):
+* `Newton()`
+
+Box constrained minimization:
+* `Fminbox()`
```
johnmyleswhite: I'll leave this up to @timholy, but maybe this should be `FMinBox`? Absent him saying something in one or the other, I think it's ok.

README.md
```diff
@@ -140,39 +149,53 @@ In addition to the `method` keyword, you can alter the behavior of the Optim pac
...
                [0.0, 0.0],
-               method = :gradient_descent,
+               method = GradientDescent(),
                grtol = 1e-12,
                iterations = 10,
                store_trace = true,
                show_trace = false)
 ```
 
+It is also possible to call the type-stable low-level interface directly using `OptimizationOptions`:
```
johnmyleswhite: Maybe refer to this as statically dispatched instead of type-stable?

johnmyleswhite:
Made some comments that could be up for discussion, but I'm happy with this and think it's ready to merge unless anyone wants to make changes.

It's an impressively thorough body of high-quality work. Well done, @Evizero.

README.md
pkofod: I agree that it should go away most places, and it is my fault it is still in there. As mentioned above, in my latest comment, I figured it could get changed when some of the other API changes are made.

johnmyleswhite: Works for me.

pkofod:
Okay, so I created https://github.com/Evizero/Optim.jl/pull/5 . It is up to you @Evizero if you want to merge it. If you'd rather not, I'll just commit it after this merges.

I agree this is solid work, and I really think it moves the package forwards.

README.md
pkofod: I didn't address this in https://github.com/Evizero/Optim.jl/pull/5 as I just used the github web interface, but you're probably right. It makes sense given the other method names.

Evizero:
Thanks for the review and kind comments. I am glad that I could contribute something useful and will hopefully find myself in this position again.

KristofferC:
:tada: :balloon:

mlubin:
Thanks @Evizero, this is the biggest change to Optim in quite a while.
Would someone like to summarize the changes in a post to the julia-opt list?

pkofod:
By the way, regarding the lowercase letters in `Fminbox`. What is one of the nice things in this PR is that if we want it to be `FMinBox` at some point, we can now deprecate `Fminbox` which was _not_ possible with `:fminbox`. Very handy for potential breaking changes going forward.

PR #154, opened 2015-12-05
fix ftol convergence criterion in l_bfgs_b

maximerischard:
I think this is what was intended.

### DIFF
--- a/src/l_bfgs.jl
+++ b/src/l_bfgs.jl
@@ -193,7 +193,8 @@ function l_bfgs{T}(d::Union{DifferentiableFunction,
         copy!(gr_previous, gr)
 
         # Update the function value and gradient
-        f_x_previous, f_x = NaN, d.fg!(x, gr)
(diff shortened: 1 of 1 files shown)


mlubin:
@johnmyleswhite, any comments? Seems like you wrote the original code.

KristofferC:
This looks like a copy paste error from: https://github.com/JuliaOpt/Optim.jl/blob/d9f959480b8ced00572f3140dd85822e74fc0d03/src/l_bfgs.jl#L127

Looking at the bfgs version here: https://github.com/JuliaOpt/Optim.jl/blob/d9f959480b8ced00572f3140dd85822e74fc0d03/src/bfgs.jl#L142 we can see it corresponds to the PR here.

I believe this should be merged.

johnmyleswhite:
Thanks for merging this. It was definitely a copy-paste error as @KristofferC suggested.

PR #155, opened 2015-12-25
allow optimize() to override default neighbor!() for simulated_annealing

cstook:
Suggest adding neighbor! keyword parameter to optimize() and passing through to simulated_annealing() to allow overriding default_neighbor!

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -216,7 +216,8 @@ function optimize(f::Function,
                   callback = nothing,
                   show_every = 1,
                   linesearch!::Function = hz_linesearch!,
-                  bfgs_initial_invH = nothing)
+                  bfgs_initial_invH = nothing,
+                  neighbor!::Function = default_neighbor!)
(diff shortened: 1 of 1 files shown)


mlubin:
The infrastructure for passing options is changing with https://github.com/JuliaOpt/Optim.jl/pull/149, so we may want to wait for that to land first.

cstook:
OK, sounds good to me.

I'm learning julia.  Looking forward to reading the new code.

pkofod:
@cstook I think it's almost there, so if you want to know how the changes are going to materialize, you can read the PR now. Great to see a PR which adds useful functionality :+1:

Evizero:
Missed this issue before. #149  takes care of what you ask. See https://github.com/Evizero/Optim.jl/blob/chris/src/simulated_annealing.jl#L40-L43

pkofod:
In that case, I'll close this PR. Thanks for the effort @cstook , but it looks like it's about to be implemented in #149 .

PR #164, opened 2016-01-30
fixes for 0.5 function changes

mlubin:
This gets tests passing on 0.5.

### DIFF
--- a/test/brent.jl
+++ b/test/brent.jl
@@ -1,6 +1,6 @@
-f(x) = 2x^2+3x+1
+f_b(x) = 2x^2+3x+1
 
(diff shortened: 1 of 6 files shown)


pkofod:
Sorry for my ignorance, but aren't these just name changes?

KristofferC:
https://github.com/JuliaLang/julia/pull/14759#issue-128035834

> A few packages are failing because their tests repeatedly write f(x) = ..., intending to create a different function each time.

pkofod:
Thanks @KristofferC , seems like a very sensible change indeed!

Issue #140, opened 2015-10-10
Naming conventions and APIs - breaking changes?


pkofod:
As discussed in #139 we might want to make a little push on some of the features of Optim. One such thing is a proper API for accessing convergence information and optimization output in general. @johnmyleswhite already has a branch with a lot of functions added.

One thing that has been brought up (in #76 for example) is the improper use of the term minimum in this package. If we do go ahead and change that, what would be the best way to do it? Some peoples code might depend directly on the field name (which is exactly why we need a proper API so we can use the deprecation tools in the future for such changes). I guess we simply make a breaking change? Or are there deprecation tools for this sort of thing? We could just copy the contents of the .minimizer field to a .minimum, such that old programs still work, but this seems awkward.

Edit: another way is of course to add only the functions for extracting information in the first round, and the change the internal names when people have had a chance to change their code to using the new function based API.

mlubin:
Unfortunately there's no way to deprecate field access. My vote is to just break things (with notice to julia-opt list and documentation in the README) if we have a significant improvement to make. The package isn't mature enough to be held back by legacy code.

johnmyleswhite:
I'm ok with breaking things, although it seems better to me to start by creating a formal API before breaking the informal API (i.e. the direct field access strategy).

johnmyleswhite:
How willing are we to break things? I think we could simplify the code considerably if we made the optimization methods into abstract types and used type dispatch to replace the big manual keyword dispatch that's happening in `optimize`.

mlubin:
No objections here

pkofod:
Fine with me

Ken-B:
With another package I'm also replacing manual dispatch to type dispatch, but I wonder whether it's best to use abstract types for this or type instances? 

The manual seems to [suggest](http://docs.julialang.org/en/release-0.4/manual/style-guide/#avoid-confusion-about-whether-something-is-an-instance-or-a-type) type instances. Any reason you'd prefer an abstract type for each optimization method instead of a type instance?

Apologies if I seem to be hijacking the issue, but I thought this comment might be useful here before starting the new implementation.

oxinabox:
GadFly uses both type instances and abstract types, where the type instances (when instantiated) hold the special method parameters. (sometimes)

pkofod:
The questions have been answered.

PR #165, opened 2016-02-01
Getting the API train running

### LINKED ISSUE #76: minimum versus minimizer

pkofod:
So I figured that these changes / additions were way over due, and starting a PR would be a good motivational factor.

Most of it is based of off @johnmyleswhite 's branch.

- Renamed `it` to `iterations`
- Added `iteration_converged` to UnivariateOptimizationResults
- Added API methods for accessing fields of OptimizationResults

I have a small though, and you may not agree with it. Is there a reason why we define the "exceeded iterations" convergence criterion opposite of the others, such that "false" is good and "true" is bad? 

Please add comments or ideas about improvements to the API. 

Fixes #76 

**Update**
Since #149 was merged, it's probably worth it to continue with this. Please comment and discuss.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -33,6 +33,9 @@ module Optim
     # Types
     include("types.jl")
 
+    # API
+    include("api.jl")
+
     # Automatic differentiation utilities
     include("autodiff.jl")
 
--- a/src/api.jl
+++ b/src/api.jl
@@ -1,0 +1,69 @@
+method(r::OptimizationResults) = r.method
+minimizer(r::OptimizationResults) = r.minimum
+minimum(r::OptimizationResults) = r.f_minimum
+iterations(r::OptimizationResults) = r.iterations
+iteration_limit_reached(r::OptimizationResults) = r.iteration_converged
+trace(r::OptimizationResults) = length(r.trace.states) > 0 ? r.trace : error("No trace in optimization results. To get a trace, run optimize() with store_trace = true.")
+
+function x_trace(r::UnivariateOptimizationResults)
+	tr = trace(r)
+	!haskey(tr.states[1].metadata, "x_minimum") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
+    [state.metadata["x_minimum"] for state in tr.states]
+end
+function x_lower_trace(r::UnivariateOptimizationResults)
+	tr = trace(r)
+	!haskey(tr.states[1].metadata, "x_lower") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
+    [state.metadata["x_lower"] for state in tr.states]
+end
+x_lower_trace(r::MultivariateOptimizationResults) = error("x_lower_trace is not implemented for $(method(r)).")
+function x_upper_trace(r::UnivariateOptimizationResults)
+	tr = trace(r)
+	!haskey(tr.states[1].metadata, "x_upper") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
+    [state.metadata["x_upper"] for state in tr.states]
+end
+x_upper_trace(r::MultivariateOptimizationResults) = error("x_upper_trace is not implemented for $(method(r)).")
+
+function x_trace(r::MultivariateOptimizationResults)
+	tr = trace(r)
+	!haskey(tr.states[1].metadata, "x") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
+    [state.metadata["x"] for state in tr.states]
+end
+
+f_trace(r::OptimizationResults) = [state.value for state in trace(r).states]
+g_norm_trace(r::OptimizationResults) = error("g_norm_trace is not implemented for $(method(r)).")
+g_norm_trace(r::MultivariateOptimizationResults) = [state.gradnorm for state in trace(r).states]
+
+f_calls(r::OptimizationResults) = r.f_calls
+
+g_calls(r::OptimizationResults) = error("g_calls is not implemented for $(method(r)).")
+g_calls(r::MultivariateOptimizationResults) = r.g_calls
+
+converged(r::UnivariateOptimizationResults) = r.converged
+converged(r::MultivariateOptimizationResults) = r.x_converged || r.f_converged || r.gr_converged
+x_converged(r::OptimizationResults) = error("x_converged is not implemented for $(method(r)).")
+x_converged(r::MultivariateOptimizationResults) = r.x_converged
+f_converged(r::OptimizationResults) = error("f_converged is not implemented for $(method(r)).")
+f_converged(r::MultivariateOptimizationResults) = r.f_converged
+g_converged(r::OptimizationResults) = error("g_converged is not implemented for $(method(r)).")
+g_converged(r::MultivariateOptimizationResults) = r.gr_converged
+
+x_tol(r::OptimizationResults) = error("x_tol is not implemented for $(method(r)).")
+x_tol(r::MultivariateOptimizationResults) = r.xtol
+f_tol(r::OptimizationResults) = error("f_tol is not implemented for $(method(r)).")
+f_tol(r::MultivariateOptimizationResults) = r.ftol
+g_tol(r::OptimizationResults) = error("g_tol is not implemented for $(method(r)).")
+g_tol(r::MultivariateOptimizationResults) = r.grtol
+
+
+initial_state(r::OptimizationResults) = error("initial_state is not implemented for $(method(r)).")
+initial_state(r::MultivariateOptimizationResults) = r.initial_x
+
+lower_bound(r::OptimizationResults) = error("lower_bound is not implemented for $(method(r)).")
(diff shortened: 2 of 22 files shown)


mlubin:
This is great! I'd like to get https://github.com/JuliaOpt/Optim.jl/pull/149 merged first though. ``iterations_converged`` is a bad name. It should be ``stopped_by_iteration_limit`` or something similarly explicit.

andreasnoack:
It would be great to have #76 fixed soon and maybe here.

pkofod:
@mlubin Yeah, it seemed counter intuitive to me. I'm fine with merging #149 first, I just wanted to take a first iteration on the API.

@andreasnoack That is "fixed" here using the API. As you can see, `minimizer` extracts the `minimum` field, and `minimum` extracts the `f_minimum` field. I think it would be a good idea to introduce the API in this PR. When people have had a fair warning, and have hopefully changed their `res.minimum` to `minimizer(res)`, then we can easily change it "behind the scenes" in a new PR. I think this was the consensus in #140 .

pkofod:
Any great ideas for changing `iterations_converged`? Should we go with @mlubin 's `stopped_by_iteration_limit`? It's a bit long, but it's hard to convey the message with fewer word - at least I don't have a great improvement in mind.

Other comments are also highly appreciated.

KristofferC:
`iteration_limit_reached`?

pkofod:
I like it, :+1:

Evizero:
> `iteration_limit_reached`

:+1: best suggestion yet

pkofod:
I'm just waiting for the `Optimizer` PR to merge before I continue, but how do people feel about having multiple functions for the same purpose? Right now I took `history` from the `api` branch, but since it is called a trace internally, and most people probably know it conceptually as a trace, I have added an identical function `trace`. Personally I am okay with multiple functions doing the same things (as long as the functionality is only coded in one of them, and the others are just "aliases"), if it means more people get what they expect when they try `history` or `trace` for example. However, I would love feedback on this.

Also, I intend to add stuff like `x_trace` and `f_trace` functions. If people object to the names, I will adjust to the popular opinion :)

KristofferC:
Personally I don't like having multiple functions that does the same thing. It is extra code, extra documentation and creates inconsistencies in user code / examples written by other people for i.e. courses etc. 

A related thing I am a bit worried about is the generality of the function names we export. 

For example with this branch we already have:
```jl
julia> using Optim

julia> using NLsolve
WARNING: using NLsolve.converged in module Main conflicts with an existing identifier.
```

With time I am getting more and more wary about dumping too much exports on the user. Not saying this should hold anything up, just something to keep in mind.

pkofod:
In that case I guess I prefer `trace`, but again, I don't have many feelings invested, and we can use `history` if people prefer that.

Regarding the `conflicts with an existing indentifier`-thing. Yes that is true, and kind of a general issue for Julia packages. Would you rather see the API in the form of `Optim.converged`, `Optim.trace`, or do you mean something like `optim_converged` etc? I'm not sure how else to handle this problem without doing something to how base Julia handles exports.

pkofod:
Choo choo, this train is running again as #149 is merged. Currently, I've added some functions, but both comments and opinions on names and extent of this PR are welcome. Also updated the first post.

johnmyleswhite:
Will review this tonight. My one change I'll suggest in advance is that we shouldn't try to export so many methods. I'm generally moving away from the idea that libraries should export almost anything and my usage of Optim in practice suggests that we really don't need to export anything more than `optimize`, if that. Unlike something like Distributions.jl, I think most use cases for Optim will involve very few of the names from the Optim namespace.

pkofod:
Okay, so that is basically what @KristofferC was also worried about. I won't change anything right now, but what you propose is that people use `Optim.trace(result)`, correct?

johnmyleswhite:
Exactly.

pkofod:
Okay, I can see the problem in exporting a million functions with generic names, and will adjust in next commit.

src/Optim.jl
```diff
@@ -28,11 +28,39 @@ module Optim
            MomentumGradientDescent,
            NelderMead,
            Newton,
-           SimulatedAnnealing
+           SimulatedAnnealing,
+
+           # API
```
johnmyleswhite: Not a classic Julian style, but I think we should standardize on using underscores at all word-like breaks. As is, it's too hard to predict why `iteration_limit_reached` has underscores, but `gnormtrace` doesn't. This is a place where I personally think the style guide for Base is not correct.

src/api.jl
```diff
@@ -0,0 +1,35 @@
+method(r) = r.method
+minimizer(r) = r.minimum
+Base.minimum(r::OptimizationResults) = r.f_minimum
```
johnmyleswhite: If we keep things in the `Optim` namespace, I think we should avoid adding a method to `Base.minimum`.

src/api.jl
```diff
@@ -0,0 +1,35 @@
...
+iterations(r) = r.iterations
+iteration_limit_reached(r) = r.iteration_converged
+trace(r) = r.trace
+function xtrace(r)
```
johnmyleswhite: At some point in the future, I suspect we should replace these dictionaries with types with proper (possibly Nullable) fields. But not for now.

src/api.jl
```diff
@@ -0,0 +1,35 @@
...
+gtol(r::MultivariateOptimizationResults) = r.grtol
+
+initial_state(r::MultivariateOptimizationResults) = r.initial_x
+
+converged(r::UnivariateOptimizationResults) = r.converged
+
+lowerbound(r::UnivariateOptimizationResults) = r.initial_lower
+upperbound(r::UnivariateOptimizationResults) = r.initial_upper
+
+reltol(r::UnivariateOptimizationResults) = r.rel_tol
```
johnmyleswhite: Should we support `rel_tol` for multivariate results? Even if we're not sure, we could introduce methods that throw NotYetImplemented errors.

johnmyleswhite:
Made some comments for discussion. Happy to see this merged whenever since I believe has nothing breaking in it.

src/Optim.jl
pkofod: I agree, but would probably write `gnorm_trace`  and not `g_norm_trace`, for example. I defaulted to no underscores in this first pass, as it is the standard as per the style guide.

johnmyleswhite: Fair enough. I will say that `gnorm` looks like it might actually be an English word, which is what makes it particularly confusing to me.

src/api.jl
pkofod: Havn't really looked into Nullables. I guess the point is that you can have a Null value which indicated that there could be a value, but there isn't one. As opposed to nans which can mean that something went wrong. I like that, as `haskey` checks are not exactly elegant.

johnmyleswhite: Ok. I'll write up an example soon. It would be a breaking change, so it'll take some time to figure out whether it's justifiable. When this code was originally written, I didn't have a good feel for parametric types, so there's places where the design could be cleaner.

pkofod:
Thanks for the comments. I'll see if I can address them today.

src/Optim.jl
pkofod: Didn't see this comment.

Coming back to this, I see your point. Now I can't unsee "gnorm" as some weird "gnome"-typo.

pkofod:
Added errors, gnorm renaming, and very basic readme documentation. Should probably add some tests as well. And squash some of those "oops" commits.

pkofod:
Squashed it all, as there's not really any need for multiple commits.

pkofod:
Should I change all internal use of ".field"  - in a seperate commit, but this PR? Or do we still need to discuss the function names?

pkofod:
Since there has been no objections in the past two or so weeks, I think this can be merged. Nothing should be breaking, but this allows us to break the internal names etc in the future and use deprecation tools.

mlubin:
Any tests for the new API?

pkofod:
Not really. I'll cook something up. I guess you mean something like testing of
```julia
method(res) == res.method
``` 
? or is 
```julia
res = optimize(..., BFGS())
@test method(res) == "BFGS"
```
better

mlubin:
The latter seems more appropriate.

pkofod:
Maybe I should change the keywords as well? `grtol -> g_tol`. It doesn't really make sense to have a discrepancy between the keywords, and the result retrieval methods. I'll do it in a separate commit in this PR.

pkofod:
Tests were added. I'm not sure if I should use `approx` versions or not. On one hand minor changes to algorithms can cause them to fail, on the other hand, it might be relevant to be aware of changes in the achieved minimizer/minimum if something is changed in the package. [off-topic] maybe we should have a script that checks if the minimizer/minimum differs from a "last known good branch" output?[/offt-opic]

pkofod:
With tests added, and the green light given above, I'm going to merge it. About time #76 is closed :)

PR #166, opened 2016-02-03
Use alpha = 1 in linesearch for each new Newton iteration

### LINKED ISSUE #91: Possible inefficiency with step size in Newton method + backtracking linesearch
I noticed that the Newton method is implemented in such a way that the step size (as embodied in the parameter `alpha`) is not reset across iterations. This affects at least the backtracking linesearch.

In practice, if at some iteration the step size as decided by the line search is very small, then at the next iteration, the line search will start with the previous value for `alpha`, instead of starting with one. So the parameter `alpha` can only decrease across iterations when using the backtracking linesearch.

This can be highly inefficient for some problems, resulting in many more iterations than necessary.

Is this a deliberate choice? In my `NLsolve` package, I chose to always start linesearches with `alpha=1`, since it results in much more efficients outcomes for the problems that I tested: https://github.com/EconForge/NLsolve.jl/commit/b14786644a03e9e45de7dd61277839110fab482b

KristofferC:
Fixes #91

### DIFF
--- a/src/newton.jl
+++ b/src/newton.jl
@@ -98,7 +98,7 @@ function newton{T}(d::TwiceDifferentiableFunction,
 
         # Determine the distance of movement along the search line
         alpha, f_update, g_update =
-          linesearch!(d, x, s, x_ls, gr_ls, lsr, alpha, mayterminate)
+          linesearch!(d, x, s, x_ls, gr_ls, lsr, one(T), mayterminate)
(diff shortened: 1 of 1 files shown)


KristofferC:
Actually, I can not comfortable say that this is a net improvement for all the different line searches so someone with more knowledge should probably do this.

Issue #171, opened 2016-02-25
Calling Newton() without gradient and hessian


pkofod:
If you call `optimize` with the `Newton` method without gradient or hessian (which you should not do!), you get a `StackOverflowError`. I'm not quite sure why (just tried this on the way out of the door), but it happens when we try to autodiff. We should probably throw an error telling the user to provide a gradient and a hessian for `Newton()`. Up for grabs, or I'll get back to it when I have time.

```jl
julia> optimize(x->sum(x), [0,3.], method=Newton())
ERROR: StackOverflowError:
 in call at /home/pkm/.julia/v0.4/Optim/src/types.jl:200
 in optimize at /home/pkm/.julia/v0.4/Optim/src/optimize.jl:146
 in optimize at /home/pkm/.julia/v0.4/Optim/src/optimize.jl:150 (repeats 145 times)
```

Evizero:
This looks like it's my fault. Basically, what I assume happens is that the two following methods ping pong each other. Both are fallback methods for when there is no specific one available for the chosen method

https://github.com/JuliaOpt/Optim.jl/blob/master/src/optimize.jl#L141-L158

This ordeal happens because it is hard to decide in the second fallback method (for a `DifferentiableFunction`) if the specified `method` needs only the function itself (without the gradient), or also needs the Hessian. Basically I decided to assume the former, which results in a call to `optimize` with just the function `d.f`. However, there is no such method implemented for `Newton`, which results in the first fallback method being called that transforms the function into a `DifferentiableFunction` and the cycle starts again.

pkofod:
Ah, you're right of course. Since `Newton()` is the only such method (currently), maybe we should just add a hand-written
```julia
optimize(f::Function, initial_x, method::Newton, options::OptimizationOptions) = error("Newton's method without user supplied gradients and hessian is currently not supported.")
```
or something along those lines.

Evizero:
yes, I am working on a patch right now. I'll take your error message

PR #172, opened 2016-02-25
Fix recursion issue when calling Newton without hessian

Evizero:
addresses #171


pkofod:
Great. I know you copy+pasted my error, so it's on me, but maybe "users supplied gradients and hessians" or "a user supplied gradient and hessian"?

pkofod:
lgtm

johnmyleswhite:
Thanks for fixing this.

PR #174, opened 2016-02-28
addition of an alternative version of simulated annealing

mcreel:
This version of SA contracts the search space as the temperature decreases, which causes more intensive search of the promising regions. It also stops when function and/or parameter convergence criteria are met. There is a test file "samin_test.jl" which gives basic usage examples. The source for samin.jl also has explanatory comments. This code has an extensive history, and has been used quite a bit over the years. All authors of code within the history of the Julia version have authorized publishing with the MIT license.

The code follows the article 
Goffe, William L. (1996) "SIMANN: A Global Optimization Algorithm
    using Simulated Annealing " Studies in Nonlinear Dynamics & Econometrics
    Oct96, Vol. 1 Issue 3.

### DIFF
--- a/src/samin.jl
+++ b/src/samin.jl
@@ -11,8 +11,8 @@
   
   The original Fortran code accompanies  the article
   Goffe, William L. (1996) "SIMANN: A Global Optimization Algorithm
-	using Simulated Annealing " Studies in Nonlinear Dynamics & Econometrics
-  	Oct96, Vol. 1 Issue 3.
+    using Simulated Annealing " Studies in Nonlinear Dynamics & Econometrics
+    Oct96, Vol. 1 Issue 3.
 
   This Julia code uses the same names for control variables,
   for the most part. A notable difference is that the initial
@@ -25,12 +25,12 @@
 
   Other relevant articles (among many):
   Corana et. al., (1987) "Minimizing Multimodal Functions of Continuous
-	Variables with the "Simulated Annealing" Algorithm",
- 	ACM Transactions on Mathematical Software, V. 13, N. 3.
+    Variables with the "Simulated Annealing" Algorithm",
+    ACM Transactions on Mathematical Software, V. 13, N. 3.
 
   Goffe, et. al. (1994) "Global Optimization of Statistical Functions
- 	with Simulated Annealing", Journal of Econometrics,
- 	V. 60, N. 1/2.    
+    with Simulated Annealing", Journal of Econometrics,
+    V. 60, N. 1/2.    
 
 usage: x, obj, convergence, details = samin(f,
                                             x_init,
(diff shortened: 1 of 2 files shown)


Evizero:
That is really cool!

Would you mind using 4 spaces instead of tabs, which is the Julia convention AFAIK? Currently it's a mixture of both and it will likely result in strange diffs as soon as someone with auto tab replacement edits the file

mcreel:
Is there an easy way to do the conversion automatically (I'm a vim user)? I
haven't paid any attention to tabs/spaces, to tell you the truth.

On Sun, Feb 28, 2016 at 12:56 PM, Christof Stocker <notifications@github.com

Evizero:
one thing you can do is add this to your `.vimrc` to avoid this problem in the future

```vim
set shiftwidth=4  " operation >> indents 4 columns; << unindents 4 columns
set tabstop=4     " a hard TAB displays as 4 columns
set expandtab     " insert spaces when hitting TABs
set softtabstop=4 " insert/delete 4 spaces when hitting a TAB/BACKSPACE
set shiftround    " round indent to multiple of 'shiftwidth
```

to replace existing tabs with 4 spaces I think the following command should do that (note that there are four spaces)

```vim
:%s/\t/    /g
```

or if you want to be on the supersafe side and see what's going on you can use the following which will promp you for each replace

```vim
:%s/\t/    /gc
```

mcreel:
Ah, found how to do it. Thanks for the reminder about the style.

pkofod:
Thanks for taking to time to sort out the licensing issues. I think it is interesting to get this fit into Optim.jl, especially given the current SA-algorithm. I have a few questions, or thought.

1. Should this be seen as a complement to or substitute for our current SA-algorithm? I'm thinking the latter, right?
2. This should really have the same optimize-API as the rest of the functions. I'm seeing a few different options. One option is to simply make a method with the same syntax as the rest, and then fix it "later". Another option could be to make a branch `samin`, and make the PR to that branch. Then we (we, you, I) can fix all syntax, testing, and so on there, before we make it an officially supported method. Thought?

mcreel:
Getting the agreement was no problem, a couple of quick email messages. Regarding your points:
1. The current code in Optim has two problems, in my view. It just runs for a set number of iterations, without checking for convergence, and it does not contract the search space as the temperature declines. The proposed code does both of these, but it applies to a parameter space defined by box constraints. It seems to me that Optim should have only one SA algorithm, so combining the two implementations seems the way to go to me. 
2. It would be pretty easy to take Optim's current SA, add box constraints, shrinking search area, and convergence checks, just by pulling out parts of the code I proposed. The option to increase the temperature is less important, I would say. Any route the community wants to take is fine with me. I'm busy with other work now, so my participation has to be limited.

pkofod:
All right. In that case, can I make you make a new PR directed at the `pkm/samin` branch instead then? Then I will merge it into our framework over there.

mcreel:
Hi Patrick,
I have done that.
Cheers,
M.

pkofod:
Merged into `pkm/samin`. Thanks again.

mcreel:
Thank you. I'll follow the developments and help out if I can.

PR #178, opened 2016-03-01
fix wrong length in buffers in LM

KristofferC:
also adds a tab -> space. For no whitespace diff: https://github.com/KristofferC/Optim.jl/commit/41a8eb3906da81af693203815cb0d60c644ef5bf?w=1

Confirmed that LsqFit passes after this. We should have a in repo test for this as well but I am a bit short of time now.

### DIFF
--- a/src/levenberg_marquardt.jl
+++ b/src/levenberg_marquardt.jl
@@ -1,145 +1,145 @@
 function levenberg_marquardt{T}(f::Function, g::Function, x0::AbstractVector{T};
-	tolX::Real = 1e-8, tolG::Real = 1e-12, maxIter::Integer = 100,
-	lambda::Real = 10.0, show_trace::Bool = false)
-	# finds argmin sum(f(x).^2) using the Levenberg-Marquardt algorithm
-	#          x
-	# The function f should take an input vector of length n and return an output vector of length m
-	# The function g is the Jacobian of f, and should be an m x n matrix
-	# x0 is an initial guess for the solution
-	# fargs is a tuple of additional arguments to pass to f
-	# available options:
-	#   tolX - search tolerance in x
-	#   tolG - search tolerance in gradient
-	#   maxIter - maximum number of iterations
-	#   lambda - (inverse of) initial trust region radius
-	#   show_trace - print a status summary on each iteration if true
-	# returns: x, J
-	#   x - least squares solution for x
-	#   J - estimate of the Jacobian of f at x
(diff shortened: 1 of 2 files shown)


KristofferC:
cc @yuyichao can you confirm with `LsqFit`?

yuyichao:
`LsqFit` test passes but the change looks wrong?

Sth like
```
diff --git a/src/levenberg_marquardt.jl b/src/levenberg_marquardt.jl
index 5afe1e6..00cad6a 100644
--- a/src/levenberg_marquardt.jl
+++ b/src/levenberg_marquardt.jl
@@ -97,9 +97,9 @@ function levenberg_marquardt{T}(f::Function, g::Function, x0::AbstractVector{T};
         end
         # try the step and compute its quality
         @simd for i in 1:n
-            @inbounds x_buff[i] = x[i] + delta_x[i]
+            @inbounds Jf_buff[i] = x[i] + delta_x[i]
         end
-        trial_f = f(x_buff)
+        trial_f = f(Jf_buff)
         f_calls += 1
         trial_residual = sumabs2(trial_f)
         # step quality = residual change / predicted residual change
```
? Or use a new buffer to avoid aliasing.

KristofferC:
Yeah, what I had only worked because M > N or vice versa.

yuyichao:
Well it is not giving the right dimension to `f` either.

KristofferC:
Bump, should be good now but good to have more eyes. @yuyichao

yuyichao:
The change LGTM. Probably need a test.

pkofod:
Took the liberty of making a PR to your ( @KristofferC ) branch with a test. Basically, it's the test from `LsqFit.jl`, but stripped to only use `Optim.jl` functions. Fails on maste, but works with this pr.

pkofod:
Currently has conflicts, but I will merge when they're resolved.

KristofferC:
Ok will do it now.

pkofod:
Thanks. I think they're partially because of the changed API for getting the results, but it might be other stuff as well.

KristofferC:
I might have messed up and lost your commit authorship in the rebase. Is that ok, else I can try some other way.

KristofferC:
Think I fixed it.

pkofod:
The authorship doesn't matter, I was just trying to be helpful :)

If the tests pass, I'll merge. Great to get this :bug: fixed.

KristofferC:
:+1:

PR #180, opened 2016-03-02
Fix minimum value, max iteration break condition, fix the use of appe…

### LINKED ISSUE #167: CG results f_minimum differs from f(minimum)
Multivariate results [type](https://github.com/JuliaOpt/Optim.jl/blob/master/src/types.jl#L24) contains `f_minimum` field that differs from the value of objective function at the minimum point:

    rosenbrock(x::Array) = (1.0 - x[1])^2 + 100.0*(x[2] - x[1]^2)^2
    optres = Optim.fminbox(Optim.DifferentiableFunction(rosenbrock),
                       zeros(2), -10*ones(2), 10*ones(2),
                       optimizer = Optim.cg, iterations = 1_000)
    println(optres.f_minimum)
    println(Optim.DifferentiableFunction(rosenbrock).f(optres.minimum))

-4.5951198447245615e-8
5.597549090140484e-17

Noticed that first when receiving negative values for minimal sum-squares etc.

I am working with the `teh/constrained` branch by @timholy but Optim-master shows the same problem.

pkofod:
…nd, and remove export.

Fixes #167 

I don't think it's unreasonable for `f_minimum` to return the objective function value at the minimizer, even if a barrier function was added during optimization.

<s>I also added `results.iterations >= iterations`. Before, the algorithm could run beyond the stated maximum number of iterations.</s>
I've also added separate outer and inner options.

Added `first = false` if in the `first == true` branch of the `append!` part of the code, this fixes the returned `MultivariateOptimizationResults` such that it actually adds all calls and iterations together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,10 +278,20 @@ x0 = [2.0, 2.0]
 results = optimize(d4, x0, l, u, Fminbox())  # d4 from rosenbrock example
 ```
 
-This performs optimization with a barrier penalty, successively scaling down the barrier coefficient and using `cg` for convergence at each step.
+This performs optimization with a barrier penalty, successively scaling down the barrier coefficient and using `ConjugateGradient` for convergence at each step.
 
-This algorithm uses diagonal preconditioning to improve the accuracy, and hence is a good example of how to use `cg` with preconditioning. Only the box constraints are used. If you can analytically compute the diagonal of the Hessian of your objective function, you may want to consider writing your own preconditioner (see `nnls` for an example).
+This algorithm uses diagonal preconditioning to improve the accuracy, and hence is a good example of how to use `ConjugateGradient` with preconditioning. Only the box constraints are used. If you can analytically compute the diagonal of the Hessian of your objective function, you may want to consider writing your own preconditioner.
 
+There are two iterations parameters: an outer iterations parameter used to control `Fminbox` and an inner iterations parameter used to control `ConjugateGradient`. For this reason, the options syntax is a bit different from the rest of the package. All parameters regarding the outer iterations are passed as keyword arguments, and options for the interior optimizer is passed as an `OptimizationOptions` type using the keyword `optimizer_o`.
+
+For example, the following restricts the optimization to 2 major iterations
+```julia
+results = optimize(objective, x0, l, u, Fminbox(); iterations = 2)
+```
+In contrast, the following sets the maximum number of iterations for each `ConjugateGradient` optimization to 2
+```julia
+results = Optim.optimize(objective, x0, l, u, Fminbox(); optimizer_o = OptimizationOptions(iterations = 2))
+```
 ### Linear programming
 
 For linear programming and extensions, see the [JuMP](https://github.com/JuliaOpt/JuMP.jl) and [MathProgBase](https://github.com/JuliaOpt/MathProgBase.jl) packages.
--- a/src/fminbox.jl
+++ b/src/fminbox.jl
@@ -128,6 +128,9 @@ function optimize{T<:AbstractFloat}(
         mufactor::T = convert(T, 0.001),
         precondprep = (P, x, l, u, mu) -> precondprepbox(P, x, l, u, mu),
         optimizer = ConjugateGradient,
+        optimizer_o = OptimizationOptions(store_trace = store_trace,
+                                          show_trace = show_trace,
+                                          extended_trace = extended_trace),
         nargs...)
 
     x = copy(initial_x)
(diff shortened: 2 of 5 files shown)


pkofod:
Does this make more sense @multidis ? I think it does. Thanks for the comment.

multidis:
Sorry I am not up to date with the recent API changes, but I assume the iteration limit for `cg` can be passed inside `optimizer_o`. If that is the case then everything seems reasonable. Probably would be helpful to make a note somewhere in the future docs/readme that in the case of `fminbox`, the `iterations` argument relates to barrier function iterations and not the underlying optimizer.

Line 211 is also very helpful, thanks for taking care of that.

pkofod:
Yes, I will make sure to thoroughly document how it is to be used before a version of this pr is merged.

pkofod:
Test error is unrelated.

src/fminbox.jl
```diff
@@ -177,10 +186,10 @@ function optimize{T<:AbstractFloat}(
         end
         pcp = (P, x) -> precondprep(P, x, l, u, mu)
         resultsnew = optimize(dfbox, x, optimizer(eta = eta, linesearch! = linesearch!, P = P, precondprep = pcp),
-                              OptimizationOptions(xtol=xtol, ftol=ftol, grtol=grtol, iterations=iterations,
-                                                  store_trace=store_trace, show_trace=show_trace, extended_trace=extended_trace))
+                              optimizer_o)
         if first == true
```
johnmyleswhite: I believe this doesn't need to be a test against `== true`, `if first` should be sufficient.

test/constrained.jl
```diff
@@ -49,3 +49,11 @@ objective.fg!(results.minimum, g)
...
     @test abs(g[i]) < 3e-3 || (results.minimum[i] < -boxl+1e-3 && g[i] > 0) || (results.minimum[i] > boxl-1e-3 && g[i] < 0)
 end
+
+# tests for #180
+results = Optim.optimize(objective, x0, l, u, Fminbox(); iterations = 2)
+@test results.iterations == 2
+@test results.f_minimum == objective.f(results.minimum)
+
+results = Optim.optimize(objective, x0, l, u, Fminbox(); optimizer_o = OptimizationOptions(iterations = 2))
+@test results.iterations == 470
```
johnmyleswhite: This seems troubling to me. What guarantee do we have that this number of iterations will be performed? If we improved our line search procedure, would this test break?

README.md
```diff
@@ -278,10 +278,20 @@ x0 = [2.0, 2.0]
...
 
-This algorithm uses diagonal preconditioning to improve the accuracy, and hence is a good example of how to use `cg` with preconditioning. Only the box constraints are used. If you can analytically compute the diagonal of the Hessian of your objective function, you may want to consider writing your own preconditioner (see `nnls` for an example).
+This algorithm uses diagonal preconditioning to improve the accuracy, and hence is a good example of how to use `ConjugateGradient` with preconditioning. Only the box constraints are used. If you can analytically compute the diagonal of the Hessian of your objective function, you may want to consider writing your own preconditioner.
 
+As there are both parameters for the `Fminbox` outer iterations, and the inner `ConjugateGradient` iterations, the options syntax is a bit different from the rest of the package. All parameters regarding the outer iterations are passed as keyword arguments, and options for the interior optimizer is passed a an `OptimizationOptions` type using the keyword `optimizer_o`.
```
johnmyleswhite: Maybe write this as "there are two iterations parameters: an outer iterations parameter used to control `Fminbox` and an inner iterations parameter used to control `ConjugateGradient`. As such, the options [...]"

johnmyleswhite:
Made some small stylistic comments. I'm a little worried about the test case, but otherwise think you should merge this whenever you want.

src/fminbox.jl
pkofod: Yes of course.

test/constrained.jl
pkofod: I see your point. You mean the second test right? The first one should not be affected, as a better linesearch would only lower the number of inner iterations. I will have to think about another way to test it.

README.md
pkofod: Thanks, the current version is bad, I see that.

pkofod:
Thank you for the comments. I will address the test, and see if I can come up with something better.

timholy:
I'm not following this closely (I'm on vacation), but just chiming in to say thanks for all the awesome contributions, @pkofod. The impact of your work is going to be huge.

pkofod:
> Made some small stylistic comments. I'm a little worried about the test case, but otherwise think you should merge this whenever you want.

I am worried as well. I added a comment in case a future committer breaks the test by improving an algorithm. I know this is not optimal, but I intend to give `/test/` an overhaul in the future, and this will be the first test I revisit! Is this OK with you?



> I'm not following this closely (I'm on vacation), but just chiming in to say thanks for all the awesome contributions, @pkofod. The impact of your work is going to be huge.

Thanks, but the good old trope of standing on the shoulders of giants is more true than ever in this case. Though, I certainly do intend to throw some time after `Optim.jl` this year.

PR #186, opened 2016-03-29
Getting ready for Optim v0.5

pkofod:
<s>Notice I did _not_ change the names in `MultivariateOptimizationResults`, as this would be breaking changes. I really wanted these names to match the API, and I also wanted `gr` to go in favor of `g`.

This was possible because we are deprecating the functions in favor of a function (`optimize`) with another name. For this reason, I am _sadly_ not able to harmonize `levenberg_marquadt`'s `tolX`, `tolG`, etc.

I propose to tag a minor version soon, and in the ANN on the user-groups let people know that these will change in the next major version bump (not too far in the future, but we still want to give people a chance to adapt). At the same time, we can change the internal names, so they match the function API.</s> see below


pkofod:
I intend to change some variable names as well before merging this.

pkofod:
Added. I think the keyword change in the first commit is rather important for our general API, but of course the name changes of the internal variables are of more cosmetic character. I could leave them be with their old names, but I think consistency tends to increase readability.

pkofod:
Okay, this should probably be in a breaking version bump, as we cannot deprecate the `optimize(f, x_0, method = :bfgs)` version of the old API, and I figure that is quite normal, since that is what's shown in the tutorial.

pkofod:
Notice, I am changing the main tolerance in `NelderMead` to `g_tol`. This is because a) `f_tol` has been reinstated to take take the default value `1e-32` which would either ruin NelderMead (way to low a tol for nmobjective) or we would have to manually overwrite it in the Nelder Mead optimize code, and b) `nmobjective` doesn't measure the relative function value convergence anyway.

pkofod:
So, relative code coverage is down, because a lot of lines in `deprecate` were covered, but no lines are uncovered now, that weren't before.

Allowing myself to call @johnmyleswhite @Evizero : Hello.

The point of this PR is to remove the deprecated functions, and to finalize the API transition. This includes keyword renaming, and also a lot of cosmetic variable renaming (`gr->g` being the most prominent).

I am calling you for two reasons.

I would love a :+1: or :-1: on changing `f_tol` back to `1e-32`. I think it was changed to `1e-8` somewhere in the API redesign. This means that I have to do _something_ to make Nelder Mead stop at a reasonable precision, which should not be `nmobjective < 1e-32`! So I changed the main tolerance parameter in Nelder Mead to `g_tol`. As mentioned above, I know this might seem odd, but since the stopping criterion isn't the relative function value convergence either, I figure it will do. Maybe we want a generic `main_tol` keyword in the future, I'm not sure.

I would also like to confirm, that we are comfortable moving to v0.5 and break the old API for good. I will make sure to put upper bounds on all packages in METADATA, and make an announcement in julia-opt and julia-users.

PR #187, opened 2016-04-14
Adding documentation and use cases for the code

biswag:
I want to add more in-depth documentation for the API and also list some types of problems that can be solved using these algorithms.


mlubin:
This isn't a well-formed pull request, could you try again from a clean branch?

PR #194, opened 2016-05-04
Control flow check values of f_ instead of x_

anriseth:
It seems like a mistake to compare `x_new` with `x_minimum_old` etc. in order to decide whether it is a better minimiser. I assume it was meant to be a comparison between `f_new` and `f_minimum_old` etc.

### DIFF
--- a/src/brent.jl
+++ b/src/brent.jl
@@ -140,12 +140,12 @@ function optimize{T <: AbstractFloat}(
             else
                 x_upper = x_new
             end
-            if x_new <= x_minimum_old || x_minimum_old == x_minimum
+            if f_new <= f_minimum_old || x_minimum_old == x_minimum
                 x_minimum_old_old = x_minimum_old
                 f_minimum_old_old = f_minimum_old
                 x_minimum_old = x_new
                 f_minimum_old = f_new
-            elseif x_new <= x_minimum_old_old || x_minimum_old_old == x_minimum || x_minimum_old_old == x_minimum_old
+            elseif f_new <= f_minimum_old_old || x_minimum_old_old == x_minimum || x_minimum_old_old == x_minimum_old
                 x_minimum_old_old = x_new
                 f_minimum_old_old = f_new
             end


pkofod:
Thank you for the pull request. I think you are correct that the first part of the last control staatement should be f, but I don't think the rest should. I will find my Brent.

pkofod:
Okay, so I've looked at it, and you are correct: there certainly is a bug. However, it is not as you've corrected it, but instead as I wrote above. In the highlighted line at https://github.com/JuliaOpt/Optim.jl/pull/194/files#diff-45ad294d1d1ea06c582a55b8f78c92a6L143 it should be
```julia
if f_new <= f_minimum_old || x_minimum_old == x_minimum
```
and at 
```julia
elseif f_new <= f_minimum_old_old || x_minimum_old_old == x_minimum || x_minimum_old_old == x_minimum_old
```
It is very unfortunate that this bug has been here, and it should certainly be fixed. If you still disagree with me, I would like to know what source you are referring to, so I can have a look.

anriseth:
Ah, thanks for checking it. I didn't look up the algorithm anywhere - it just seemed odd to me that it wouldn't check f_new vs f_minimum_old etc. 

I've updated the code with your corrections.

pkofod:
Just to document it (I should probably comment or clean the code): Brent's notation -> Our notation

- fu -> `f_new`
- fw ->`f_minimum_old`
- fv ->`f_minimum_old_old`
- u ->`x_new`
- w ->`x_minimum_old`
- v ->`x_minimum_old_old`
- x ->`x_minimum`

The changes in this PR are in lines analogous to the lines starting 9 lines from the bottom of Brent's algorithm in his ALGOL 60 algorithm in his book.

pkofod:
I have not forgotten this PR, I just want to go through the algorithm in its entirety before merging this, and I havn't had the time.

pkofod:
Went through the algorithm, and I am quite sure that we no longer have any bugs in the implementation of Brent's algorithm. Thank you for the PR!

PR #195, opened 2016-05-11
fix DimensionMismatch error

bjarthur:
`levenberg_marquardt` throws an error when called from `LsqFit/curve_fit`.  not sure this is the best solution, but it works.

### DIFF
--- a/src/levenberg_marquardt.jl
+++ b/src/levenberg_marquardt.jl
@@ -43,6 +43,7 @@ function levenberg_marquardt{T}(f::Function, g::Function, x0::AbstractVector{T};
 	m = length(fcur)
 	n = length(x)
 	JJ = Matrix{T}(n, n)
+	Jm_buff = Vector{T}(m)
 	Jf_buff = Vector{T}(n)
 	x_buff = Vector{T}(n)
 
@@ -85,9 +86,9 @@ function levenberg_marquardt{T}(f::Function, g::Function, x0::AbstractVector{T};
 
 		# if the linear assumption is valid, our new residual should be:
 		# predicted_residual = sumabs2(J*delta_x + fcur)
-		A_mul_B!(Jf_buff, J, delta_x)
-		LinAlg.axpy!(1, fcur, Jf_buff)
-		predicted_residual = sumabs2(Jf_buff)
+		A_mul_B!(Jm_buff, J, delta_x)
+		LinAlg.axpy!(1, fcur, Jm_buff)
+		predicted_residual = sumabs2(Jm_buff)
 		# check for numerical problems in solving for delta_x by ensuring that the predicted residual is smaller
 		# than the current residual
 		if predicted_residual > residual + 2max(eps(predicted_residual),eps(residual))


KristofferC:
This has been fixed I believe in https://github.com/JuliaOpt/Optim.jl/pull/178. Does it still throw after you do a `Pkg.update()`?

bjarthur:
thanks.  my bad.  forgot to pull after checking out master.

related to this same section of code, when JJ on line 84 is singular an exception is thrown.  is there anyway to deal with this gracefully and proceed to try to find a solution?

pkofod:
> related to this same section of code, when JJ on line 84 is singular an exception is thrown. is there anyway to deal with this gracefully and proceed to try to find a solution?

what would you propose?

bjarthur:
i'm not an expert in optimization, just a user.  all i know is that matlab and python, as well as Optim.jl 0.4.4, don't throw nearly as many exceptions on my data as Optim.jl 0.4.5+.

see these gists:  [julia](https://gist.github.com/bjarthur/9c818bba8b4a74eb07f40f879e6c2df8), [matlab](https://gist.github.com/bjarthur/81a1de590934fac29393547f9812a991), [python](https://gist.github.com/bjarthur/aecaecfa137b1f5f7a312ce9341b5f6e).  the data are the same in each, and i could supply you with hundreds of more examples for which 0.4.5 throws an error, and the others don't.

would also be nice to have the option to bound the parameter values.

happy to help in anyway i can.  thanks.

KristofferC:
The best way to help is to provide code in a PR that works better than the current code with tests that used to fail but now work.

bjarthur:
ofc.  but not knowing anything about optimization, my role would best be in test.  i've given you the test case that fails.  that's half the work.

looking into this more, it is OSX that works, and linux that throws an exception; the problem is not 0.4.4 vs 0.4.5...

johnmyleswhite:
Historical note I'll throw in: the Optim package was built by people who learned about optimization so they could write code they needed for their own work; the three original core authors were a psychology graduate student, a neuroscience professor and an ML professor. Non-experts can definitely get up to speed since we're never doing anything novel in this package; just implementing textbook algorithms.

pkofod:
If Linux fails and OS X does not, then it's probably a matter of tolerances (or whatever) in the linalg lib you have. Matlab and Python are probably checking (or try catching) the factorization and doing something smart. Efforts to this effect are probably better spent in the core linalg module than here. Unless there is something sensible that we can lift from the literature.

bjarthur:
`show_trace=true` reveals that both OS X and linux in the gist above have `dx`s which wildly increase to `~1e8` after ~30 iterations.  on OS X, there is a sudden reversal in sign and decrease in magnitude which leads to `dx < 1e-6` after another ~20 iterations.  on linux though the magnitude stays high and a `SingularException` is thrown ~5 iterations later.

interestingly, when the search space is bounded with https://github.com/JuliaOpt/Optim.jl/pull/196, `dx` never exceeds 1200 on linux and a solution is reached after 15 iterations.  since this fixes my `SingularException` problem, i'm not inclined to pursue the difference in behavior between OS X and linux anymore.  feel free to close this PR (which should now be an issue i suppose) as you see fit.

pkofod:
I guess any further discussion belongs in #196 as mentioned above. Closing.

PR #196, opened 2016-05-13
added bounds argument to levenberg_marquardt

bjarthur:
clamping the search space drastically improves the fit in my use case, and prevents `SingularException`s being thrown.

while at it i also reformatted the docs to conform to current guidelines.

### DIFF
--- a/src/levenberg_marquardt.jl
+++ b/src/levenberg_marquardt.jl
@@ -1,23 +1,34 @@
 immutable LevenbergMarquardt <: Optimizer end
 
+"""
+    `levenberg_marquardt(f, g, x0; <keyword arguments>`
+
+Returns the argmin over x of `sum(f(x).^2)` using the Levenberg-Marquardt algorithm, and an
+estimate of the Jacobian of `f` at x.
+
+The function `f` should take an input vector of length n and return an output vector of length m.
+The function `g` is the Jacobian of f, and should be an m x n matrix.
+`x0` is an initial guess for the solution.
+
+Implements box constraints as described in Kanzow, Yamashita, Fukushima (2004; J Comp & Applied Math).
+
+# Keyword arguments
+* `tolX::Real=1e-8`: search tolerance in x
+* `tolG::Real=1e-12`: search tolerance in gradient
+* `maxIter::Integer=100`: maximum number of iterations
+* `lambda::Real=10.0`: (inverse of) initial trust region radius
+* `show_trace::Bool=false`: print a status summary on each iteration if true
+* `lower,upper=[]`: bound solution to these limits
+"""
 function levenberg_marquardt{T}(f::Function, g::Function, x0::AbstractVector{T};
     tolX::Real = 1e-8, tolG::Real = 1e-12, maxIter::Integer = 100,
-    lambda::Real = 10.0, show_trace::Bool = false)
-    # finds argmin sum(f(x).^2) using the Levenberg-Marquardt algorithm
-    #          x
-    # The function f should take an input vector of length n and return an output vector of length m
-    # The function g is the Jacobian of f, and should be an m x n matrix
-    # x0 is an initial guess for the solution
-    # fargs is a tuple of additional arguments to pass to f
-    # available options:
-    #   tolX - search tolerance in x
-    #   tolG - search tolerance in gradient
-    #   maxIter - maximum number of iterations
-    #   lambda - (inverse of) initial trust region radius
-    #   show_trace - print a status summary on each iteration if true
-    # returns: x, J
-    #   x - least squares solution for x
-    #   J - estimate of the Jacobian of f at x
+    lambda::Real = 10.0, show_trace::Bool = false, lower::Vector{T} = Array(T,0), upper::Vector{T} = Array(T,0))
+
+    # check parameters
+    ((isempty(lower) || length(lower)==length(x0)) && (isempty(upper) || length(upper)==length(x0))) ||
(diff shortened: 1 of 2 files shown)


pkofod:
@yuyichao or @KristofferC any chance you could take a look at this? I know these are busy times, but the change is relatively small.

I will make one comment myself. The particle swarm PR #218 also adds bounds, and we want to be consistent in the bounds interface, so we have to choose one or the other.

matthieugomez:
Is there any reference for your fix? I would do the following:
- check the bounds before computing `f(n_buffer)` 
- If `n_buffer` is outside the bound, go to the case  `rho < MIN_STEP_QUALITY`)

bjarthur:
i could be mistaken, but i believe algorithm 3.1 in http://www.mathematik.uni-wuerzburg.de/~kanzow/paper/CNSLMP.pdf is equivalent to what i propose.  specifically, their function P<sub>X</sub> is a projection onto the constrained solution space (defined at the end of section 1).  this would be a `clamp` in Julia, no?

matthieugomez:
Cool! And indeed I played with it and your solution does work better. I'm not
a maintainer here but I think it'd be great if you did the following

  * Check that you're within the bound before computing `f(n_buffer)`
  * Write the bounds as l = [1.25, -2.1] u = [Inf, Inf]  rather than [1.25 Inf ; -2.1 Inf], to be consistent with  the optim funcction



On Jul 13 2016, at 9:17 am, Ben Arthur &lt;notifications@github.com&gt; wrote:  

> i could be mistaken, but i believe algorithm 3.1 in [http://www.mathematik
.uni-wuerzburg.de/~kanzow/paper/CNSLMP.pdf](http://www.mathematik.uni-
wuerzburg.de/%7Ekanzow/paper/CNSLMP.pdf&r=cmVwbHkrMDA1ZWY3ZGQ0MmU1N2ExMmU1ZjBmMTcxYjViNGIyNGNjMmFkNGM5NmM5ZmJlN2ZiOTJjZjAwMDAwMDAxMTM5ZTAxNjY5MmExNjljZTA5Mzk4MTkwQHJlcGx5LmdpdGh1Yi5jb20=)
is equivalent to what i propose. specifically, their function PX is a
projection onto the constrained solution space (defined at the end of section
1). this would be a clamp in Julia, no?

>

> —  
> You are receiving this because you commented.  
> Reply to this email directly, [view it on
GitHub](https://github.com/JuliaOpt/Optim.jl/pull/196#issuecomment-232351512&r=cmVwbHkrMDA1ZWY3ZGQ0MmU1N2ExMmU1ZjBmMTcxYjViNGIyNGNjMmFkNGM5NmM5ZmJlN2ZiOTJjZjAwMDAwMDAxMTM5ZTAxNjY5MmExNjljZTA5Mzk4MTkwQHJlcGx5LmdpdGh1Yi5jb20=),
or [mute the
thread](https://github.com/notifications/unsubscribe/AF733ambsOuQL5nZuq1INWj2XwYZXUtvks5qVOVmgaJpZM4IeOmS&r=cmVwbHkrMDA1ZWY3ZGQ0MmU1N2ExMmU1ZjBmMTcxYjViNGIyNGNjMmFkNGM5NmM5ZmJlN2ZiOTJjZjAwMDAwMDAxMTM5ZTAxNjY5MmExNjljZTA5Mzk4MTkwQHJlcGx5LmdpdGh1Yi5jb20=).

pkofod:
Good to see further discussion here! Thanks for the link, I will have a look. I agree with the recent changes of bounds being two vectors (I think this is quite standard in many other pieces of software, and also in the particle swarm PR).

bjarthur:
box constraints now specified as two vectors, and the projection is done before the calculation of the residual.  seems to work well, though i've had to increase the maximum number of permissible iterations to get it to converge consistently.

pkofod:
Great. 

> seems to work well, though i've had to increase the maximum number of permissible iterations to get it to converge consistently.

I know this is kind of ping-pong, but if you have a reference, and can demonstrate that the "old" version is better, then by no ways am I going to force an inefficient implementation :)

bjarthur:
looking into it further, the problem was just a poor choice of tests.  kraznow et al actually say to project just @matthieugomez suggested, which is how the code stands now.

pkofod:
Okay, then I guess we're pretty close to a merge. I'm not going to ask you to document LM, as you didn't propose to add it from scratch, but thanks for putting the reference in the comments.

I would prefer if `lower` and `upper` where `x_min` and `x_max` for consistency reasons. Also, the "out of bounds x0" cases should be tested. I can make a PR against your branch, or you can include them, that's up to you.

bjarthur:
consistency is good, and i'm happy to make the change, but a quick [search](https://github.com/JuliaOpt/Optim.jl/search?utf8=%E2%9C%93&q=lower) in this package for "lower" shows that it's used quite frequently.

i'll push a test for the out of bounds shortly.

pkofod:
Point taken :) I actually so like lower and upper better now that I look at the different variable names.

pkofod:
I see you added the tests, thank you! I am wondering, is there a reason you didn't use `@test_throws`? Don't bother changing it, I was just wondering. I think the OSX travis builds are having problems today.

bjarthur:
didn't know about `@test_throws`.  much more concise!  change made.  thanks.

pkofod:
As `lower` and `upper` are used in `teh/constrained`, we'll go with that! Thanks a lot, sorry for the long processing time.

matthieugomez:
Thanks @bjarthur and @pkofod

PR #210, opened 2016-05-25
Re-implemented Preconditioning

cortner:
This uses only method that are available in base and hence can be easily overloaded without needing to import anything from `Optim.jl`.  The main "conceptual change" is that `P` now described the matrix and not its inverse. This is good for consistency with the iterative linear algebra packages.

A few specific items to consider when reviewing:
* `Base.A_ldiv_B!(x, P::AbstractMatrix, b)` is not implemented in Base, though there is a pull request for it. I now added it to precon.jl, but will remove it once it is no longer necessary.
* I kept the name `precondprep!` for the sake of history, though was tempted to replace it with `updateP!`, which I think is more descriptive. (but no big deal)\
* for the diagonal preconditions I replaced SIMD loops with standard vectorised instructions. I just thought this was overkill, but I am happy to reverse this if so desired.

**EDIT:** I am no convinced that `precondprep!` needs to remain an argument.

* `precondprep!`: ~~My original intent was to replace it with a dispatch mechanism, but then this creates the problem that it has to imported as there is no general abstract interface that one could import instead. passing it as an argument remains more flexible. However, I would be happy to discuss this point.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,7 +372,7 @@ The `GradientDescent`, `ConjugateGradient` and `LBFGS` methods support precondit
 can be thought of as a change of coordinates under which the Hessian is better conditioned. With a
 "good" preconditioner substantially improved convergence is possible.
 
-An example of this is shown below (`Optimizer` ∈ {`GradientDescent`, `ConjugateGradient`, `LBFGS`}).
+An example of this is
 ```jl
 using ForwardDiff
 plap(U; n=length(U)) = (n-1) * sum( (0.1 + diff(U).^2).^2 ) - sum(U) / (n-1)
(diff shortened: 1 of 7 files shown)


cortner:
fixing it now . . .

cortner:
I don't understand codecov - can you explain?

KristofferC:
The percentage of lines covered in testing went down with 0.01 percentage units. Don't worry about it.

pkofod:
Codecov basically runs the test folder files, and keeps track of which lines are "touched" throughout the tests. So for example, it might be that the shrink branch in nelder mead is never active in any of our tests (as it is quite rare in practice to see a shrink step), so in that case our code coverage can never be 100%; we need to find a problem which is known to produce a lot of shrink-steps. I'm not sure what happened here, but as mentioned above, it doesn't matter.

What does matter is the merge conflict. This happened, because I've made some changes earlier, and git can't figure out how to combine our changes. Do you know how to rebase? It's a good think to know, but if you don't feel comfortable doing it, I can make a PR against your branch that rebases it.

cortner:
I'll have a go at rebase - I'll let you know if I get stuck.

cortner:
look ok now? I really hope I didn't mess it up. I always worry when I `push -f`

pkofod:
Sorry if I'm being a bit slow here. Why is it that the preconditioner in fminbox saves the inverse diagonal, but in the other algorithms it can be something else? I mean, why is it forced?

cortner:
It really doesn't matter, I just kept @timholy's implementation

cortner:
I probably misunderstood: fminbox chooses a specific preconditioner to precondition the barrier, I think. I haven't thought about how to precondition the rest of the algorithm. I'd need some use cases before I can implement that.

pkofod:
I realize you just kept the original setup, I was just wondering if we wanted to maybe change it, but let's keep it as it is. The original setup assumes ( and I believe this is stated in the comments somewhere) that the passed preconditioner will store the inverse of `P`. Now, however, we assume that it is always `P^-1` and also a diagonal matrix. We may want to warn the users that passing the a preconditioner to fminbox has to respect that `P` is of the InverseDiagonal type.

So I propose that we don't change it (I intend to do something about the unfinished `teh/constrained` branch anyway), just document that `InverseDiagonal` is forced in `fminbox`. Does that make sense?

cortner:
No, this is a misunderstanding. InverseDiagonal is a type that stores a matrix via the inverse a of its diagonal elements.

At the moment a user should not pass a preconditioner to fminbox unless they really know what they are doing.

pkofod:
It probably won't matter, but what am I misunderstanding? The preconditioner setup in Fminbox only works with `P <: InverseDiagonal`, cf https://github.com/JuliaOpt/Optim.jl/pull/210/files#diff-5f609c3f539ad021915e56d12f4b21d3R141. If someone was to pass another `precondprep!` and assumed they could do the same as with preconditioning in cg, lbfgs, etc, then they would get an error when trying to write into `P`. Unless their preconditioner was given as an inverse of a diagonal matrix. If what you're saying is that the preconditioner shouldn't be changed, then I'm just saying that it shouldn't be accessible through the API, or it should at least be documented. It doesn't have to happen in this PR.

cortner:
I see, that is actually a strong argument in favour of replacing precondprep with dispatch. Let me think about it please

cortner:
* dispatch is not flexible enough for precondprep, so we leave it as it is
* Re fminbox: two options
   - leave as is and document that if a different preconditioner is passed then a new precondprep! Must be provided
   - Check whether `precondprep! == nothing` and in this case apply the default 

I think the latter is cleaner for the user but the code a little messier.

pkofod:
Thanks for the good work, and I hope someone out there benefits from the changes to how we implement preconditioning.

cortner:
wonderful - thank you. It is really useful for me to have these algorithms available in preconditioned form.

PR #212, opened 2016-05-28
 Small test for content of extended trace in BFGS and NelderMead.

pkofod:
Making sure that the fields are actually there, as required.


pkofod:
Finally codecov is happy :)

Issue #213, opened 2016-05-28
What is the history of history?


pkofod:
What is the history of adding `history.jl`? It doesn't seem to be used. @johnmyleswhite do you remember? I know it is 3 years ago :) If there no reason, I'm removing it.

johnmyleswhite:
You can remove that. It was supposed to be a template for a function I planned to write to store traces, but I never worked on it.

pkofod:
Cool, thanks.

PR #216, opened 2016-05-31
Pkm/clean up

pkofod:


pkofod:
The inbounds macro adds a tiny (epsilon) speedup for very large vectors (300). Everything counts, right? The other things closes #213 and adds some simple trace tests.

Issue #224, opened 2016-06-14
GSoC: Documentation


pkofod:
ref #200 

I went ahead and tried to use Documenter.jl to generate our coming docs. It can be found at http://pkofod.github.io/OptimDoc.jl/latest/, as it won't generate, unless something is pushed to master (a PR to Optim wouldn't create a doc you could see). It's still WIP, but comments are still welcome.
- [ ] Add more solver details
- [x] Fix nelder mead once #220 is merged
- [x] Make sure to have a stable and latest version of the docs
- [x] Document AD better

edit: stable and latest are online http://www.juliaopt.org/Optim.jl/stable/ http://www.juliaopt.org/Optim.jl/latest/

KristofferC:
Do you want comments about small stuff like grammar / spelling? Any way comments can be done inline (like on a commit)?

pkofod:
Maybe I should go ahead and create a mirrored PR as well? I can see how comments can be a pain to give by referencing the generated docs. Grammar/spelling/ideas/don't do it like this/... everything is welcome if it improves the docs :)

pkofod:
Soo, I'm trying to get it to build, but I had some issues because I didn't have admin access at first. Let's see if the latest commit works! I'm only going to have a "latest" version until we tag the next version, and then we're going to have a latest and a stable version.

pkofod:
Seems to be working

PR #237, opened 2016-06-30
Improve Readme for Box minimization

matthieugomez:
Make d4 explicit + show optimizer option


pkofod:
Thanks. I'll merge it, but we're moving towards a proper documentation system, see #224

PR #238, opened 2016-06-30
Newton trust region

rgiordan:
This is a pretty straightforward implementation of the Nocedal and Wright's exact trust region method.  I've tried to write it flexibly enough that in the future people can implement their own sub-problem solutions.

This is my first PR to a major julia package, so please let me know if there's anything you want me to do differently!


pkofod:
Interesting! I've thought about how we could implement a `Newton()` constructor such that we could use dispatch to swtich between line search and trust region (given someone (you) implemented a trust region implementation).

I will review you PR, but I am trying to wrap up some code to do benchmarking and performance tracking of our solvers, so I am focusing on that at the moment (trying to figure out how to run the very high dimensional problems...). Don't be discouraged if I don't reply straight away, I will come back to it! Maybe others will come with comments as well.

rgiordan:
Sounds good, I'll be happy to support you review whenever you have time to get to it.  I the meantime, I'll try to figure out why it's failing on v0.5.

test/newton_trust_region.jl
```diff
@@ -0,0 +1,144 @@
...
+delta = 0.5 * sqrt(s_norm2)
+m, interior, lambda = Optim.solve_tr_subproblem!(gr, H, delta, s)
+@assert !interior
+@assert abs(lambda + L[1]) < 1e-4
+@assert abs(norm(s) - delta) < 1e-12
+
+
+#######################################
+# Next, test on actual optimization problems.
+using Optim
```
anriseth: Duplicate using statements from lines 1-2

rgiordan: Fixed, sorry about that.

rgiordan:
Travis will have the final say, but I believe the julia 5 tests are now fixed.

pkofod:
What was the problem?

rgiordan:
The main problem was that my test redefined functions (f, g!, and h!) in
the outermost scope, and that was somehow messing up the creation of a
TwiceDifferentiableFunction -- fg! would not return the correct
derivatives.  I don't know exactly why that was happening, but when I
mimicked the test for newton's method and renamed then f_2, g!_2, and h!_2
the problem went away.

Ryan

pkofod:
I believe v0.5 has introduced some changes to the ability to overwrite methods. Good to know it was an easy fix.

pkofod:
I believe `chol(A::Hermitian)` might be a v0.5 signature.

anriseth:
Are you sure you need to do the f_2, g!_2 and h!_2 renaming as well? That seems a bit arbitrary.

pkofod:
https://github.com/JuliaOpt/Optim.jl/pull/164

rgiordan:
Fixed: I check for the version number in the library and use or do not use Hermitian depending on the version.

src/newton_trust_region.jl
```diff
@@ -0,0 +1,378 @@
...
+#
+# Args:
+#  H_eigv: The eigenvalues of H, low to high
+#  qg: The inner product of the eigenvalues and the gradient in the same order
+#
+# Returns:
+#  hard_case: Whether it is a candidate for the hard case
+#  lambda_1_multiplicity: The number of times the lowest eigenvalue is repeated,
+#                         which is only correct if hard_case is true.
+function check_hard_case_candidate(H_eigv, qg)
```
johnmyleswhite: We should make sure this has the same 4-spaces indentation as the rest of the codebase.

src/newton_trust_region.jl
```diff
@@ -0,0 +1,378 @@
...
+    const delta_sq = delta^2
+
+    @assert n == length(s)
+    @assert (n, n) == size(H)
+    @assert max_iters >= 1
+
+    # symmetrizing H...this shouldn't be necessary, but currently the
+    # eigenvalues are only sorted if H is perfectly symmetric.
+    # (Julia issue #17093)
+    H += H'
```
johnmyleswhite: Why not sort the eigenvalues manually? As far as I can tell, for a NxN matrix, this computation is O(N^2) whereas the sorting is only O(N log N).

src/newton_trust_region.jl
```diff
@@ -0,0 +1,378 @@
...
+    H /= 2.
```
johnmyleswhite: You also might want to think about doing these operations in place.

src/newton_trust_region.jl
```diff
@@ -0,0 +1,378 @@
...
+    end
+
+    m = vecdot(gr, s) + 0.5 * vecdot(s, H_ridged * s)
+
+    return m, interior, lambda
+end
+
+
+immutable NewtonTrustRegion <: Optimizer
+  initial_delta::Real
```
johnmyleswhite: Don't create types with fields of type `Real`. If you want these fields to be of arbitrary type, please use a parametric type like:

```
immutable NewtonTrustRegion{T <: Real} <: Optimizer
    initial_delta::T
    delta_hat::T
    eta::T
    rho_lower::T
    rho_upper::T
end

File: test/newton_trust_region.jl [hunk 6]
```diff
@@ -0,0 +1,142 @@
...
+  0.5 * (x[1]^2 + eta * x[2]^2)
+end
+
+function g!_2(x::Vector, storage::Vector)
+  storage[1] = x[1]
+  storage[2] = eta * x[2]
+end
+
+function h!_2(x::Vector, storage::Matrix)
+  storage[1, 1] = 1.0
```
johnmyleswhite: Strive for consistent whitespace here as well.

johnmyleswhite:
This looks really like a really high quality PR. I'm not that familiar with trust region methods, so I'll let @pkofod do review of that, but I made some Julia-level comments.

Thanks for working on this!

src/newton_trust_region.jl
rgiordan: Actually, per a suggestion made in the referenced issue, using Symmetric() should suffice.  As far as I can tell, Symmetric doesn't do any actual computation.  So I can just remove the symmetrization here.

src/newton_trust_region.jl
rgiordan: See above -- I'll just not do this.

src/newton_trust_region.jl
rgiordan: Done.

src/newton_trust_region.jl
rgiordan: Sorry about that, fixed.

test/newton_trust_region.jl
rgiordan: Done.

johnmyleswhite:
Thanks for all the changes!

pkofod:
I'll comment on it tonight local time, so within the next 15 hours :)

src/optimize.jl
```diff
@@ -70,6 +70,30 @@ function optimize(f::Function,
     optimize(f, g!, h!, initial_x, method, options)
 end
 
+function optimize(f::Function,
```
pkofod: Is it intentional that you're making NewtonTrustRegion the default for the case where a function, a gradient, a hessian, and an initial x is provided? If so, the above method definition should be removed. I do kinda feel we need some benchmarks first, though.

rgiordan: Thank you for pointing this out.  No, that wasn't intentional.  In fact, now that I look at it, I'm not sure how julia would resolve a call to optimize(f, g!, h!, initial_x) since both the Newton and NewtonTrustRegion would match that signature, only with different arguments.  Are you implying that julia will take the later definition?  Or is there some other julia-fu that I'm missing?

I'll be happy to change this however you see fit.  I agree that it is probably premature to make this the default.

pkofod: It will just get overwritten, such that only your signature will be used.
```julia
julia> f(x;y=1)=x+y
f (generic function with 1 method)

julia> f(x;y=2)=x+y
f (generic function with 1 method)

julia> f(3)
5
```

rgiordan: Gotcha.  This was a mistake on my part.  I think there is no harm in simply removing the trust region default definition here.  I'll push this change shortly.

src/newton_trust_region.jl
```diff
@@ -0,0 +1,375 @@
...
+            for i=1:n
+                H_ridged[i, i] = H[i, i] + lambda
+            end
+
+            for iter in 1:max_iters
+                lambda_previous = lambda
+
+                # Version 0.5 requires an exactly symmetric matrix, but
+                # version 0.4 does not have this function signature for chol().
+                R = VERSION < v"0.5-" ?
```
pkofod: small thing, but I think this should just be one line for readability, or an explicit if-else-end statement

src/newton_trust_region.jl
```diff
@@ -0,0 +1,375 @@
...
+#
+# Returns:
+#  m - The numeric value of the quadratic minimization.
+#  interior - A boolean indicating whether the solution was interior
+#  lambda - The chosen regularizing quantity
+function solve_tr_subproblem!{T}(gr::Vector{T},
+                                 H::Matrix{T},
+                                 delta::T,
+                                 s::Vector{T};
+                                 tolerance=1e-10,
```
pkofod: do you have any idea which one is hit most often with the small julia-function test suite we're using currently in the tests ? I mran is 1e-10 reached often, or does it basically just run 5 times each time?

src/newton_trust_region.jl
rgiordan: Done, will be in next push.

src/newton_trust_region.jl
```diff
@@ -0,0 +1,375 @@
...
+                                 max_iters=5)
+    const n = length(gr)
+    const delta_sq = delta^2
+
+    @assert n == length(s)
```
pkofod: are these asserts really needed?  I mean, s is `Array(T, n)`, and max_iters should be common sense .

src/newton_trust_region.jl
```diff
@@ -0,0 +1,375 @@
...
+#  m - The numeric value of the quadratic minimization.
+#  interior - A boolean indicating whether the solution was interior
+#  lambda - The chosen regularizing quantity
+function solve_tr_subproblem!{T}(gr::Vector{T},
+                                 H::Matrix{T},
+                                 delta::T,
+                                 s::Vector{T};
+                                 tolerance=1e-10,
+                                 max_iters=5)
+    const n = length(gr)
```
pkofod: I'm pretty sure these const'ss are unnecessary, n and delta_sq could still be changed afterwards. 

```jlcon
julia> function f(x)
       const y = 4
       y = x
       end
f (generic function with 1 method)

julia> f(1)
1
```

pkofod:
Looks pretty good to me. I'll take one more look at the subproblem-function tomorrow, but nothing really stood out. I'm curious if the `1e-10` tolerance is a bit tight, but I've never actually implemented a trust-region solver, so I am actually just curious. Have you tested if it just hits the max_iters-upper bound all the time? I also really love all the comments.

Would it be possible to add a simple page in the docs?

src/newton_trust_region.jl
rgiordan: I don't know.  Let me know if you want me to dig in and find out.

We've been using this code for a while on a [particular problem in astronomy](https://github.com/jeff-regier/Celeste.jl).  When I played around with this in that context, I decided that when you need more than five iterations, you're probably not close to the optimum anyway and getting lambda exactly right wasn't worth it.  I didn't demonstrate this carefully, however, and I haven't explored the question on other problems.

Perhaps this should be made a user-specifiable option.  LMK if you want me to do that.

src/newton_trust_region.jl
rgiordan: My style is to err on the side of more asserts than fewer.  It'll be your code going forwards, though, so I'll defer to you if you want to get rid of them.

src/newton_trust_region.jl
rgiordan: Huh, that's news to me.  If it doesn't actually mean ```const```, then I'll get rid of it.  Done, next push.

rgiordan:
I'll be happy to add a doc page tonight or tomorrow sometime.  Now that you bring it up, I do think that I should at least add the outcome of the subproblem to the trace, so I'll do that too.

pkofod:
> I should at least add the outcome of the subproblem to the trace

Yes, I think it's a good idea to add it to the extended trace. This gives a bit more "debugging" information to users.

rgiordan:
On the test problems, it is the case that most of the time the subproblem is being solved in less than five iterations.

However, I discovered a different problem through the Exponential test problem -- I think line 369 can cause problems by setting a very small delta when the problem is near convergence.  The Exponential problem is getting to the right solution, but it's setting a bad delta right at the optimum and so terminating due to hitting max iters rather than detecting convergence.  The tests passed because they only checked that it got to the solution, not that it actually converged according to something other than hitting max iters.

I'll fix this and ping this PR when I'm done.

pkofod:
Really great stuff. I think it's pretty close to done. Dont go too crazy with the docs, it doesn't have to be a whole text book chapter :)

rgiordan:
The problem was that when you get so close to convergence that the function difference is below machine precision (which can happen when f_tol is 1e-32 as it is by default), then the algorithm doesn't know the difference between a mismatch between the function and the quadratic and floating point error.  I decided to go with a fix that assesses convergence when the quadratic decrease is less than machine precision in the hope that the gradient will converge quickly whenever this happens.

I'll whip up a doc shortly.

rgiordan:
Documentation done, PTAL.

docs/src/algo/newton_trust_region.jl
```diff
@@ -0,0 +1,51 @@
...
+
+If the trust region size, $\Delta_k$, is large enough that the minimizer of the quadratic approximation $m_k(p)$ has $||p|| \le \Delta_k$, then the step is the same as an ordinary Newton step.  However, if the minimizer lies outside the trust region, then the minimizer will occur on the boundary, and the step will have $||p|| = \Delta_k$.  It turns out that when the Cholesky decomposition of $B_k$ can be computed, the optimal $p$ can be found numerically with relative ease.  (See [1], section 4.3)  This is the method currently used in ```Optim.jl```.
+
+It makes sense to adapt the trust region size, $\Delta_k$, as we move through the space and assess the quality of the quadratic fit.  This adaptation is controlled by the parameters $\eta$, $\rho_{lower}$, and $\rho_{upper}$, which are parameters to the ```NewtonTrustRegion``` optimization module.  For each step, we calculate
```
pkofod: "This adaptation is controlled by the parameters $\eta$, $\rho_{lower}$, and $\rho_{upper}$ in the ```NewtonTrustRegion``` type"?

At least, I'm not sure module is the right word here, especially as you are using the exact name of the type.

docs/src/algo/newton_trust_region.jl
```diff
@@ -0,0 +1,51 @@
...
+
+$\rho_k := \frac{f(x_{k+1}) - f(x_k)}{m_k}$
```
pkofod: shouldn't it be `m_k(0)-m_k(x_{k+1}-x_{k})` in the denominator?

pkofod:
Looks really great. You should add the new file to the mkdocs yml file like for newton: https://github.com/JuliaOpt/Optim.jl/blob/48a4d82a6db8a503ec539167b2e70f2ba29c5b46/docs/mkdocs.yml#L46 or else it will not appear on the website.

Other than that, this looks pretty good to me! It comes at a quite convenient time, as we'll have AD Hessians when #244 is merged :)

docs/src/algo/newton_trust_region.jl
rgiordan: Good catch, I've changed it to "method" in the next push.

docs/src/algo/newton_trust_region.jl
rgiordan: Yes indeed (though with the sign switched relative to how you have it).  I got mixed up by the difference between N&W's notation and the code.  :P  (Note also in the code that the numerator is f(x_k) - f(x_{k+1}), in contrast to the formula in the doc.)  Good catch.

rgiordan:
How can I best test that the docs are correct?  (I'm not familiar with Documenter so I thought I would just ask.)

docs/src/algo/newton_trust_region.md
```diff
@@ -0,0 +1,57 @@
...
+
+Finally, we only accept a point if its decrease is appreciable compared to the quadratic approximation.  Specifically, a step is only accepted $\rho_k > \eta$.  As long as we choose $\eta$ to be less than $\rho_{lower}$, we will shrink the trust region whenever we reject a step.  Eventually, if the objective function is locally quadratic, $\Delta_k$ will become small enough that a quadratic approximation will be accurate enough to make progress again.
+
+## Example
+
+```julia
+using Optim
+prob = Optim.UnconstrainedProblems.examples["Rosenbrock"];
+ddf = TwiceDifferentiableFunction(prob.f, prob.g!,prob.h!)
+res = Optim.optimize(prob.f, prob.g!, prob.h!, prob.initial_x, method=NewtonTrustRegion())
```
anriseth: Should it be using `ddf` in the argument to `optimize` here?

rgiordan: Thank you, yes.  Actually, let me just get rid of ddf.

pkofod:
If you have mkdocs installed, you can browse to the directory and write `mkdocs serve`. Other than that, I'm not really sure. I wish it was easier to preview it for PRs.

rgiordan:
It fails with the below error.  Is there maybe an additional step that I'm
missing?

A little README.md in this directory would probably help contributors like
me a lot.

```
$ mkdocs serve
Traceback (most recent call last):
  File "/usr/bin/mkdocs", line 9, in <module>
    load_entry_point('mkdocs==0.11.1', 'console_scripts', 'mkdocs')()
  File "/usr/lib/python3/dist-packages/mkdocs/main.py", line 60, in run_main
    main(cmd, args=sys.argv[2:], options=dict(opts))
  File "/usr/lib/python3/dist-packages/mkdocs/main.py", line 33, in main
    serve(config, options=options)
  File "/usr/lib/python3/dist-packages/mkdocs/serve.py", line 82, in serve
    build(config, live_server=True)
  File "/usr/lib/python3/dist-packages/mkdocs/build.py", line 267, in build
    build_pages(config)
  File "/usr/lib/python3/dist-packages/mkdocs/build.py", line 207, in
build_pages
    input_content = io.open(input_path, 'r', encoding="utf_8").read()
FileNotFoundError: [Errno 2] No such file or directory: 'build/Home'
```

pkofod:
You probably have to run `julia make.jl` first, as the `build` folder is in `.gitignore`

pkofod:
> A little README.md in this directory would probably help contributors like me a lot.

I'll add it to the contributors part of the docs. Good to get some feedback on this process :)

rgiordan:
Yes, I had guessed that, but it still doesn't work.  The output is this:

```
Optim.jl/docs$ julia make.jl
Documenter: setting up build directory.
Documenter: redirecting output streams.
Documenter: expanding markdown templates.
Documenter: building cross-references.
Documenter: running document checks.
Documenter: restoring output streams.
 !! Skipped doctesting.
Documenter: populating indices.
Documenter: rendering document.
Documenter: copying assets to build directory.
Documenter: skipping docs deployment.
rgiordan@kokshetau:~/Documents/git_repos/Optim.jl/docs$ mkdocs serve
Traceback (most recent call last):
  File "/usr/bin/mkdocs", line 9, in <module>
    load_entry_point('mkdocs==0.11.1', 'console_scripts', 'mkdocs')()
  File "/usr/lib/python3/dist-packages/mkdocs/main.py", line 60, in run_main
    main(cmd, args=sys.argv[2:], options=dict(opts))
  File "/usr/lib/python3/dist-packages/mkdocs/main.py", line 33, in main
    serve(config, options=options)
  File "/usr/lib/python3/dist-packages/mkdocs/serve.py", line 82, in serve
    build(config, live_server=True)
  File "/usr/lib/python3/dist-packages/mkdocs/build.py", line 267, in build
    build_pages(config)
  File "/usr/lib/python3/dist-packages/mkdocs/build.py", line 207, in
build_pages
    input_content = io.open(input_path, 'r', encoding="utf_8").read()
FileNotFoundError: [Errno 2] No such file or directory: 'build/Home'
```

pkofod:
I don't really know what the problem is. Maybe python related? I have zero experience with python, but I always end up tracking down errors for hours when running into stuff like this. Anyway, it runs locally, I've sent you a pdf with the page to your e-mail.

rgiordan:
The emailed pdf looked good to me.  Sorry I didn't have the bandwidth this afternoon to work through the Documenter bugs.

pkofod:
Did you change anything after I sent you the pdf? The line just before "Conversely, if" looks odd in the pdf, but the source looks fine.

If you rebase this, I think we're good to go!

rgiordan:
I've rebased and pushed.

The pdf does look strange, as if there were an extra $.  I also agree the
source looks fine.  I'm sorry I can't build myself and check -- could you
re-build the docs and see if the problem is still there?  If so, maybe I
need to get the documentation building working to debug this.

pkofod:
There seems to be some git-weirdness going on here. You now have all the latest commits from other pull requests also in here? Tests didn't pass either.

rgiordan:
Yes indeed, looking into it now.

pkofod:
Also, I'll fix the doc problems if they show op in master, don't worry about that.

rgiordan:
I'm not sure what happened, but the rebase seems to have gone wrong somehow.

Would you have any objection to just making a new branch and new PR?  My git-fu is not strong enough to be comfortable rewinding history.  (Unless you have specific advice on how to do that.)

rgiordan:
I created https://github.com/JuliaOpt/Optim.jl/pull/245.  Sorry about the trouble.  I wish I understood what went wrong...

PR #242, opened 2016-07-12
Define endof for OptimizationTrace

ahwillia:
I tried to access `trace[end]` in some code and ran into this error. I opened a PR since I don't see any reason to not add this.


pkofod:
Thanks, I can see the use for this. Could you add a test for it?

ahwillia:
Done. I also added `Base.length` which could be useful.

pkofod:
Thanks!

PR #244, opened 2016-07-12
Switch to ForwardDiff for AD

### LINKED ISSUE #228: Use ForwardDiff instead of DualNumbers for AD
ForwardDiff should have better performance and also support computing Hessians. I recently upgraded NLsolve to use the new ForwardDiff (https://github.com/EconForge/NLsolve.jl/pull/54) so I don't mind doing it myself here but with JuliaCon + vacation afterwards it might take a bit. 

Any opinions?

pkofod:
This PR replaces the "manual" DualNumbers automatic differentiation with the functionality in ForwardDiff. Additionally, it add AD functionality for Newton's method.

Fixes #228 @KristofferC 

Todo
- [x] test all the "hessian" signatures (optimize(f, g!, ..) etc.)


src/optimize.jl
```diff
@@ -145,11 +145,68 @@ function optimize(f::Function,
     if !options.autodiff
         d = DifferentiableFunction(f)
     else
-        d = Optim.autodiff(f, eltype(initial_x), length(initial_x))
+        g! = (x, out) -> ForwardDiff.gradient!(out, f, x)
+
+        function fg!(x, out)
```
KristofferC: This is not the most performant way of doing it. You can extract the function value from the gradient call. You can see how I did it in NLsolve.

pkofod: I see, I will switch to `GradientResult`.

src/optimize.jl
```diff
@@ -138,18 +138,77 @@ function optimize(f::Function,
...
 
+function optimize(f::Function,
+                  g!::Function,
+                  initial_x::Array,
+                  method::Newton,
+                  options::OptimizationOptions)
+    if !options.autodiff
+        error("No Hessian was provided. Either provide a Hessian, set autodiff = true in the OptimizationOptions if applicable, or choose a solver that doesn't require a Hessian.")
+    else
+        function fg!(x, out)
```
pkofod: Should we use AD on the gradient and fg! if autodiff=true, but a gradient was provided, or do it like I've done here? I basically did what Optim would do in the gradient based methods if only f and g! is provided for fg!, and then construct h! with ForwardDiff.

src/optimize.jl
pkofod: Thanks for the tip. Done.

mlubin:
CC @jrevels

src/optimize.jl
```diff
@@ -138,18 +138,77 @@ function optimize(f::Function,
...
+                  initial_x::Array{T},
+                  method::Newton,
+                  options::OptimizationOptions)
+    if !options.autodiff
+        error("No gradient or Hessian was provided. Either provide a gradient and Hessian, set autodiff = true in the OptimizationOptions if applicable, or choose a solver that doesn't require a Hessian.")
+    else
+        g!(x, out) = ForwardDiff.gradient!(out, f, x)
+
+        function fg!(x, out)
+            gr_res = ForwardDiff.GradientResult(zero(T),out)
```
jrevels: `zero(T)` -> `zero(eltype(out))`, in case the `f`'s input and output have different number types? Unless, of course, Optim already assumes that this won't be the case.

jrevels:
LGTM. Would it make sense to take advantage of ForwardDiff's `HessianResult` type for simultaneously getting the value/gradient/hessian for `TwiceDifferentiableFunction`s? Might not be useful if these functions are just being passed as oracles to solvers, though.

src/optimize.jl
pkofod: > zero(T) -> zero(eltype(out)), in case the f's input and output have different number types? Unless, of course, Optim already assumes that this won't be the case.

Usually the array that is going to be out is going to be initialized as `Array(T, n)` where `T = eltype(initial_x)` and `n = length(initial_x)` (https://github.com/JuliaOpt/Optim.jl/blob/48a4d82a6db8a503ec539167b2e70f2ba29c5b46/src/newton.jl#L58) So it should be fine, but I had actually originally written it as you do. I think it's safe to assume `T == eltype(out)` throughout Optim though.

pkofod:
> if these functions are just being passed as **oracles** to solvers, though.

What does oracles mean in this context? Sorry for my ignorance :)

jrevels:
No worries, I'm new to this stuff myself. I'm using "oracle" to refer to a function that is passed to solvers, which the solvers can then call to retrieve derivative information. I'm not sure whether that's how Optim works, though, or if derivative information is propagated some other way.

pkofod:
> No worries, I'm new to this stuff myself. I'm using "oracle" to refer to a function that is passed to solvers, which the solvers can then call to retrieve derivative information. I'm not sure whether that's how Optim works, though, or if derivative information is propagated some other way.

Ah, an object that has all the answers? :) Well, I think it's something like that. The "TwiceDifferentiableFunction" is passed to a solver that can, in any order and frequency, call f, g!, fg!, or h!. I think the final design choices are yet to be done here, so I think we should just write it down as a note, that this "feature" is available in ForwardDiff.

PR #245, opened 2016-07-15
Newton trust region rebase

rgiordan:
Rather than attempt to fix https://github.com/JuliaOpt/Optim.jl/pull/238, here is a new branch without the messed up commits.


pkofod:
Good idea :+1: sometimes that's just easier!

pkofod:
Merged! Thanks for this great PR!

rgiordan:
Thanks for your help (and patience), and for Optim.jl in general.  :)

PR #247, opened 2016-07-16
Placing the HOW section after WHAT and WHY.

Ace139:
I as a user would like to know first what is it, then install the package. :)


pkofod:
Fine by me, thanks.

Issue #255, opened 2016-08-09
Should OptimizationTrace be an AbstractArray?


ahwillia:
I wanted to be able to gather the value of the objective function along the optimization path:

```julia
result = optimize(f, g!, x0; store_trace=true)
mytrace = Optim.trace(result)
objective_history = [ t.value for t in mytrace ]
```

But I get:

```
ERROR: MethodError: no method matching start(::Optim.OptimizationTrace)
```

All of these methods would be defined by default if we declare `OptimizationTrace` to be an `AbstractArray`.

Or do we even need a special type? Why not just have `Optim.trace(result)` return a `Array{OptimizationState,1}`? Pretty printing purposes?

**Also Related:** https://github.com/JuliaOpt/Optim.jl/pull/242

**Edit:**  I suppose the way to do this currently is:

```julia
objective_history = [ t.value for t in mytrace.states ]
```

johnmyleswhite:
We should almost certainly do this, but what `AbstractArray` means is pretty vague. Your first example only really requires it implement an Iterator-like trait.

pkofod:
I don't really have a good reason why we shouldn't ditch `OptimizationTrace{T}` in favour of `Vector{OptimizationState{T}}` where `T<:Optimizer`. Essentially, all we need for "pretty printing" is the last part. All instances of `OptimizationTrace` should simply be replaced by `Vector{OptimizationState}` I guess. :+1: from here

ahwillia:
I'll open a PR later today to explore further.

PR #257, opened 2016-08-09
Simplify OptimizationTrace as a Vector{T<:OptimizationState}

ahwillia:
Here is a sketch of how I would implement `OptimizationTrace`. I didn't add any new tests, but it looks like the appropriate checks are already in place.

The only thing that is slightly funky is that `optimize` end up initializing the `OptimizationTrace` like this:

```julia
mo = NelderMead()
OptimizationTrace{typeof(mo)}()
```

There might be something more elegant, but this seems okay to me at least. Let me know your thoughts.

### DIFF
--- a/src/accelerated_gradient_descent.jl
+++ b/src/accelerated_gradient_descent.jl
@@ -80,7 +80,7 @@ function optimize{T}(d::DifferentiableFunction,
     lsr = LineSearchResults(T)
 
     # Trace the history of states visited
-    tr = OptimizationTrace(mo)
+    tr = OptimizationTrace{typeof(mo)}()
     tracing = o.store_trace || o.show_trace || o.extended_trace || o.callback != nothing
     @agdtrace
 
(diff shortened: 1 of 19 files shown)


pkofod:
> The only thing that is slightly funky is that optimize end up initializing the OptimizationTrace like this:

I don't think it's too funky.

Looks good, but I will go over it again tomorrow.

ahwillia:
Forgot to cross-reference the relevant issue: https://github.com/JuliaOpt/Optim.jl/issues/255

src/accelerated_gradient_descent.jl
```diff
@@ -80,7 +80,7 @@ function optimize{T}(d::DifferentiableFunction,
     lsr = LineSearchResults(T)
 
     # Trace the history of states visited
-    tr = OptimizationTrace(mo)
+    tr = OptimizationTrace{typeof(mo)}()
```
Evizero: I think the old interface could be kept by creating a outer constructor for `OptimizationTrace` like

```julia
OptimizationTrace{T<:OptimizerOrWhatever}(mo::T) = OptimizationTrace{T}()
```

pkofod:
Thank you for doing this! It's cleaner than it was.

PR #260, opened 2016-08-11
Cleanup of README.md, optimize.jl, and deprecation of NelderMead cons…

pkofod:
…tructor.

All information from the `README.md` *should* be in the docs, but I will double check. In the future, we'll have a very simple readme, and links to the docs. The `stable` link here is not live yet, but will be after we tag again (very soon; tomorrow?).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,452 +1,63 @@
-
-[![](https://img.shields.io/badge/docs-latest-blue.svg)](https://juliaopt.github.io/Optim.jl/latest)
-
-*This it the development branch of Optim.jl. Please visit [this branch](https://github.com/JuliaOpt/Optim.jl/tree/v0.4.5) to find the README.md belonging to the latest official release of Optim.jl*
-
-
 Optim.jl
 ========
 
-[![Join the chat at https://gitter.im/JuliaOpt/Optim.jl](https://badges.gitter.im/JuliaOpt/Optim.jl.svg)](https://gitter.im/JuliaOpt/Optim.jl?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+Univariate and multivariate optimization in Julia.
 
-[![Optim](http://pkg.julialang.org/badges/Optim_0.3.svg)](http://pkg.julialang.org/?pkg=Optim&ver=0.3)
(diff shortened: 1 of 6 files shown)


Evizero:
It might make more sense to have the installation before the example. Or maybe I am just used to that.

README.md
```diff
@@ -1,452 +1,51 @@
...
-* Linear conjugate gradients
-* L-BFGS-B (note that this functionality is already available in fminbox)
-
-# Citations
-
-W. W. Hager and H. Zhang (2006) Algorithm 851: CG_DESCENT, a conjugate gradient method with guaranteed descent. ACM Transactions on Mathematical Software 32: 113-137.
-
-R. P. Brent (2002) Algorithms for Minimization Without Derivatives. Dover reedition.
+[![](https://img.shields.io/badge/docs-stable-blue.svg)](https://juliaopt.github.io/Optim.jl/stable)
+[![](https://img.shields.io/badge/docs-latest-blue.svg)](https://juliaopt.github.io/Optim.jl/latest)
```
Evizero: duplicate or on purpose?

Evizero:
A personal opinion on badges layout. I like how Documenter.jl does it with a table: https://github.com/JuliaDocs/Documenter.jl

Very visually pleasing

pkofod:
I agree on the table, I'll fix it! On the ordering. It was recently changed in the docs because it was argued that it was "more important" to know what the package was about that how it is installed. I can go either way.

tbreloff:
My vote would be example first.  People only care about installing if they actually want to use the package...

mlubin:
I think the "installation instructions first" trend started about the same time as when the package manager was created and very few people knew how to install packages. It's probably less important these days.

pkofod:
I'll let it stay, but I guess it really only is important if a package is not tagged, and it's important to explicitly state that you need to `Pkg.clone("Optim")`. It's still useful for new users of Julia, so I'll let it stay in the less prominent position.

pkofod:
Fixed. Now I just need to make sure we didn't lose any information.

Evizero:
Nice! I notice there is no travis badge for master. Is this on purpose?

pkofod:
Don't think we ever had one. I'll add it.

pkofod:
@Evizero Better? :)

pkofod:
I added a small section on using automatic differentiation. It has no performance comparisons, I just wanted to show how it works, and that you can obtain Hessians now.

Issue #254, opened 2016-08-08
Refactoring iterative methods with an update!() function


pkofod:
There was a discussion over at https://gitter.im/JuliaML/chat about how to train models in the JuliaML project. It was mentioned that Optim might be useful. However, when training, we often do not "run until convergence" in the sense we do it here. We might want to calculate different things in between, and we might not want to end it when the gradient is below some threshold. The idea that was discussed was to refactor Optim solvers such that we get something along the lines of 

``` julia
function optimize(...)
    s = initialize_state(mo, o)
    while !converged && iteration < o.iteration
        update!(s, ...)
        ... = converged(....)
    end
    return MultivariateOptimizationResults(...)
end
```

and then each solver has to implement `initialize_state` and `update!` instead of implementing an `optimize(df, initial_x, mo, o)` method. We could also add a callback after update to allow for more flexible stopping.

Any thoughts on this? I've had the request before from other people.

johnmyleswhite:
I've done a lot of this kind of work with rigged up local copies of Optim. I think you need to do this and also provide a mechanism for providing all of the intermediate data structures, so you don't waste time allocating them on each iteration. Definitely worth doing.

KristofferC:
Sounds a bit like https://github.com/spencerlyon2/IterationManagers.jl

mlubin:
This is generally useful. Would have erased the need for this [juliacon talk](https://www.youtube.com/watch?v=qAD0ZXriBJs).

pkofod:
Thanks for the quick feedback/comments. I think it's clear that this is happening. I hope it will clean up some code, as well as increase the usefulness of the package!

> I've done a lot of this kind of work with rigged up local copies of Optim. I think you need to do this and also provide a mechanism for providing all of the intermediate data structures, so you don't waste time allocating them on each iteration. Definitely worth doing.

Great; it will be next on my list, but for v0.7. I'll cook up a PR, but I'll be very happy to take ideas and criticism from people who've used this kind of functionality.

> Sounds a bit like https://github.com/spencerlyon2/IterationManagers.jl

Yes, it's kind of related to that. 

> This is generally useful. Would have erased the need for this juliacon talk.
I didn't see that talk all the way through, maybe I should before I start working on this :)

cortner:
Kind of a side issue, but something worth thinking about? 

[...]

PR #262, opened 2016-08-12
Assess g_convergence before applying algorithm

### LINKED ISSUE #261: Linesearch fails when derivative is zero
I have a problem using (L)BFGS when the initial guess is the exact minimiser. The linesearch algorithm does not seem to like this:
```julia
f(x) = x[1]^2
optimize(f, [0.0], LBFGS(), OptimizationOptions(autodiff=true, show_trace=true, extended_trace=true))
```
Prints:
```julia
Iter     Function value   Gradient norm 
     0     0.000000e+00     0.000000e+00
 * Current step size: 1.0
 * g(x): [0.0]
 * x: [0.0]
ERROR: Search direction is not a direction of descent; this error may indicate that user-provided derivatives are inaccurate. (dphia = 0.000000; dphib = 0.000000)
 in error at ./error.jl:21
 in secant2! at /home/riseth/.julia/v0.4/Optim/src/linesearch/hz_linesearch.jl:412
 in hz_linesearch! at /home/riseth/.julia/v0.4/Optim/src/linesearch/hz_linesearch.jl:321
 in hz_linesearch! at /home/riseth/.julia/v0.4/Optim/src/linesearch/hz_linesearch.jl:176 (repeats 10 times)
 in optimize at /home/riseth/.julia/v0.4/Optim/src/l_bfgs.jl:186
 in optimize at /home/riseth/.julia/v0.4/Optim/src/optimize.jl:150
```

anriseth:
This would close #261.

### DIFF
--- a/src/accelerated_gradient_descent.jl
+++ b/src/accelerated_gradient_descent.jl
@@ -85,10 +85,11 @@ function optimize{T}(d::DifferentiableFunction,
     @agdtrace
 
     # Assess types of convergence
-    x_converged, f_converged, g_converged = false, false, false
+    x_converged, f_converged = false, false
+    g_converged = vecnorm(g, Inf) < o.g_tol
 
     # Iterate until convergence
-    converged = false
+    converged = g_converged
     while !converged && iteration < o.iterations
         # Increment the number of steps we've had to perform
         iteration += 1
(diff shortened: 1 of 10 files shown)


pkofod:
Thank you very much. By the way, I think this is a good argument in favour of refactoring of the code at #254 . It shouldn't be necessary to change 7 files to do this.

anriseth:
I just realised that this PR would claim convergence if you start at a local maximum. Does anyone know how other optimisation packages handle this?

pkofod:
That is true. I'm not sure. You could try starting an NLOpt-solver at a local max to check.

PR #264, opened 2016-08-15
Fix AD for NewtonTrustRegion.

pkofod:


pkofod:
Trivial addition; merging.

Issue #263, opened 2016-08-15
When should we throw errors versus return a result that flags a potentially bad result?


johnmyleswhite:
Continuing discussion in #261, we should decide when we will throw an error and when we will return a result with a no-convergence flag. One proposal was to avoid throwing errors for line search failures or other problematic states (e.g. non-descent directions) and instead return an object that has an internal flags that prevent direct access to results without checking their validity. Other options are to raise warnings but still return results or to continue throwing errors and not returning results.

pkofod:
As far as I could understand, you main concern is that people write 
```julia
using Optim
...
result = optimize(f, inital_x)
my_function_that_depends_on_result(Optim.minimizer(result), other_vars...)
```
but if there's a line search error they don't realize that `optimize` didn't actually succeed. Maybe this is even true for `converged == false`.

My main thought was that there shouldn't be any `Error`s that halts `optimize` altogether. Instead it could break the while loop, and report to the user that: this is the trace I got, this is the value of `x`, and so on. We could then set a flag in `MultivariateOptimizationResults` that indicated if a field could easily be retrieved or not. Warnings are not going to be enough, as code might be run in a script where no one is monitoring the output.

The `unsafe`-api seems almost too much, but maybe it really *is* **important** to make sure that `Optim.minimizer(result)` does not simply give a (meaningless) answer in the non-convergence case.

johnmyleswhite:
Sounds like we generally agree. To try to clarify what I was saying, here's how I see things:

* We should not throw errors as a way of indicating that the results of the optimization process might be wrong. But `optimize` can still throw errors: we just don't add additional errors beyond the basic ones that are intrinsic to Julia code (e.g. array bounds errors). It sounds like we fully agree on this point.

* We should not just warn people when their results are likely to be wrong. In particular, the API for accessing fields like the minimizer should not return a result that we know to be wrong. We could return vectors containing `NaN`, but it seems more informative to actually throw an error when we know the answer is suspect. From my perspective, this is equivalent to moving the errors we're currently throwing to a better place in the control flow: if we throw errors when you already have access to a results object, you can use trace information to debug things in a way that's generally impossible now.

[...]

PR #266, opened 2016-08-16
[WIP] [RFC!] Refactoring...

pkofod:
A first attempt, proof of concept, work in progress, and whatever other word you might want to use for temporary or unfinished.

I didn't want to go through the entire package without any comments, so I would love to hear everyone's input.

There are some weird things in there, and specifically a few new types are simply invented for dispatch reasons. They will go again.

Right now, I'm seeing `state.` here, `state.` there, `state.`s everywhere! Should we use something like the macros in Parameters such that we can omit the `state.`s?

I'll keep on working on it, but criticism is very welcome.

TODO

- [x] Add all multivariate solvers
- [x] handle generic fields (includes remove the `method_name` in favour of a default field in the types)
- [x] cleanup (white space, `Refactored` etc)
- [x] move callback and uncomment
- [ ] `state.`s, `state.`s everywhere!


tbreloff:
I'm not as familiar with the codebase as others, but from my quick review I like what I see.  :+1:

KristofferC:
More removed lines than added? Ship it!

pkofod:
@KristofferC I also sneaked in the transition from macros to functions in `trace` that we discussed a while ago.

src/bfgs.jl
```diff
@@ -1,196 +1,150 @@
...
+    I::Array{T}
+    invH::Array{T}
+    f_x::T
+    f_x_previous::T
+    x_ls::Array{T}
+    g_ls::Array{T}
+    alpha::T
+    mayterminate::Bool
+    f_calls::Int64
+    g_calls::Int64
```
Evizero: Things like `f_calls` etc seem generic enough to not be in each solver specific state.
or alternatively one could create a little macro that adds these common fields to the current type

```julia
type BFGSState{T}
    bfgs_specific_variable::Int64

    @add_generic_fields()
end
```

src/optimize.jl
```diff
@@ -200,21 +203,70 @@ function optimize(d::DifferentiableFunction,
...
+
+    tr = OptimizationTrace{typeof(method)}()
+    tracing = options.store_trace || options.show_trace || options.extended_trace || options.callback != nothing
+
+    x_converged, f_converged = false, false
+    g_converged = if typeof(method) <: NelderMead
+        nmobjective(state.f_simplex, state.m, state.n) < options.g_tol
+        else
+            vecnorm(state.g, Inf) < options.g_tol
+        end
```
Evizero: some indentation artefact here

src/bfgs.jl
pkofod: Sure! *At least* all the fields for multivariateoptimizationoptions Will be the same.

pkofod:
Added `ParticleSwarm`. I just need `SimulatedAnnealing` and the two `Newton`'s.

Tests are running longer now, mostly because we recently added more tests I think. One of the `NelderMead` tests might also be part of it. This means that I've had to reset the tests a few time due to time outs. Let's see if it continues to be a problem, and we might have to do something about it.

src/accelerated_gradient_descent.jl
```diff
@@ -35,127 +14,86 @@ end
...
+      f_x::T
+      f_x_previous::T
+      s::Array{T}
+      x_ls::Array{T}
+      g_ls::Array{T}
+      alpha::T
+      mayterminate::Bool
+      f_calls::Int64
+      g_calls::Int64
+      lsr
```
Evizero: Is the type of this not known? only asking since you went through the trouble of using leaf types everywhere else. a little further down it is initialized with `LineSearchResults(T)`

pkofod: I may have forgotten some annotations in the process :) thanks for reading the code so thoroughly.

src/simulated_annealing.jl
```diff
@@ -42,88 +21,70 @@ SimulatedAnnealing(; neighbor!::Function = default_neighbor!,
...
+  n::Int64
+  iteration::Int64
+  x_current::Array{T}
+  x_proposal
+  x
+  f_x
+  f_x_current::T
+  f_proposal::T
+  f_calls::Int64
+  g_calls::Int64
```
Evizero: some indentation artefacts here

src/simulated_annealing.jl
```diff
@@ -42,88 +21,70 @@ SimulatedAnnealing(; neighbor!::Function = default_neighbor!,
...
-    f_x = cost(x)
-    f_calls += 1
+    f_x = f(initial_x)
+    f_calls = 1
 
     # Store the best state ever visited
-    best_x = copy(x)
+    best_x = copy(initial_x)
     best_f_x = f_x
+    iteration = 1
```
Evizero: I doesn't look like this variable is used anywhere. maybe an artefact of something you tried but decided against?

pkofod: indeed, that is the `1` just after the `n` just below, thanks !

pkofod:
fails because nelder mead is slow, maybe I should drop that test again.

pkofod:
The fields that are in all states are now added with a macro. This is what you meant right @Evizero ? What do you think @johnmyleswhite (and others). Is it a few lines fewer at the cost of a decrease in readability?

Evizero:
That was what I meant, yes. If it is worth the reduction of code repetition I can't say, but it seems worth considering.

pkofod:
> That was what I meant, yes. If it is worth the reduction of code repetition I can't say, but it seems worth considering.

With the general refactoring + the macros (the `initialize_linesearch` could be a function I guess) there's a pretty nice +/- line ratio, at least.

src/accelerated_gradient_descent.jl
```diff
@@ -35,127 +14,75 @@ end
...
-        @simd for i in 1:n
-            @inbounds s[i] = -g[i]
-        end
-
-        # Refresh the line search cache
-        dphi0 = vecdot(g, s)
-        clear!(lsr)
-        push!(lsr, zero(T), f_x, dphi0)
+type AcceleratedGradientDescentState{T}
+    @add_generic_fields()
```
johnmyleswhite: Can I suggest we use composition rather than macros? In that approach, you'd create a new type called `GenericOptimizationState` and place that type inside of each of these types.

src/accelerated_gradient_descent.jl
```diff
@@ -35,127 +14,75 @@ end
...
+    iteration::Int64
+    x_previous::Array{T}
```
johnmyleswhite: Do we enforce the invariant that the whole computation is done in constant precision for `x`, `f(x)` and `g(x)`? My impression is that we've been somewhat cavalier with this in the past, although I'm glad to see us trying to make some progress here.

src/bfgs.jl
```diff
@@ -1,196 +1,119 @@
...
+    invH::Array{T}
+    s::Array{T}
+    @add_linesearch_fields()
+end
 
-    # Store f(x) in f_x
-    f_x_previous, f_x = NaN, d.fg!(x, g)
-    f_calls, g_calls = f_calls + 1, g_calls + 1
-    copy!(g_previous, g)
+function initialize_state{T}(method::BFGS, options, d, initial_x::Array{T})
```
johnmyleswhite: Extremely pedantic nitpicking: should this be called `initial_state` since it doesn't mutate an existing state object, but creates a new state object?

pkofod: I think you're right. I'll change it.

src/accelerated_gradient_descent.jl
pkofod: We could do that. If I had to try to defend it, it would probably be that it adds a layer of dots (`state.generic.x`) but I guess this can be handled with an "unpack"-like functionality as in Parameters.

src/accelerated_gradient_descent.jl
pkofod: I am not sure we have enforced it earlier, I just figured it made sense.

ahwillia:
Still really excited about this! One thing I'd really like to try is to build on top of this to optimize problems with constraints.

I'm not sure if this will end up being a dead end, but I decided to start experiment by doing projected gradient descent to solve a non-negative least squares problem (a terrible way to solve it, but a useful toy problem nonetheless).

```julia
using Optim

const A = rand(10,10)
const b = rand(10)

f(x) = 0.5*sumabs2(A*x - b)
d = DifferentiableFunction(f)
x0 = randn(10)

# project onto non-negative orthant
project!(x) = map!(z-> z<0 ? 0.0 : z, x)

gd = GradientDescent()
o = OptimizationOptions()

s = Optim.initial_state(gd, o, d, x0)

for iter = 1:iterations
    Optim.update!(d, s, gd)
    project!(s.x)
end
```

Right now I get this error:

```
ERROR: LoadError: AssertionError: c > 0
 in hz_linesearch!(::Optim.DifferentiableFunction, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Optim.LineSearchResults{Float64}, ::Float64, ::Bool, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Int64, ::Float64, ::Int64, ::Int64) at /Users/alex/.julia/v0.5/Optim/src/linesearch/hz_linesearch.jl:186
 in hz_linesearch!(::Optim.DifferentiableFunction, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Optim.LineSearchResults{Float64}, ::Float64, ::Bool, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Int64, ::Float64, ::Int64) at /Users/alex/.julia/v0.5/Optim/src/linesearch/hz_linesearch.jl:176 (repeats 4 times)
 in update!(::Optim.DifferentiableFunction, ::Optim.GradientDescentState{Float64}, ::Optim.GradientDescent{Void}) at /Users/alex/.julia/v0.5/Optim/src/gradient_descent.jl:52
 in macro expansion; at /Users/alex/.julia/v0.5/Optim/examples/projected_gd.jl:23 [inlined]
 in macro expansion; at /Users/alex/.julia/v0.5/ProgressMeter/src/ProgressMeter.jl:473 [inlined]
 in anonymous at ./<missing>:?
 in include_from_node1(::String) at ./loading.jl:426
while loading /Users/alex/.julia/v0.5/Optim/examples/projected_gd.jl, in expression starting on line 22
```

I'll keep playing around with this, but am curious to hear feedback on whether this will mesh well with the intended approach here.

(I also want to mention https://github.com/JuliaOpt/Optim.jl/issues/254 so that earlier discussion is linked to this PR)

pkofod:
Great to see a simple example of how this could be useful outside of Optim.

I actually found a place in NelderMead where I hadn't added the `state.` from playing around with your example :) (apparently none of our tests causes a shrink!)

I'll see if I can further investigate it later.

src/gradient_descent.jl
```diff
@@ -32,118 +8,60 @@ GradientDescent(; linesearch!::Function = hz_linesearch!,
...
+    method.linesearch!(d, state.x, state.s, state.x_ls, state.g_ls,
+                       state.lsr, state.alpha, state.mayterminate)
+    state.f_calls, state.g_calls = state.f_calls + f_update, state.g_calls + g_update
+
+    # Maintain a record of previous position
+    copy!(state.x_previous, state.x)
+
+    # Update current position # x = x + alpha * s
+    LinAlg.axpy!(state.alpha, state.s, state.x)
+
```
ahwillia: This is where I would need to add the projection step to do projected gradient descent.

ahwillia:
Changing the inner loop like this seems to work:

```julia
s = Optim.initial_state(gd, o, d, x0)
for iter = 1:iterations
    Optim.update!(d, s, gd)
    s = Optim.initial_state(gd, o, d, project!(s.x))
end
```

Looks like the problem is that `update!` calculates the gradient at the end of the function so I need to manually recalculate the gradient before moving to the next loop iteration. I haven't given this much thought, but would it make sense for `update!` to calculate the gradient *before* doing linesearch and moving the parameters?

johnmyleswhite:
>  I haven't given this much thought, but would it make sense for update! to calculate the gradient before doing linesearch and moving the parameters?

I think this could be made to work if we want to make `update!` idempotent in the sense that it will not modify the current state if it detects convergence. But you also need to decide if you want to check for convergence at all.

pkofod:
> Changing the inner loop like this seems to work (though there is a lot of unnecessary overhead):
> 
> s = Optim.initial_state(gd, o, d, x0)
> for iter = 1:iterations
>     Optim.update!(d, s, gd)
>     s = Optim.initial_state(gd, o, d, project!(s.x))
> end
> 
> Looks like the problem is that update! calculates the gradient at the end of the function so I need to manually recalculate the gradient before moving to the next loop iteration. I haven't given this much thought, but would it make sense for update! to calculate the gradient before doing linesearch and moving the parameters?

Again, I think your example really shows why this refactoring is worth it, but we need to think about how to do this the best way.

We could move the gradient calls to the top. What is the issue with that? Well, then we have a superfluous gradient calculation in the first iteration, as we have already calculated the gradient to do the pre-iteration convergence check. If we don't do it, we revive the "linesearch error if initial point is a stationary point". Instead, we could remove the initial convergence check, and then turn the line search *error* into a line search warning. We could then add a flag, such that when the warning is thrown, a flag is changed to `true`, and as soon as `update!` is over, the major loop terminates.

This would then require some form of #263 to be implemented. I think that might work.

ahwillia:
> Instead, we could remove the initial convergence check, and then turn the line search error into a line search warning.

:+1: this would be great!

ahwillia:
I'm hesitant to pollute this PR with more half-baked examples, but I think coordinate descent and other sequential/alternating optimization schemes would fit quite nicely with how things are being refactored. I put a simple example (fitting PCA) here: https://gist.github.com/ahwillia/8d04471daf0a9a12794107dec8fe3884

Again, we'd need `update!` to recalculate gradients at the top of the function call for this to work. So I think there is a strong argument for doing this. I suppose an alternative would be have `update!` only change `x`, and leave recalculating gradients/hessians to another function:

```julia
while !converged
    update!(x, ....)
    calcgrads!(x, ....)
end
```

This alternative maybe be more flexible, but I can't think of a good use case that requires it.

pkofod:
> :+1: this would be great!

I promise I'll get back to this, I'm just a little tied up in some other code these days... you know, work and all. I'll make sure the gradient is evaluated as the first thing instead of the last. It doesn't make sense to go into the update and simply say "oh well, let's hope someone has updated the gradient I'm about to use".

pkofod:
It's a little bit annoying to move the gradient calls up, but I do realize why it should be done. The reason why it is annoying is that for stuff like BFGS, it is cleaner to update the inverse hessian approximation at the end, because then we can simply provide an initial inverse hessian, and then everything follows nicely. Correct me if I'm wrong, but by moving the gradient call up, we **have** to have a `if iteration > 1` control flow statement around the hessian update, correct?

pkofod:
Callbacks now work again, and you can stop using something like
```julia
optimize(f, zeros(2), NelderMead(), OptimizationOptions(callback = x -> x.iteration == 5 ? true : false))
```
or early stopping based on time since start
```julia
problem = Optim.UnconstrainedProblems.examples["Rosenbrock"]

f = problem.f
initial_x = problem.initial_x

function slow(x)
    sleep(0.1)
    f(x)
end

start_time = time()

optimize(slow, zeros(2), NelderMead(), OptimizationOptions(callback = x -> time()-start_time > 3 ? true : false))
```

ahwillia:
Sorry I missed this until now. Ultimately I don't really feel strongly about most of these points either way. But for what its worth:

> should we then check for convergence in `update!`? That seems unnatural...

Why unnatural? If it can be done with little overhead I think having `update!` return true/false would enable some nice things:

```julia
for iter in iterations
    # break loop if convergence detected
    Optim.update!(x, state, solver) && break
end
```

Although you'd also have to pass `OptimizationOptions` if you decided to do this...

> Would you be happy @ahwillia with I simply made the "update and store new gradient" an opt out feature?

So something like?

```julia
update!(x, state, solver; update_gradient = true) # default 
update!(x, state, solver; update_gradient = false) # opt out
```

I could live with this, but I think I would prefer splitting this into different functions:

```julia
update_state!(x, state, solver) # assumes current gradients/hessians are fresh
update_grad!(x, state, solver)
update_hess!(x, state, solver)

function update!(x, state, solver)
    update_state!(x, state, solver)
    update_grad!(x, state, solver)
    update_hess!(x, state, solver)
end
```

I guess `update_hess!(...)` would be a no-op for things like gradient descent and `update_grad!(...)` would be a no-op for Nelder Mead.

This seems like it would give me more flexibility than using keyword arguements. Also I've heard that keywords have a non-trivial amount of overhead in Julia currently, which could be bad for functions called within an inner loop? Does anyone know more about this?

> by moving the gradient call up, we have to have an if iteration > 1 control flow statement around the hessian update, correct?

Yeah I see what you're saying - I suppose you could set a flag in the BFGS type to do this. I think my suggestion of breaking the functions into `update_state!`, etc. would fix my problem if you preferred to not do this.

johnmyleswhite:
> keywords have a non-trivial amount of overhead in Julia currently

My understanding is that keyword arguments are totally unspecialized, so there's no code-generation that exploits knowledge about their types.

pkofod:
> My understanding is that keyword arguments are totally unspecialized, so there's no code-generation that exploits knowledge about their types.

I don't have a reference, but I don't think that is correct. I think a specialized method *is* made when some combination of normal inputs and keywords are called, you just can't dispatch on the keywords. I will have to think about where I think I heard that :)

johnmyleswhite:
I'm not sure if that's right either. For example, this LLVM output looks pretty dire for keyword arguments in extremely simple functions: https://gist.github.com/johnmyleswhite/7484870a0f3656c735ae38bd1d5a0268

pkofod:
> I'm not sure if that's right either. For example, this LLVM output looks pretty dire for keyword arguments in extremely simple functions: https://gist.github.com/johnmyleswhite/7484870a0f3656c735ae38bd1d5a0268

I think this is what I was remembering: http://docs.julialang.org/en/release-0.4/manual/performance-tips/#declare-types-of-keyword-arguments

johnmyleswhite:
Hmm. I think I don't understand that section either or how it relates to the performance of the variant `foo(x; y::Int = 0)`, which still generates insane LLVM.

pkofod:
If you want a more strictly enforced time limit:
```julia
using Optim
problem = Optim.UnconstrainedProblems.examples["Rosenbrock"]

f = problem.f
initial_x = problem.initial_x

function very_slow(x)
    sleep(.5)
    f(x)
end

start_time = time()
time_to_setup = zeros(1)
function advanced_time_control(x)
    println(" * Iteration:       ", x.iteration)
    so_far =  time()-start_time
    println(" * Time so far:     ", so_far)
    if x.iteration == 0
        time_to_setup[:] = time()-start_time
    else
        expected_next_time = so_far + (time()-start_time-time_to_setup[1])/(x.iteration)
        println(" * Next iteration ≈ ", expected_next_time)
        println()
        return expected_next_time < 13 ? false : true
    end
    println()
    false
end
```
with output
```jlcon
julia> optimize(very_slow, zeros(2), NelderMead(), OptimizationOptions(callback = advanced_time_control))
 * Iteration:       0
 * Time so far:     2.219298839569092

 * Iteration:       1
 * Time so far:     3.4006409645080566
 * Next iteration ≈ 4.5429909229278564

 * Iteration:       2
 * Time so far:     4.403923988342285
 * Next iteration ≈ 5.476739525794983

 * Iteration:       3
 * Time so far:     5.407265901565552
 * Next iteration ≈ 6.4569235642751055

 * Iteration:       4
 * Time so far:     5.909044027328491
 * Next iteration ≈ 6.821732044219971

 * Iteration:       5
 * Time so far:     6.912338972091675
 * Next iteration ≈ 7.843148183822632

 * Iteration:       6
 * Time so far:     7.9156060218811035
 * Next iteration ≈ 8.85849153995514

 * Iteration:       7
 * Time so far:     8.918903827667236
 * Next iteration ≈ 9.870419979095459

 * Iteration:       8
 * Time so far:     9.922197818756104
 * Next iteration ≈ 10.880185931921005

 * Iteration:       9
 * Time so far:     10.925468921661377
 * Next iteration ≈ 11.888488478130764

 * Iteration:       10
 * Time so far:     11.92870283126831
 * Next iteration ≈ 12.895747828483582

 * Iteration:       11
 * Time so far:     12.932114839553833
 * Next iteration ≈ 13.902462200684981

 * Algorithm: Nelder-Mead
 * Starting Point: [0.0,0.0]
 * Minimizer: [0.23359374999999996,0.042187499999999996, ...]
 * Minimum: 6.291677e-01
 * Iterations: 11
 * Convergence: false
   *  √(Σ(yᵢ-ȳ)²)/n < 1.0e-08: false
 * Objective Function Calls: 24
```

pkofod:
> Still really excited about this! One thing I'd like to try is to build on top of this to optimize problems with constraints.
> 
> I'm not sure if this will end up being a dead end, but I decided to start experiment by doing projected gradient descent to solve a non-negative least squares problem (a terrible way to solve it, but a useful toy problem nonetheless).
> 
> using Optim
> 
> const A = rand(10,10)
> const b = rand(10)
> 
> f(x) = 0.5*sumabs2(A*x - b)
> d = DifferentiableFunction(f)
> x0 = randn(10)
> 
> # project onto non-negative orthant
> project!(x) = map!(z-> z<0 ? 0.0 : z, x)
> 
> gd = GradientDescent()
> o = OptimizationOptions()
> 
> s = Optim.initial_state(gd, o, d, x0)
> 
> for iter = 1:iterations
>     Optim.update!(d, s, gd)
>     project!(s.x)
> end
> 
> Right now I get this error:
> 
> ERROR: LoadError: AssertionError: c > 0
>  in hz_linesearch!(::Optim.DifferentiableFunction, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Optim.LineSearchResults{Float64}, ::Float64, ::Bool, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Int64, ::Float64, ::Int64, ::Int64) at /Users/alex/.julia/v0.5/Optim/src/linesearch/hz_linesearch.jl:186
>  in hz_linesearch!(::Optim.DifferentiableFunction, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Optim.LineSearchResults{Float64}, ::Float64, ::Bool, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Int64, ::Float64, ::Int64) at /Users/alex/.julia/v0.5/Optim/src/linesearch/hz_linesearch.jl:176 (repeats 4 times)
>  in update!(::Optim.DifferentiableFunction, ::Optim.GradientDescentState{Float64}, ::Optim.GradientDescent{Void}) at /Users/alex/.julia/v0.5/Optim/src/gradient_descent.jl:52
>  in macro expansion; at /Users/alex/.julia/v0.5/Optim/examples/projected_gd.jl:23 [inlined]
>  in macro expansion; at /Users/alex/.julia/v0.5/ProgressMeter/src/ProgressMeter.jl:473 [inlined]
>  in anonymous at ./<missing>:?
>  in include_from_node1(::String) at ./loading.jl:426
> while loading /Users/alex/.julia/v0.5/Optim/examples/projected_gd.jl, in expression starting on line 22
> 
> I'll keep playing around with this, but am curious to hear feedback on whether this will mesh well with the intended approach here.
> 
> (I also want to mention #254 so that earlier discussion is linked to this PR)

This now works (I think)
```julia
using Optim

const A = rand(10,10)
const b = rand(10)

f(x) = 0.5*sumabs2(A*x - b)
d = DifferentiableFunction(f)
x0 = randn(10)

# project onto non-negative orthant
project!(x) = map!(z-> z<0 ? 0.0 : z, x)

gd = GradientDescent()
o = OptimizationOptions()

s = Optim.initial_state(gd, o, d, x0)
iterations = 100
for iter = 1:iterations
    Optim.update_state!(d, s, gd)
    project!(s.x)
    Optim.update_g!(d, s, gd)
end
```

pkofod:
cc: @ahwillia 

(btw, if I edit a post and add a mention, are people then pinged?)

src/accelerated_gradient_descent.jl
pkofod: Any comments for my comment @johnmyleswhite ? At least I think we should do this with the line search part, so
```julia
method.linesearch!(d, state.x, state.s, state.x_ls, state.g_ls, state.lsr,
                     state.alpha, state.mayterminate)
```
becomes
```julia
method.linesearch!(d, state.x, state.s, state.linesearch_buffers)
```
or whatever name the field might get.

pkofod:
I believe most of the work in the PR is done by now. The only problem right now is the appearance of `state.` everywhere. We may want to unpack them ala Parameters.jl, but that is more of a prettifying thing. If it blocks work such as the linesearch PR, then I think it can wait. Comments are still very welcome.

ahwillia:
👏 nice. I will play around with this later this week. Let me know if there is anything specific you'd like feedback on.

pkofod:
Alright, I'll go ahead and merge this, but this doesn't have to be the final say. We can test it out, and tweak it before the next tag. Thanks for all the comments!

pkofod:
> I'm not sure if that's right either. For example, this LLVM output looks pretty dire for keyword arguments in extremely simple functions: https://gist.github.com/johnmyleswhite/7484870a0f3656c735ae38bd1d5a0268

I know this is closed now, but @johnmyleswhite we were not the first to be surprised/confused by the wording in the docs https://github.com/JuliaLang/julia/issues/9551#issuecomment-73364527

PR #267, opened 2016-08-16
fixed levenberg_marquardt/show_trace crash

bjarthur:
this fixes a bug which crept in recently when using `show_trace=true` in levenberg_marquardt.


pkofod:
Thanks for adding the test as well!

bjarthur:
selfish reasons really.  otherwise, someone would likely break it again and then I would have to fix it again!

pkofod:
I'll let the CI run, and then I'll merge.

PR #269, opened 2016-08-17
Prepare for v0.6.1.

### LINKED ISSUE #261: Linesearch fails when derivative is zero

pkofod:
I thought we might as well ship some of the post v0.6.0 commits already as a patch. I don't intend to ANN patches, but these will be included in the ANN for v0.7. 

I added a note about the convergence check added in #262 which should close #261 for now.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+# Optim v0.6.1 release notes
+* Assess convergence in *g* before iterating to avoid line search errors if `initial_x` is a stationary point
+* Fix trace bug in LevenbergMarquardt.
+* Add ForwardDiff AD functionality to NewtonTrustRegion
+* Make documentation even more noticable in README.md
+
 # Optim v0.6.0 release notes
 
 * Various bug fixes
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,3 +1,9 @@
+# Optim v0.6.1 release notes
+* Assess convergence in *g* before iterating to avoid line search errors if `initial_x` is a minimizer
```
johnmyleswhite: "is a minimizer" -> "is a stationary point"?

pkofod: Ah, of course, that's the whole point :) Ill fix it, thanks

PR #273, opened 2016-09-05
Small cleanup of fminbox code

anriseth:
Currently, Fminbox calls df.fg! on line 158, but never uses the resulting function evaluation. I've just changed the code to only store the gradient information.


pkofod:
Great, also great with the more verbose error

PR #274, opened 2016-09-07
Correct bound constraints

matthieugomez:
In previous version, estimate could get out of the box constraints @bjarthur


bjarthur:
good catch.  can you please add a test for which the old code would fail?  thanks.

pkofod:
Nice to see a bug being fixed, but as mentioned above, it would be very nice if you could add a test so the bug doesn't reappear.

matthieugomez:
See #274

PR #276, opened 2016-09-07
Remove obsolete line search test file

anriseth:
This test file is not used, and is testing a non-existent function `Optim.backtracking_line_search!`


pkofod:
Thanks

PR #277, opened 2016-09-07
RFC: Move linesearches to LineSearches.jl

anriseth:
After the discussion in #263 , I made an attempt to pull the line search code out of Optim and set up a package (https://github.com/anriseth/LineSearches.jl).  If people are happy to use this as a starting point, would probably be better to transfer the repo to JuliaOpt.

I don't have time to implement any new functionality, but can finalize the move if this is of interest.

TODO:
- [x] Add NEWS.md
- [x] Update documentation
- [x] Ensure we don't break user code (Reexport.jl, or export + deprecate from LineSearches.jl)
- [x] Move line search tests to LineSearches.jl
- [x] Publish LineSearches
- [x] ~~Add LineSearches.jl to JuliaOpt?~~ (Not a requirement for the PR)
- [x] Change deprecation names (see anriseth/LineSearches.jl#8 )


pkofod:
Thanks for this! Love the quick response.

The only problem with the code change, as it stands right now, is that it breaks a few packages. I can't remember the packages, but some packages use Optim only for the line search functionality. Either we have to deprecate it properly, or make sure that LineSearches.jl is publishes and that packages switch from Optim, before a version of Optim with this change is tagged. The unfortunate thing about the last option is that it potentially breaks user code.

I am 100% for this separation of Optim and the line search functionality, but it is always nice to think about the implications for the package ecosystem as a whole. 

Thoughts?

pkofod:
Also, it would be nice if you could add the change to `NEWS.md`, and also update the (incomplete) line search documentation to reflect that line searches are implemented in a separate, but "related", package.

ahwillia:
Maybe you could use Reexport.jl so that packages using Optim don't even
notice the difference?

anriseth:
I've set up a TODO list in the first post (anything missing?)

I'm in favour of deprecating the linesearch functions in Optim, and set a date/milestone for when we remove the functionality from Optim.

pkofod:
> I'm in favour of deprecating the linesearch functions in Optim, and set a date/milestone for when we remove the functionality from Optim.

That would simply be two major tags from "now" :) One round with deprecations, and then complete removal in the next tag. Reexport seems like the right thing to do tbh.

LineSearches.jl definitely belongs to JuliaOpt imo, but we can always move it at a later point. Luckily, Github is smart enough to figure that out :)

anriseth:
Optim does not currently export any of the linesearch functions, so Reexport does not do exactly what we're aiming for here I'd say. I've just added the functions with `@deprecate` instead.

pkofod:
Any reason why you removed the indentation in the latest commit? (that's what it looks like at least)

anriseth:
Julia's code style is to not indent code within `module` blocks, but I can
redo it if you want to keep it the old way.

anriseth:
LineSearches v0.0.1 is in METADATA. Just closing and re-opening to rerun Travis

pkofod:
@johnmyleswhite do you have any comments or objections here?

src/types.jl
```diff
@@ -83,13 +83,15 @@ type UnivariateOptimizationResults{T,M} <: OptimizationResults
     f_calls::Int
 end
 
-immutable DifferentiableFunction
+immutable DifferentiableFunction <: LineSearches.AbstractDifferentiableFunction
```
pkofod: Why is this needed ?

src/types.jl
```diff
@@ -83,13 +83,15 @@ type UnivariateOptimizationResults{T,M} <: OptimizationResults
...
     f::Function
     g!::Function
     fg!::Function
 end
 
-immutable TwiceDifferentiableFunction
+abstract AbstractTwiceDifferentiableFunction <: LineSearches.AbstractDifferentiableFunction
```
pkofod: ... and this?

src/types.jl
```diff
@@ -83,13 +83,15 @@ type UnivariateOptimizationResults{T,M} <: OptimizationResults
...
+
+immutable TwiceDifferentiableFunction <: AbstractTwiceDifferentiableFunction
```
pkofod: ... and this? :)

src/types.jl
anriseth: Because the signatures of all the linesearch functions previously required the first argument to be either a DifferentiableFunction or a TwiceDifferentiableFunction. But these are an integral part of Optim, so I decided to make abstract types in LineSearches instead of moving DifferentiableFunction and TwiceDifferentiableFunction to LineSearches.jl.

src/types.jl
anriseth: This one was just an abstract that says that C^2 functions are subsets of C^1 functions. It doesn't do much here, so we could just subtype TwiceDifferentiableFunction to LineSearches.AbstractDifferentiableFunction. (LineSearches don't use second-order information, so I didn't implement it there)

src/types.jl
anriseth: Alternatively, we can create functions to convert from the Optim types to a similar type in LineSearches, but I think that would slow things down?

pkofod: Okay, thank you.

I guess the easier thing would just be to not put a signature there? I mean. That function name is not used for anything else, so do we really need to decorate `df`? I'm just trying to separate the two packages as much as possible.  It would of course require a short explanation that line searches accept some input `df` which is supposed to have fields `f`, `g!` and `fg!` (depending on the specific line search, the last one is not necessary). Then the users *can* import the `*Function` family from Optim if they want, but it is not strictly necessary.

anriseth: Oh yeah, that does make more sense. 
In the unlikely event that there are ambiguity clashes in the future, we might have to rethink - but it seems unlikely.

anriseth:
New version of LineSearches without the abstract types.

pkofod:
Just pinging again, @johnmyleswhite are you :+1:  or :-1:  on this one?

cortner:
Is there a symbol for thumb-twiddling? I'd love to see this merged soon, since I am holding back some small (but important for me!) changes to Optim.

pkofod:
A test was failing, and I just ran it again to be sure. I'm going to take https://github.com/JuliaOpt/Optim.jl/issues/263#issuecomment-245292789 as a :+1: . Given https://github.com/anriseth/LineSearches.jl/blob/e37e4f1cdd24424562712c0926ab6b0fe9e63085/src/deprecate.jl , I feel safe to merge this afterwards.

anriseth:
There's a conflict in NEWS.md now, due to yesterdays PR merge #288. Shall I rebase, or does github deal with things like that in its merging toolbox?

pkofod:
Yes, I know, that was my fault. No, unfortunately, I can't fix it from github. It should be a simple merge.

pkofod:
Bye bye LineSearches, I hope you enjoy your new home.

PR #284, opened 2016-09-22
Fixes solve_tr_subproblem!

jeff-regier:
This PR adds unit tests for `solve_tr_subproblem!` based on randomly generated Hessians and gradients. For some indefinite Hessians, the tests fail.

This PR also includes a fix: `lambda_lb` does not need to be so large. With large values of `lambda_lb`, the trust region method finds interior solutions to the quadratic minimization problem when it should find a boundary solution. All tests pass with a smaller value for `lambda_lb`.

I've also verified that the fix doesn't break anything in the Celeste.jl package, which depends on the newton trust region code in Optim.jl, and may call `solve_tr_subproblem!` with Hessians that are more poorly conditioned than the Hessians these new unit tests generate at random.

### DIFF
--- a/src/newton_trust_region.jl
+++ b/src/newton_trust_region.jl
@@ -211,7 +211,7 @@ function solve_tr_subproblem!{T}(gr::Vector{T},
         end
     end
 
-    m = vecdot(gr, s) + 0.5 * vecdot(s, H_ridged * s)
+    m = vecdot(gr, s) + 0.5 * vecdot(s, H * s)
 
     return m, interior, lambda, hard_case, reached_solution
 end
--- a/src/newton_trust_region.jl
+++ b/src/newton_trust_region.jl
@@ -139,7 +139,7 @@ function solve_tr_subproblem!{T}(gr::Vector{T},
 
         # Solutions smaller than this lower bound on lambda are not allowed:
         # they don't ridge H enough to make H_ridge PSD.
-        lambda_lb = -min_H_ev + max(1e-4, abs(min_H_ev) * 1e-4)
+        lambda_lb = -min_H_ev + max(1e-8, 1e-8 * (max_H_ev - min_H_ev))
         lambda = lambda_lb
 
         hard_case = false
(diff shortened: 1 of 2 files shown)


pkofod:
Thanks for the "fix"! I can't help but ask: is this value your own heuristic, or is there some sort of suggestion you're following?

jeff-regier:
It's just a heuristic. I'd prefer a more principled approach, though this
seems like a clear improvement over what it replaces. (I made up that
number too.) This number, `1e-8`, becomes the minimum eigenvalue of a
matrix that we do a Cholesky decomposition on. Much smaller and the matrix
isn't PSD. I'm not sure how much smaller we could go without causing the
decomposition to fail. (Does it depend on the maximum eigenvalue too?)

pkofod:
Great, I just wanted to know if this was based on something that would warrant a reference. It would probably be impossible to find a "one size fits all" solution here anyway.

pkofod:
@jeff-regier should I consider this PR a WIP? I just noticed you've been pushing.

jeff-regier:
Hi Patrick, I didn't mean for this to be a WIP but I came across a really poorly conditioned Hessian today while using my fork of Optim.jl, so I made the change. I hope that's OK. I've squashed 2 of my 3 commits now so the modification I made today isn't in it's own commit. I think this PR should be fine to merge once the travis tests pass.

jeff-regier:
Looks like an issue with travis:
```
The command "curl -s -L --retry 7 'https://s3.amazonaws.com/julialang/bin/linux/x64/0.5/julia-0.5-latest-linux-x86_64.tar.gz' | tar -C ~/julia -x -z --strip-components=1 -f -" failed and exited with 2 during .
```
Would you try restarting the travis job?

anriseth:
You can restart Travis yourself by closing and reopening the PR

jeff-regier:
It's passing now

pkofod:
Wonderful, thanks!

PR #288, opened 2016-09-26
minimum -> minimizer and f_minimum -> minimum.

### LINKED ISSUE #286: Minimum vs Minimiser
This is in some ways a trivial issue, but I stumbled over it several times already and then have to go look which fields of `result` have which meaning: the *minimum* is the value of the energy at the *minimiser* (or, minimizer if you must) - see e.g. Nocedal/Wright, Sec. 2.1. 

I suggest to rename `result.minimum` to become `result.minimizer` or `result.solution` or `result.final_x`. Anything will do.

pkofod:
fix #286 , ref #76 #165

- [x] mention in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+# Optim (unversioned) release notes
+* Make move to minimizer (the argmin) and minimum (objective at argmin) final: field names are now in sync with function based API.
+
 # Optim v0.6.1 release notes
 * Assess convergence in *g* before iterating to avoid line search errors if `initial_x` is a stationary point
 * Fix trace bug in LevenbergMarquardt.
--- a/benchmarks/adfd.jl
+++ b/benchmarks/adfd.jl
@@ -63,7 +63,7 @@ for (name, problem) in Optim.UnconstrainedProblems.examples
                            method = algorithm,
                            g_tol = 1e-8,
                            autodiff = ad)
-        errors = minimum(map(sol -> norm(results.minimum - sol), problem.solutions))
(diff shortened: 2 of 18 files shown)


pkofod:
@rgiordan @jeff-regier do you have any idea why it fails with `Float32`? Is this a numerical issue, or ?

jeff-regier:
@pkofod Perhaps the Hessian is too poorly conditioned for `Float32`? Do you know the condition number of the problematic test function?

If poor conditioning is the problem, we may want to change the magic number (`1e-8`) you and I discussed previously, in my last PR. We may want that magic number to be data type specific: larger for `Float32`, as it is now for `Float64`. I'm not really sure how to pick that number though for `Float32`. Any idea how poorly a `Float32` matrix be conditioned without causing `chol` to fail?

jeff-regier:
Also, what changed in this PR the revealed the problem? New unit tests?

anriseth:
I think this happened because of the #266 PR, not this PR:
https://travis-ci.org/JuliaOpt/Optim.jl/builds/162901749

pkofod:
Exactly, it is because in the refactoring I added `NewtonTrustRegion` to the type stability tests. Then I merged your fix, and I guess I should have re-fired my tests after that, see https://github.com/JuliaOpt/Optim.jl/blob/11c935e3689ce05e8ef75ea30fff7f592a26f56d/test/type_stability.jl#L46

I do think we should maybe make it `T`-type specific. I think I will have to uncomment the tests again, and then we can work on it in another PR.

Issue #294, opened 2016-10-09
StepLength Extrapolation


cortner:
Aside from 
- search direction 
- line search
  the step-length extrapolation (from step-length, gradient and function history) is maybe the most crucial aspect of the optimisation algorithms implemented in Optim. 

I suggest to (minimally) refactor `Optim` to allow the user to choose how to perform this? E.g. `cg.jl` uses `alphatry` which is HZ's method, `LBFGS` uses just the previous step. But I'd really like to play around or even try my own variants.

CC: @pkofod @anriseth 
   (this should probably be done in parallel in `Optim` and `LineSearches`)

cortner:
Maybe to simplify this, it might be better to make step length extrapolation part of the line-search type.

anriseth:
I'll try to deal with this tomorrow or Friday

pkofod:
Awesome!

Issue #282, opened 2016-09-16
Multiple f evaluations for same state


cortner:
So far I only checked this for LBFGS, but I suspect this occurs throughout Optim: 

The call to line search is this:

``` julia
 alpha, f_update, g_update = mo.linesearch!(d, x, s, x_ls, g_ls, lsr, alphaguess, mayterminate)
f_calls, g_calls = f_calls + f_update, g_calls + g_update
```

Because the current gradient and objective are not passed in, this means that there is one extra evaluation in `linesearch!`, since in the main loop there is also a call 

``` juli
f_x = d.fg!(x, gr)
```

So I propose to change the `linesearch!` call  to something like 

``` juli
alpha, f_update, g_update = mo.linesearch!(d, x, s, f_x, gr, x_ls, g_ls, lsr, alphaguess, mayterminate)
```

In fact, one could also pass in `dphi0` (or pass this instead of `gr`) which is also recomputed in the linesearch.

This won't affect problems where a lot of work is spent in the line-search, but it will very much affect well-scaled (e.g. preconditioned) problems when solved with LBFGS where ideally the Wolfe conditions are checked and then returned without any evaluation.

I'll test this and post an update here.

cortner:
@anriseth since you are maintaining `LineSearches` (??) it would be good to have your thoughts.

KristofferC:
I have worked around this in NLsolve in a slightly ugly way: https://github.com/EconForge/NLsolve.jl/blob/e41f1701ffe7f2e7060b686f922a7ea15b4a3708/src/newton.jl#L82

Note that in:
https://github.com/EconForge/NLsolve.jl/commit/5fb93cf1aa767f471d672fc56ff9d6e260374ffd
and
https://github.com/EconForge/NLsolve.jl/commit/04ec307c3c6316c76225308f733f2968679eb2dd

this reduced the number of f and g calls significantly.

cortner:
I've been wondering whether the line search interface is not very flexible right now and one should just pass the entire Optimisation State to the line search?

KristofferC:
Could have something like

```
f_updated::Bool
g_updated::Bool
```

and then have each call to `f` or `g` set `f_updated`, `g_updated` to `true` respectively. Any move of the current point sets both these to `false`. If `f_updated` is true, a call to `f` will just return the last value.

cortner:
[...]

PR #295, opened 2016-10-22
Extraplbfgs2

cortner:
This is small modification to LBFGS which should leave the default behaviour completely unchanged. But for some problems I get significant speed-up even over the HZ-CG.

```
--------------------
Rosenbrock Example:
--------------------
LBFGS Default Options: g_calls = 157, f_calls = 157
CG Default Options: g_calls = 65, f_calls = 86
LBFGS + Backtracking + Extrapolation: g_calls = 23, f_calls = 52
--------------------------------------
p-Laplacian Example (preconditioned):
--------------------------------------
LBFGS Default Options: g_calls = 15, f_calls = 15
CG Default Options: g_calls = 8, f_calls = 11
LBFGS + Backtracking + Extrapolation: g_calls = 9, f_calls = 18
--------------------------------------
JuLIP Example (preconditioned):
--------------------------------------
LBFGS Default Options: g_calls = 168, f_calls = 168
CG Default Options: g_calls = 93, f_calls = 135
LBFGS + Backtracking + Extrapolation: g_calls = 47, f_calls = 124
```

(only the first two of these tests are contained in this PR)

The modification adds a simple step-size extrapolation step to LBFGS, which reverts to 1.0 when safe to do so.


test/runtests.jl
```diff
@@ -7,30 +7,31 @@ using Base.Test
...
-    "api.jl",
-    "golden_section.jl",
-    "brent.jl",
-    "type_stability.jl",
-    "array.jl",
-    "constrained.jl",
-    "callbacks.jl",
-    "precon.jl",
-    "initial_convergence.jl"
+   #  "types.jl",
```
anriseth: You've commented out all the other tests.

pkofod: Can't seem to reply to review comments but see #259 for benchmark related discussions. I intend to move it to a package and maybe leverage pkgbenchmarks.jl

cortner:
> Improvements are great. Are there any references in the literature on this approach?
Yes, as in this is one of the standard extrapolation techniques from N/W 
No, as in I don't think anybody in optimisation recommends this for BFGS (but I could be wrong) 

> I think it could be good for future performance improvements, if we create a larger set of benchmark problems (test them in Cutest?) so that can make decisions on default behavior the algorithms.

That would be great.

cortner:
I should also say: this is a stop-gap solution; in the future this should be re-organised; see #294 .

pkofod:
I see my comment didn't make it. But I hope to implement #259 in a separate package soon. Then it shouldn't be more than running the benchmarks there.

cortner:
any idea why the linux test doesn't pass?

cortner:
I just noticed that each iteration there is still one additional function call. This means that my benchmarks in the beginning should actually read 
```
--------------------
Rosenbrock Example:
--------------------
LBFGS Default Options: g_calls = 157, f_calls = 157
CG Default Options: g_calls = 65, f_calls = 86
LBFGS + Backtracking + Extrapolation: g_calls = 23, f_calls = 29
--------------------------------------
p-Laplacian Example (preconditioned):
--------------------------------------
LBFGS Default Options: g_calls = 15, f_calls = 15
CG Default Options: g_calls = 8, f_calls = 11
LBFGS + Backtracking + Extrapolation: g_calls = 9, f_calls = 9
--------------------------------------
JuLIP Example (preconditioned):
--------------------------------------
LBFGS Default Options: g_calls = 168, f_calls = 168
CG Default Options: g_calls = 93, f_calls = 135
LBFGS + Backtracking + Extrapolation: g_calls = 47, f_calls = 77
```

This is because at the function value from the final line-search iterate is not passed back to the main loop (in `update_g!` I think). #282 discusses this, but I wonder what a stop-gap solution would be?

anriseth:
I just had a look at the "Initial step length" section in Nocedal and Wright.
They say that "For Newton and quasi-Newton methods, the step $\alpha_0=1$ should always be used as the initial trial step length". Have you tried to compare the performance you see here with that method? I would guess that when the quasi-Newton has a good approximation of the Hessian, this would perform better. 

The other two step length guesses that they mention (including the one implemented here), are said to be important for conjugate gradient and gradient descent. 

More flexibility in choice is of course best here - so as you say, separating this out from the specific algorithm will help us experiment with the best default options (#294).

cortner:
> I just had a look at the "Initial step length" section in Nocedal and Wright.
They say that "For Newton and quasi-Newton methods, the step $\alpha_0=1$ should always be used as the initial trial step length".

The current LBFGS doesn't do that at all. Mine actually does: when the step choice is close to 1, then it "snaps" onto 1. The HZ line-search gets away with it because (a) it uses Wolfe and (b) it becomes more and more accurate with each iteration hence ought to approach 1 in the limit.

>  Have you tried to compare the performance you see here with that method? I would guess that when the quasi-Newton has a good approximation of the Hessian, this would perform better.

This is almost never the case in the pre-asymptotic regime. Anyhow, here are some tests in which I now always set `alphaguess = 1.0` unless the extrapolation is turned on. In the first two examples the performance is comparable (these are very easy problems) but in the last one it gets slightly worse.
```
INFO: Testing Optim
Running tests:
 * extrapolate.jl
--------------------
Rosenbrock Example:
--------------------
LBFGS Default Options: g_calls = 69, f_calls = 69
CG Default Options: g_calls = 65, f_calls = 86
LBFGS + Backtracking + Extrapolation: g_calls = 23, f_calls = 51
LBFGS + Backtracking: g_calls = 22, f_calls = 53
--------------------------------------
p-Laplacian Example (preconditioned):
--------------------------------------
LBFGS Default Options: g_calls = 17, f_calls = 17
CG Default Options: g_calls = 8, f_calls = 11
LBFGS + Backtracking + Extrapolation: g_calls = 11, f_calls = 22
LBFGS + Backtracking: g_calls = 9, f_calls = 18
--------------------------------------
JuLIP Example (preconditioned):
--------------------------------------
LBFGS Default Options: g_calls = 200, f_calls = 200
CG Default Options: g_calls = 99, f_calls = 143
LBFGS + Backtracking + Extrapolation: g_calls = 50, f_calls = 131
LBFGS + Backtracking: g_calls = 55, f_calls = 190
CG + Backtracking:  g_calls = 47, f_calls = 142
INFO: Optim tests passed
```

cortner:
For what it's worth, if we can redesign `LineSearches` according to [LineSearches.jl #9](https://github.com/anriseth/LineSearches.jl/issues/9) then it will be easy to set various options. 

As I said above: this is a stop-gap solution, and it actually *doesn't change default behaviour*

cortner:
any reason why this can't be merged?

pkofod:
> any reason why this can't be merged?

Yes, I haven't had time. Not a good reason, but one nonetheless :)

There's just one thing I'm not to sure about, but I'll mention it at the line.

src/l_bfgs.jl
```diff
@@ -146,10 +149,26 @@ function update_state!{T}(d, state::LBFGSState{T}, method::LBFGS)
     LineSearches.clear!(state.lsr)
     push!(state.lsr, zero(T), state.f_x, dphi0)
 
+    # compute an initial guess for the linesearch based on
+    # Nocedal/Wright, 2nd ed, (3.60)
+    # TODO: this is a temporary fix, but should eventually be split off into
+    #       a separate type and possibly live in LineSearches; see #294
+    if method.extrapolate && state.pseudo_iteration > 1
+        alphaguess = 2.0 * (state.f_x - state.f_x_previous) / dphi0
+        alphaguess = max(alphaguess, state.alpha/4.0)  # not too much reduction
```
pkofod: What does this line do exactly? Where does the `4.0` come from?

cortner:
thank you - much appreciated.

PR #296, opened 2016-10-27
Concerning optimisation of univariate functions using bracketing methods - Documentation

ngiann:
I recently run into a little bug (in my own code) when optimising a univariate function using the optim package.

In particular I was optimising an upper bound which was expressed as the sum of univariate functions f_i.
These univariate functions were bounded but not necessarily unimodal.
Calling "optimize(f_i, L, U)" would hence sometimes lead to the upper bound becoming lower than higher.
I finally realised that the algorithm relies of course on being able to bracket the functions f_i; if this is not possible then just a local minimum will be found as opposed to a global minimum.

I think that this should be pointed out to users that do not realise that optimising a univariate function relies on being able to bracket it in a given interval.

### DIFF
--- a/docs/src/user/minimization.md
+++ b/docs/src/user/minimization.md
@@ -117,6 +117,8 @@ In addition to the `iterations`, `store_trace`, `show_trace` and
 * `rel_tol`: The relative tolerance used for determining convergence. Defaults to `sqrt(eps(T))`.
 * `abs_tol`: The absolute tolerance used for determining convergence. Defaults to `eps(T)`.
 
+Important notice: the minimization of a univariate function in a bounded interval assumes that your function can be "bracketet" in the given interval. If this is not the case, optimisation of the univariate function may not necessarily yield a global minimum.
+
 ## Obtaining results
 After we have our results in `res`, we can use the API for getting optimization results.
 This consists of a collection of functions. They are not exported, so they have to be prefixed by `Optim.`.


docs/src/user/minimization.md
```diff
@@ -117,6 +117,8 @@ In addition to the `iterations`, `store_trace`, `show_trace` and
 * `rel_tol`: The relative tolerance used for determining convergence. Defaults to `sqrt(eps(T))`.
 * `abs_tol`: The absolute tolerance used for determining convergence. Defaults to `eps(T)`.
 
+Important notice: the minimization of a univariate function in a bounded interval assumes that your function can be "bracketet" in the given interval. If this is not the case, optimisation of the univariate function may not necessarily yield a global minimum.
```
anriseth: "bracketed"

pkofod:
I'm not sure I understand what you mean here? That for a function with more than one critical point you can end up with a local minimum? That should hopefully not surprise anyone. I'm not sure I understand your use of "bracket" fully, care to elaborate?

ngiann:
Hello Patrick,

indeed this is exactly what I mean.
Of course this is not a surprise to people familiar with optimisation.
People that are not that familiar, may be surprised.

By "bracketed" I mean finding an interval (a bracket) in which it is guaranteed that the function has a single optimum by the intermediate value theorem.
I think the term "bracketing" is fairly common, but perhaps it should be avoided.

Maybe my proposed sentence should be rephrased so that it does not sound as much as a warning but as a reminder?
Or if people think it is superfluous, we may as well discard it.
Thanks for your comment.

ngiann:
Hello Patrick,

indeed this is exactly what I mean.
Of course this is not surprise people familiar with optimisation.
People that are not that familiar perhaps may be surprised.

By "bracketed" I mean finding an interval (a bracket) in which it is
guaranteed that the function has a single optimum by the intermediate value
theorem.
I think the term "bracketing" is fairly common, but perhaps it should be
avoided,

Maybe my proposed sentence should be rephrased so that it does not sound as
much as a warning but as a reminder?
Or if people think it is superfluous, we may as well discard it.
Thanks for your comment.

2016-10-27 13:06 GMT+02:00 Patrick Kofod Mogensen <notifications@github.com>
:

pkofod:
Okay, then I understand what you mean. However, you can also do 1D optimization using other methods (not all though), so it might be a bit misleading. Maybe you can instead add a small section somewhere that says: this package generally does not provide any guarantees for finding global optima. Personally, I think it is superfluous, but my view might not coincide with that of our users, so...

johnmyleswhite:
I think it can't hurt to document these kinds of general optimization concepts as this package isn't meant to be used only by experts in optimization, even if correct usage will always require the user to understand what they're doing at some level.

I think the proposed text is a little confusing since it makes use of jargon about bracketing that people who aren't familiar with optimization are unlikely to understand. Can we iterate on it?

ngiann:
Hello everyone,

I have come around to Patrick's opinion that we do not need to mention explicitly this issue. After all, no  optimisation routine guarantees that the global minimum will be found. Moreover, it may sound a bit weird to mention this fact just for Brent's method and the Golden Section and not for any of the other algorithms. I will remove this issue in a few days if that's ok. Thanks for all the comments, cheers.

pkofod:
> Hello everyone,
> 
> I have come around to Patrick's opinion that we do not need to mention explicitly this issue. After all, no optimisation routine guarantees that the global minimum will be found. Moreover, it may sound a bit weird to mention this fact just for Brent's method and the Golden Section and not for any of the other algorithms. I will remove this issue in a few days if that's ok. Thanks for all the comments, cheers.

Maybe just mention it very generally on the introduction page, and maybe refer to some global optimization packages (BlackBoxOptim?)

ngiann:
> Maybe just mention it very generally on the introduction page, and maybe refer to some global optimization packages (BlackBoxOptim?)

I second that.

PR #301, opened 2016-11-04
Update index.md - Following discussion on "Concerning optimisation of univariate functions using bracketing methods"

ngiann:
After some previous discussion (see closed pull request "Concerning optimisation of univariate functions using bracketing methods - Documentation #296"), some people suggested that it may be useful to those not familiar with optimisation in general, to add a brief comment saying that the Optim package provides solver that will find local optima and cannot be used to locate a global optimum.

I have added now inserted such a brief comment. Let me know what you think, cheers.


pkofod:
Thanks.

PR #302, opened 2016-11-06
Add `optimize` call to catch all the cases

### LINKED ISSUE #300: Dispatch error on univariate optimization
* `optimize(cos, 0, 2pi)` works;
* `optimize(cos, 0, 2pi, Brent())` does not work (although `optimize(cos, 0.0, 2pi, Brent())` does);
* `optimize(cos, 0, 2pi, method = Brent())` works.

To my understanding, this happens because we have [`optimize{T <: AbstractFloat}(f::Function, lower::T, upper::T; kwargs...)`](https://github.com/JuliaOpt/Optim.jl/blob/master/src/optimize.jl#L286) and [`optimize(f::Function, lower::Real, upper::Real; kwargs...)`](https://github.com/JuliaOpt/Optim.jl/blob/master/src/optimize.jl#L316) in `optimize.jl` but the actual optimizer is defined in terms of [`optimize{T <: AbstractFloat}(f::Function, lower::T, upper::T, mo::Brent; kwargs...)`](https://github.com/JuliaOpt/Optim.jl/blob/master/src/brent.jl#L25) (and the equivalent for `GoldenSection`).

What is the best way to fix this problem?

Armavica:
Fixes #300.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,5 @@
 # Optim (unversioned) release notes
+* Fix dispatch bug for univariate optimization.
 * The line search functionality has been separated into a new package
   [LineSearches.jl](https://github.com/anriseth/LineSearches.jl), see
   [277](https://github.com/JuliaOpt/Optim.jl/pull/277).
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -322,3 +322,15 @@ function optimize(f::Function,
              Float64(upper);
              kwargs...)
 end
+
+function optimize(f::Function,
+                  lower::Real,
+                  upper::Real,
+                  mo::Union{Brent, GoldenSection};
+                  kwargs...)
+    optimize(f,
+             Float64(lower),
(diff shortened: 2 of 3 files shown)


pkofod:
Thank you for the PR! Tests seem to pass, but before merging this, I would have to ask you to include a test. Just to make sure that this doesn't regress!

It should be a simple test. Just show that you old "issue" no longer throws a `MethodError` using `@test_throws`.

pkofod:
@Armavica just a reminder that I'm ready to merge as soon as the test is added!

Armavica:
Sorry for the delay. I don't know how to use `@test_throws` to test for the absence of an error, but I hope that this will work as well.

pkofod:
I just came to say that what I said makes no sense. I'll look when I'm at a computer, and merge.

pkofod:
Thanks.

PR #303, opened 2016-11-08
Constrained optimization episode 2: revenge of the slack variables

timholy:
No matter how the American election turns out, today we have something to celebrate: the beginning of what I hope will be a merge-worthy native Julia constrained optimization framework.

Long ago said that I would not work on hard-core optimization without a debugger (errors triggered 1 hour into an optimization run are too painful to fix otherwise), but @Keno has delivered, so now it's time for me to pony up. Besides, I've grown tired of maintaining my own fork of Optim and want access to all the goodies and cleanups (*especially* the cleanups) that others have contributed.

Rather than implementing the whole framework at once, I decided to try to keep things "simple" (if a ~~1600~~2000 line diff can be called simple) by implementing just one method. I picked the interior-point Newton method, since that seems to be the mostly widely-used and/or described method, and for which the iteration is relatively simple. There are also many, many bells and whistles one could add. But I want to get a basic implementation of *something* merged before going on to other things.

Compared to #50, there are almost too many differences to mention, as this was a clean start. A few of the highlights:
- it's more in line with the MathProgBase interface (still not a perfect match, but pretty close), for example in [how the bounds are specified](https://github.com/JuliaOpt/Optim.jl/pull/50#issuecomment-37338779) and [how the constraint hessian is requested](https://github.com/JuliaOpt/Optim.jl/pull/50#issuecomment-37261452).
- it supports both equality and inequality constraints
- it should exhibit far greater numerical stability near the edges of the feasible region (this has been a big problem with the approach in #50)

It's also worth briefly mentioning that this might have some nice features, though it needs real-world testing for us to know whether these will materialize. For example, unlike Ipopt this is guaranteed to roughly preserve the descent direction of the user's objective function (it will not choose such a high barrier coefficient that it pushes the solution into a different basin).  

Testing are a mixed bag: there's quite a few tests for the low-level mathematics, but any overall tests (e.g., ones that involve a call to `optimize`) are a work in progress. I'm hoping to make use of CUTEst (https://github.com/JuliaSmoothOptimizers/CUTEst.jl/issues/43#issuecomment-256709346) and do a fairly thorough job, but that will take some time.

It's worth noting a few details:
- argument order may be somewhat confused; in some places I adopt mutated-first, in others (where I'm trying to match some preexisting API) it's last. It's my understanding that Optim will eventually migrate to mutated-first, but perhaps this is too inconsistent now
- I needed to customize the line search algorithm to account for constraints, so this goes against the migration to [LineSearches.jl](https://github.com/anriseth/LineSearches.jl) and reintroduces a linesearch method to Optim. Aside from simplicity in making updates (I don't want to be bothered with tagging another package just to make tests for a RFC pass), I kept this here because I was also trying to move towards a more sane linesearch API and so it doesn't fit with our classic functions. (It's based on the fact that anonymous functions are fast with julia 0.5 and higher---if people are unhappy with the overhead this causes on 0.4, tough. We have to write code for the future, not the past.)

CC @Cody-G, @dpo, @abelsiqueira.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -19,6 +19,7 @@ module Optim
     export optimize,
            isfeasible,
            isinterior,
+           nconstraints,
            DifferentiableFunction,
            TwiceDifferentiableFunction,
            DifferentiableConstraintsFunction,
@@ -110,8 +111,9 @@ module Optim
     include("utilities/trace.jl")
 
     # Examples for testing
-    include(joinpath("problems", "unconstrained.jl"))
+    include(joinpath("problems", "multivariate.jl"))
     include(joinpath("problems", "univariate.jl"))
+    using .MultivariateProblems
 
     cgdescent(args...) = error("API has changed. Please use cg.")
 end
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -17,6 +17,8 @@ module Optim
            Base.setindex!
 
     export optimize,
+           isfeasible,
+           isinterior,
            DifferentiableFunction,
            TwiceDifferentiableFunction,
            DifferentiableConstraintsFunction,
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -34,6 +34,7 @@ module Optim
            Fminbox,
            GoldenSection,
            GradientDescent,
+           IPNewton,
            LBFGS,
            MomentumGradientDescent,
            NelderMead,
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -19,6 +19,8 @@ module Optim
     export optimize,
            DifferentiableFunction,
            TwiceDifferentiableFunction,
+           DifferentiableConstraintsFunction,
+           TwiceDifferentiableConstraintsFunction,
            OptimizationOptions,
            OptimizationState,
            OptimizationTrace,
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -78,9 +78,9 @@ module Optim
 
     # Constrained optimization
     include("fminbox.jl")
+    include("iplinesearch.jl")
     include("interior.jl")
     include("ipnewton.jl")
-    include("iplinesearch.jl")
 
     # trust region methods
     include("levenberg_marquardt.jl")
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -78,6 +78,7 @@ module Optim
 
     # Constrained optimization
     include("fminbox.jl")
+    include("interior.jl")
 
     # trust region methods
     include("levenberg_marquardt.jl")
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -80,6 +80,7 @@ module Optim
     include("fminbox.jl")
     include("interior.jl")
     include("ipnewton.jl")
+    include("iplinesearch.jl")
 
     # trust region methods
     include("levenberg_marquardt.jl")
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -79,6 +79,7 @@ module Optim
     # Constrained optimization
     include("fminbox.jl")
     include("interior.jl")
+    include("ipnewton.jl")
 
     # trust region methods
     include("levenberg_marquardt.jl")
--- a/src/types.jl
+++ b/src/types.jl
@@ -228,3 +228,180 @@ function TwiceDifferentiableFunction(f::Function,
     end
     return TwiceDifferentiableFunction(f, g!, fg!, h!)
 end
+
+### Constraints
+#
+# Constraints are specified by the user as
+#    lx_i ≤   x[i]  ≤ ux_i  # variable (box) constraints
+#    lc_i ≤ c(x)[i] ≤ uc_i  # linear/nonlinear constraints
+# and become equality constraints with l_i = u_i. ±∞ are allowed for l
+# and u, in which case the relevant side(s) are unbounded.
+#
+# The user supplies functions to calculate c(x) and its derivatives.
+#
+# Of course we could unify the box-constraints into the
+# linear/nonlinear constraints, but that would force the user to
+# provide the variable-derivatives manually, which would be silly.
+#
+# This parametrization of the constraints gets "parsed" into a form
+# that speeds and simplifies the algorithm, at the cost of many
+# additional variables. See `parse_constraints` for details.
+
+immutable ConstraintBounds{T}
+    # Box-constraints on variables (i.e., directly on x)
+    eqx::Vector{Int} # index-vector of equality-constrained x (not actually variable...)
+    valx::Vector{T}  # value of equality-constrained x
+    ineqx::Vector{Int}  # index-vector of other inequality-constrained variables
+    σx::Vector{Int8}    # ±1, in constraints σ(v-b) ≥ 0 (sign depends on whether v>b or v<b)
+    bx::Vector{T}       # bound (upper or lower) on variable
+    iz::Vector{Int}     # index-vector of nonnegative or nonpositive variables
+    σz::Vector{Int8}    # ±1 depending on whether nonnegative or nonpositive
+    bz::Vector{T}       # all-zeros, convenience for evaluation of barrier penalty
+    # Linear/nonlinear constraint functions and bounds
+    eqc::Vector{Int}    # index-vector equality-constrained entries in c
+    valc::Vector{T}     # value of the equality-constraint
+    ineqc::Vector{Int}  # index-vector of inequality-constraints
+    σc::Vector{Int8}    # same as σx, bx except for the nonlinear constraints
+    bc::Vector{T}
+end
+function ConstraintBounds(lx, ux, lc, uc)
+    _cb(symmetrize(lx, ux)..., symmetrize(lc, uc)...)
+end
+function _cb{Tx,Tc}(lx::AbstractArray{Tx}, ux::AbstractArray{Tx}, lc::AbstractVector{Tc}, uc::AbstractVector{Tc})
+    T = promote_type(Tx,Tc)
+    ConstraintBounds{T}(parse_constraints(T, lx, ux, true)..., parse_constraints(T, lc, uc)...)
+end
+
+Base.eltype{T}(::Type{ConstraintBounds{T}}) = T
+Base.eltype(cb::ConstraintBounds) = eltype(typeof(cb))
+
+abstract AbstractConstraintsFunction
+
+immutable DifferentiableConstraintsFunction{F,J,T} <: AbstractConstraintsFunction
+    bounds::ConstraintBounds{T}
+    c!::F         # c!(x, storage) stores the value of the constraint-functions at x
+    jacobian!::J  # jacobian!(x, storage) stores the Jacobian of the constraint-functions
+end
+
+function DifferentiableConstraintsFunction(c!, jacobian!, lx, ux, lc, uc)
+    b = ConstraintBounds(lx, ux, lc, uc)
+    DifferentiableConstraintsFunction{typeof(c!), typeof(jacobian!), eltype(b)}(b, c!, jacobian!)
+end
+
+immutable TwiceDifferentiableConstraintsFunction{F,J,H,T,N} <: AbstractConstraintsFunction
+    bounds::ConstraintBounds{T}
+    c!::F
+    jacobian!::J
+    h!::H   # Hessian of the barrier terms
+end
+function TwiceDifferentiableConstraintsFunction(c!, jacobian!, h!, lx, ux, lc, uc)
+    b = ConstraintBounds(lx, ux, lc, uc)
+    TwiceDifferentiableConstraintsFunction{typeof(c!), typeof(jacobian!), typeof(h!), eltype(b)}(b, c!, jacobian!, h!)
+end
+
+## Utilities
+
+function symmetrize(l, u)
+    if isempty(l) && !isempty(u)
+        l = fill!(similar(u), -Inf)
+    end
+    if !isempty(l) && isempty(u)
+        u = fill!(similar(l), Inf)
+    end
+    # TODO: change to indices?
+    size(l) == size(u) || throw(DimensionMismatch("bounds arrays must be consistent, got sizes $(size(l)) and $(size(u))"))
+    _symmetrize(l, u)
+end
+_symmetrize{T,N}(l::AbstractArray{T,N}, u::AbstractArray{T,N}) = l, u
+_symmetrize(l::Vector{Any}, u::Vector{Any}) = _symm(l, u)
+_symmetrize(l, u) = _symm(l, u)
+
+# Designed to ensure that bounds supplied as [] don't cause
+# unnecessary broadening of the eltype. Note this isn't type-stable; if
+# the user cares, it can be avoided by supplying the same concrete
+# type for both l and u.
+function _symm(l, u)
+    if isempty(l) && isempty(u)
+        if eltype(l) == Any
+            # prevent promotion from returning eltype Any
+            l = Array{Union{}}(0)
+        end
+        if eltype(u) == Any
+            u = Array{Union{}}(0)
+        end
+    end
+    promote(l, u)
+end
+
+"""
+    parse_constraints(T, l, u, split_signed=false) -> eq, val, ineq, σ, b, [iz, σz, bz]
+
+From user-supplied constraints of the form
+
+    l_i ≤  v_i  ≤ u_i
+
+(which include both inequality and equality constraints, the latter
+when `l_i == u_i`), convert into the following representation:
+
+    - `eq`, a vector of the indices for which `l[eq] == u[eq]`
+    - `val = l[eq] = u[eq]`
+    - `ineq`, `σ`, and `b` such that the inequality constraints can be written as
+             σ[k]*(v[ineq[k]] - b[k]) ≥ 0
+       where `σ[k] = ±1`.
+    - optionally (with `split_signed=true`), return an index-vector
+      `iz` of entries where one of `l`, `u` is zero, along with
+      whether the constraint is `≥ 0` (σz=+1) or `≤ 0` (σz=-1). Such
+      are removed from `ineq`, `σ`, and `b`. For coordinate variables
+      this can be used to reduce the number of slack variables needed,
+      since when one of the bounds is 0, the variable itself *is* a
+      slack variable.
+
+Note that since the same `v_i` might have both lower and upper bounds,
+`ineq` might have the same index twice (once with `σ`=-1 and once with `σ`=1).
+
+Supplying `±Inf` for elements of `l` and/or `u` implies that `v_i` is
+unbounded in the corresponding direction. In such cases there is no
+corresponding entry in `ineq`/`σ`/`b`.
+
+T is the element-type of the non-Int outputs
+"""
+function parse_constraints{T}(::Type{T}, l, u, split_signed::Bool=false)
+    size(l) == size(u) || throw(DimensionMismatch("l and u must be the same size, got $(size(l)) and $(size(u))"))
+    eq, ineq, iz = Int[], Int[], Int[]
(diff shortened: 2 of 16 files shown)


johnmyleswhite:
Tim Holy for BDFL!

timholy:
Noticed that @tkelman doesn't watch this repo, and that he provided very useful feedback on my previous attempt.

tkelman:
is this a primal algorithm or primal-dual?

pkofod:
Wonderful. We need more positive emoji-"like"-types I see.
> argument order may be somewhat confused; in some places I adopt mutated-first, in others (where I'm trying to match some preexisting API) it's last. It's my understanding that Optim will eventually migrate to mutated-first, but perhaps this is too inconsistent now

Mutated first seems quite standard, so I'll be happy to go through the code and clean it up. Would it be annoying if I did this now?

> it supports both equality and inequality constraints
:+1: 

I'll have a closer look one of these days.

timholy:
> is this a primal algorithm or primal-dual?

I'm so glad you asked, because as I've learned more I've grown increasingly confused about this point. I form the KKT equations and solve them, so I suspect that many would call this primal-dual. However, the matrices in the primal-dual equations are asymmetric (e.g., [the equation after eq. 5](https://en.wikipedia.org/wiki/Interior_point_method)), and for generic nonlinear optimization that's problematic because you usually want to enforce positive-definiteness somewhere. So, at least some algorithms (e.g., Ipopt and this one) use a step of elimination to solve for certain constraints, which makes the reduced problem symmetric and amenable to being forced to be positive-definite (e.g., see Wächter & Biegler 2006, "On the implementation of an interior-point filter line-search algorithm for large-scale nonlinear programming," eq. 11). From what I can tell, this basically converts the primal-dual equation back into something that's virtually a primal one---the resulting equation is almost identical to what you'd get if you went the primal route in the first place.

Can you clear this up?

tkelman:
I think you've partially misread the Ipopt paper. It doesn't force or require anything to be positive definite - if it did, it would do a very poor job of solving non convex problems. It's not really constraints that are being eliminated, but complementarity rows of the Newton step direction of the KKT conditions. The KKT linear system that Ipopt solves is symmetric, but indefinite. It requires using a sparse Bunch Kaufman factorization in order to obtain the inertia of the KKT system at each iteration. It does a perturbation if the inertia does not give n positive and m negative eigenvalues, where n is the number of primal variables and m is the number of constraints.

The main difference between primal and primal-dual interior point, if I'm remembering the literature right, is the treatment of the barrier term for inequalities. If you replace inequalities with primal objective barrier terms before posing the KKT conditions and forming the Newton system, you have a primal method and the barrier term enters the objective Hessian in a way that squares the slack. For a primal dual method, you state the KKT conditions before approximating the inequalities with barrier terms. Instead you relax the complementarity condition to equal the barrier parameter instead of exactly zero. The contribution to the objective Hessian due to the barrier terms (which here aren't actually used in the derivation, it's purely a descriptive analogy) then depends on both the slack and dual variables. This is equivalent at a KKT point, but gives a different path to get there from an initial guess. Typically primal dual follows the central path better, and gives a Newton system that isn't quite as badly conditioned.

timholy:
> Tim Holy for BDFL!

I was planning to decline, but given the results of the last few hours I might take you up on that offer, if it's still open. :hushed: 

@tkelman, thanks for diving in. I'm learning a lot by thinking through your points.

> It's not really constraints that are being eliminated, but complementarity rows of the Newton step direction of the KKT conditions.

Agreed. I simply meant it in the sense that you've introduced additional variables, and they too are subject to "constraints" (the perturbed complementarity condition). Also agreed my choice of words could be confusing, since there are also user-supplied constraints that are in a distinct category.

> It doesn't force or require anything to be positive definite - if it did, it would do a very poor job of solving non convex problems.
>  It does a perturbation if the inertia does not give n positive and m negative eigenvalues, where n is the number of primal variables and m is the number of constraints.

OK, let's apply that to a simple non-convex problem, `f(x) = -x^2 + x^4` with no constraints, so we have `n=1` and `m=0`. The second derivative (hessian) is negative over the region `|x| <= 1/sqrt(6)`. Ipopt will "perturb" the hessian until it has 1 positive "eigenvalue" (and there's only one eigenvalue in this case). Now, in this case it's very clear that the negativity of the hessian has nothing to do with roundoff error, so the "perturbation" will be *large*, bigger than the hessian itself. That's what I mean by "forcing nonnegativity." (I've also argued [elsewhere](https://github.com/JuliaOpt/Optim.jl/issues/153#issuecomment-161268535) that this very common approach to generating nonnegativity---adding a positive diagonal until the matrix is just barely positive-definite---has serious weaknesses, because it throws away useful information about the length scales over which your function is likely to change, effectively swapping "clearly negative" directions in which the objective changes over short length scales with "nearly flat" directions that require big jumps to produce changes in the objective. In my opinion we have a much better approach available in PositiveFactorizations.jl. But that's a separate discussion.)

I agree, though, that you can't impose positive-definiteness to the total problem, and that the equality constraints are what prevent that. In this PR I only impose nonnegativity on the hessian for the objective + terms that come from the inequalities. I don't have a convergence proof yet, and I wonder if I should somehow project out the components of the equality constraints first (see the discussion in the Ipopt paper following eq. 12). That's totally doable (at least for dense problems), but before worrying too much about that, I thought I'd see how the simple version fared in tests.

As far as the relationship between the primal problem and the primal-dual problem, AFAICT here's the difference: adopting the problem and approximately the notation on [this page](https://en.wikipedia.org/wiki/Interior_point_method) (i.e., temporarily ignoring equality constraints), the Newton equation from following the primal route is
primal
and the one from the dual route, after solving for `Δλ` and substituting it into the other equation, is
dual
(Note the wikipedia page is a little misleading in calling `W` the hessian matrix of `B`, it's really the coefficient of `Δx` in the first-order expansion of their eq 5, which is not the same thing.) If you apply the complementarity condition to this, a short calculation shows that these are the same. So AFAICT the dual method is simply a "lagged primal" update, where the complementarity condition is not exactly satisfied at all stages of the update but is instead "frozen" at the base point (and even the base point might not satisfy it exactly). So to the extent that the dual method is better, it arises from the "ability" to not exactly satisfy the perturbed complementarity equation.

tkelman:
Ah good point, the interpretation of ensuring positive-definiteness of the Hessian block "projected onto the null space of the constraint Jacobian" works. I keep forgetting about that way of looking at it because that's not how it's calculated, unless you look at the unconstrained case.

> In my opinion we have a much better approach available in PositiveFactorizations.jl

With apologies for my skepticism, but I'd personally like to see a convergence proof make it through peer review, and/or demonstrated results across common test sets, before being totally sold on the approach. Pure Julia implementation, abstraction and API design is hard enough work that simultaneous innovation on the algorithm side is a bit beyond my own risk tolerance when comparing to known-quantity existing solvers, but that's coming from someone who isn't in academia any more. You must have a motivating application that you've been developing this for and using it on already, right? Would be interesting if you could point to any details about the kinds of problems where this has been working well for you, in terms of size, convexity, number of constraints, Hessian and Jacobian structure, etc.

Looks like all the Hessian and Jacobian fields here are dense matrices? I hope that's a planned-to-be-temporary limitation?

tkelman:
> Note the wikipedia page is a little misleading in calling `W` the hessian matrix of `B`, it's really the coefficient of `Δx` in the first-order expansion of their eq 5, which is not the same thing.

How do they differ?

timholy:
> With apologies for my skepticism, but I'd personally like to see a convergence proof make it through peer review, and/or demonstrated results across common test sets, before being totally sold on the approach. 

I want those things too, and hopefully I'll get to it eventually. But surely you understand the problem? After correction, a Hessian `[-k 0; 0 ϵ]` becomes `[ϵ 0; 0 k+2ϵ]`, which means that you're modeling the function as having strong curvature in `x_2` and weak curvature in `x_1` whereas the reality is that it has strong curvature in `x_1` and weak curvature in `x_2`. So your Newton step will be huge in `x_1`, and then you'll be puzzled about why the function is behaving entirely differently from what you expected---now you have a huge backtrack on your hands. There's no world in which that's a good thing. That doesn't prove that my approach is better, of course, just that the common approach has *obvious* limitations and there's a ton of room for doing better.

> How do they differ?

It doesn't yet have the `JDJ^T` term, that comes about by solving and substituting for `Δλ`.

timholy:
> Looks like all the Hessian and Jacobian fields here are dense matrices? I hope that's a planned-to-be-temporary limitation?

AFAICT that's true of all of Optim. Subject of another PR, and it doesn't have to be me. I use L-BFGS, so none of this is relevant anyway. I'm just trying to get something to pass muster, and on #50 people didn't like the fact that it wasn't like the conventional Newton approach. So I'm starting with Newton, even though I have no intention of using it for heavy-duty stuff.

tkelman:
> It doesn't yet have the `JDJ^T` term, that comes about by solving and substituting for `Δλ`.

You'll get that term from the gradient of `A^T λ` since `A` is a function of x, no?

> I use L-BFGS, so none of this is relevant anyway

For constrained problems too?

timholy:
> You'll get that term from the gradient of A^T λ since A is a function of x, no?

No, that gives you the `∇∇c` term.

> For constrained problems too?

Yes.

timholy:
OK, I'd say this is no longer WIP: it seems to be working quite well on many test problems. I used [CUTEst.jl](https://github.com/JuliaSmoothOptimizers/CUTEst.jl) and selected problems of 100 variables or less (and 100 constraints or less) that had second derivatives. Out of 293 such problems, it errored on 4 (I haven't yet dug into why), failed to converge on 16, and converged on the rest. On 32 problems, it found a better minimum than the one recorded in the CUTEst database, sometimes by quite a lot (possibly an error in the database, of course). Code is https://github.com/timholy/OptimTests.jl. I don't know if you want to incorporate that here, or exactly how you want this tested, but at least now this has been tested pretty thoroughly even if the code isn't part of this PR.

This isn't perfect, but I think it's good enough to be useful to folks.

pkofod:
> OK, I'd say this is no longer WIP: it seems to be working quite well on many test problems. I used CUTEst.jl and selected problems of 100 variables or less (and 100 constraints or less) that had second derivatives. Out of 293 such problems, it errored on 4 (I haven't yet dug into why), failed to converge on 16, and converged on the rest. On 32 problems, it found a better minimum than the one recorded in the CUTEst database, sometimes by quite a lot (possibly an error in the database, of course). Code is https://github.com/timholy/OptimTests.jl. I don't know if you want to incorporate that here, or exactly how you want this tested, but at least now this has been tested pretty thoroughly even if the code isn't part of this PR.
> 
> This isn't perfect, but I think it's good enough to be useful to folks.

Do you have a list of the problems, or something I can run to generate it? I think we should just test a few problems in this package, and then have "benchmarks/extensive testing" in another package (probably under JuliaOpt). It relates to #259 that I never got around to finish, because I didn't really get around to deciding if it should be in Optim or not. I guess I think it shouldn't. Would make sense to collect it in the same package as these tests - should that package be `OptimTests.jl`?

timholy:
I think we're on the same page: a few pure-Julia tests in Optim, and then a bigger package of tests called OptimTests (which can depend on stuff like CUTEst, MathProgBase, Ipopt, etc, if we want to make comparisons). I tried to create OptimTests under the JuliaOpt umbrella but I don't have adequate privileges (ping @mlubin or someone). I'll be happy to transfer it whenever that's possible.

I'm still dinking around with this, so I put the RFC back on simply to discourage merging.

mlubin:
Done

pkofod:
@timholy I hope it's okay I simply copy+pasted your package into JuliaOpt.

pkofod:
I just saw the lines added/removed up top. Minus 5 means you didn't remove Fminbox and added the IPNewton-optimizers, but simply added the latter. Is it your intention to keep Fminbox?

timholy:
I think that once this is ready fminbox can be deprecated, but that will take multiple PRs. I want to get *something* merged before going crazy with multiple methods, e.g., the gradient-only methods that would be required before all the functionality of fminbox is subsumed.

timholy:
OK, now I've done a head-to-head against Ipopt. (I'll push the test code later.) Main conclusions:
- Out of 293 problems with <100 vars/constraints, and a feasible initial point, there are 8 where Ipopt finds a solution that satisfies the equality constraints but this does not.
- If you omit those 8 problems, then the cumulative histogram of the difference in objective values looks like this:
objdiff_reg
Things to the right are problems we do better on; things to the left are ones where Ipopt does better. Obviously the main story is that we agree most of the time, but there are ~15% of problems where we are worse and ~5% of problems where we are better.

- To evaluate progress towards the solution, if both Optim & Ipopt used at least `n` iterations before convergence, I compute the fraction of those iterations for which our objective function value was lower than Ipopt's. This gives a measure of how well we were making progress towards the minimum; it isn't a perfect measure because some problems start out violating the equality constraints, but good enough for the moment... Here's the cumulative histogram:
fracbetter_reg

You can see that on balance we actually do slightly better (and for ~20% of the problems we're better on every iteration), but it's a near-wash in terms of typical behavior.

pkofod:
(29.960126512029458 isapprox 29.7) == false


(just need to loosen the tolerance ;))

timholy:
I had actually changed the behavior of how initialization takes place, so that test needed updating.

I just need to add some pure-julia tests and then I think this might be ready either for a serious code-level review or for merging.

anriseth:
@cortner How does this fit in with https://github.com/cortner/ConstrainedOptim.jl?

cortner:
> How does this fit in with https://github.com/cortner/ConstrainedOptim.jl?

could be complementary, could be unnecessary now; I'd have to look in more detail about how much Tim Holy's code does and how efficiently for the problems I am interested in. @pkofod  actually has done more work than I did on `ConstrainedOptim` as I have only fairly limited needs myself.

pkofod:
Well, that package is not tagged for a reason. Whatever is in there is pretty much (speaking for my own contributions of course) hacked together on the fly, as I briefly needed the functionality in there, and wanted to write it up "for fun". When this PR is merged, that package can still live it's own little life where we leverage the constraints API added here.

timholy:
This handles both equality and inequality constraints, and uses an interior-point method. That said, I didn't know/had forgotten about that package.

cortner:
Do you treat linear equality constraints by projecting them out?

timholy:
@tkelman, I looked into the question of the differences between what I'd call "primal with slack" vs "primal-dual". If you have access to Nodecal & Wright, I'm talking about comparing the approaches in eq. 19.12 with 19.13 (primal-dual) vs 19.14 (primal-with-slack). I considered the example in N&W (19.1, in section 19.6), which has the advantage that the objective function is quadratic and positive-definite so we don't need to worry about backtracking, etc, as long as we use a fraction-to-the-boundary rule. Note that this is the very example they use to argue for the "badness" of primal methods.

The code and results are in https://gist.github.com/timholy/8fbbb182066d5ecc939e472d12539d73 (for simplicity, I ignored the upper bounds because they don't really matter for this problem). The bottom line seems to be that there's at best a subtle advantage for primal-dual methods **if** you use the asymmetric fraction-to-the-boundary rule. With the symmetric rule, the primal-dual's path is weirder because it takes so long to set λ appropriately, and then it jumps wildly when it realizes it needs to change it. With the asymmetric rule, it only seems to help during the initial phase at that moment where you're about to fall into the minimum; it doesn't seem to help when you're far from the minimum, nor does it help at later stages when you decrease the barrier penalty.

Note that in the book they also point out (right above eq. 19.51) that the community was also convinced (for many years) that primal was worse because of a conditioning problem, but that this isn't actually correct. To me it seems this issue is somewhere between "maaaybe there's an advantage" and folks simply drinking too much koolaid. In either case, the big gain comes from having slack variables, which both primal and primal-dual methods can use (although it seems that this was realized only retrospectively for primal methods, which might account for much of the confusion); the differences between primal and primal-dual (in terms of how you form the Hessian) seem pretty piddly by comparison.

timholy:
> Do you treat linear equality constraints by projecting them out?

There's a separation between "coordinate variables" and "general constraints," but no distinction between linear constraints and other types of constraints.

timholy:
Status update: *lots* of changes. Here's how we compare in terms of objective value to Ipopt:
objdiff
We agree on ~90% of problems, and each is better than the other on another 5%. Looks like a wash to me.

Our progress on a per-iteration basis compared to Ipopt (plots the cumulative histogram of the fraction of iterations on which we're ahead):
fracbetter
Perhaps ever-so-slightly worse than Ipopt, but in the same ballpark.

Total number of iterations:
iterations
Both were set to a maximum of 3000 iterations. We're similar, but we do have more problems that reached the maximum number of iterations without converging. In cases I've checked, the algorithm actually found the right minimum but it didn't detect that it had converged. Some of these problems also don't converge if you run them in the unconstrained suite, so this isn't very specific to this PR.

While it's been a lot of work to get this far, I'm pretty stoked that with ~1month of effort it looks like we've got something that (for dense problems) does not look too bad compared to the industry-standard solution. There are several spots (particularly the linesearch) where we are also noteworthily simpler than Ipopt, which I think is a good thing. I haven't benchmarked for speed, but I'm not going to worry about that yet. I'll focus on cleanup and test coverage and then merge this, and then work on speed if my own practical problems require it.

anriseth:
Status update on this PR @timholy ?

pkofod:
I wonder why the number of iterations in fminbox differs between the linux and mac build...

pkofod:
@timholy is there anything specific you need to clean up? Do tell if you need a hand, so we don't end up in another rebase from hell!

pkofod:
When you're ready I suggest you squash some of the 40 commits. This will make handling of the merge conflicts easier I think. The funny thing is that the pr that introduced the conflict is almost orthogonal to what happens here, but I still end up with weird conflicts when trying to rebase

src/Optim.jl
```diff
@@ -17,8 +17,13 @@ module Optim
            Base.setindex!
 
     export optimize,
```
pkofod: any reason `ContraintBounds` are not exported?

src/types.jl
```diff
@@ -228,3 +266,265 @@ function TwiceDifferentiableFunction(f::Function,
...
+        if li == ui
+            push!(eq, i)
+            push!(val, ui)
+        else
+            if isfinite(li)
+                push!(ineq, i)
+                push!(σ, 1)
+                push!(b, li)
+            end
+            ui = u[i]
```
anriseth: It seems like `ui` is already defined in line 472?

musm:
would be awesome :)

pkofod:
> would be awesome :)

Unless @timholy decides to finish this himself (ABC, Tim, Always Be Closing!),  I wouldn't say it's improbable that I merge some version of this during 2017. We've cleared some older issues recently, and after a more proper loosening of types, I think this is next on the ticket.

timholy:
I'd be happy with that. I haven't merged it because the natural next step is to implement this for something besides Newton (and it's what my lab would actually use). I started looking into that, and decided I didn't have the month to spare to tackle that :smile:.

pkofod:
> I haven't merged it because the natural next step is to implement this for something besides Newton (and it's what my lab would actually use).

What do you mean by the comment in parentheses? That your lab would use it if something besides Newton was implemented, or that your lab is using this as is?

timholy:
No, we're using the old `teh/constrained` branch, which supports more methods but has a much worse implementation.

pkofod:
> No, we're using the old teh/constrained branch, which supports more methods but has a much worse implementation.

Alright, I follow. I think the first step will be up resolve conflicts (hopefully not as horrible as teh/constrained!!!), and merge this. Then we can always play around with other versions later.

anriseth:
The functionality from this PR is now available at https://github.com/JuliaNLSolvers/ConstrainedOptim.jl

I encourage everyone who are interested to try it out and let us know how it works for you. 

The plan is to add some more constrained optimization code that is lying around to ConstrainedOptim.jl, and then move the code back into Optim at some point in the future.

Issue #243, opened 2016-07-12
Keywords...


pkofod:
I thought I had created an issue like this, but apparently I have not.

I was wondering what comments or objections people would have to removing the keywords, and forcing people to use the (dispatch based) `OptimizationOptions` for choosing options. I know it's a few more characters, but if you want to choose a solver and add a keyword, you have to write `method =` so really it isn't that many additional keystrokes. 

From a developer's point of view, I think it's much easier to ditch them, but if they're deemed very helpful, I guess they should stay. Would love to hear other people's opinion on this.

anriseth:
This change will be good. I think it is useful to put all configuration of solvers in a options-objects, without too many other approaches to confuse users.

As a long-term inspiration, I like the approach the PETSc project has taken. You have full control of a wide range of solvers with special keywords via the same interface.

pkofod:
> As a long-term inspiration, I like the approach the PETSc project has taken. You have full control of a wide range of solvers with special keywords via the same interface.

I'll take a look at PETSc. Thanks for your comment!

cortner:
I like quick one-line calls of optimisers.  If the keyword arguments for the `optimise` calls go, then will it still be possible to do something similar but just with kwargs for OptimisationOptions?

pkofod:
I found a way to clean up the code that handles the kwargs, so I let it stay for now, see #304. However, I did you ahead and make the `OptimizationOptions`->`Optim.Options` change. Deprecations should be in place. Kwargs might go later. Closing for now.

Issue #285, opened 2016-09-23
Remove `!` from `linesearch!` keyword in `Optimizer` objects


jiahao:
They are the cause of inscrutable whitespace-sensitive input errors:

```
result = optimize(rosenbrock, zeros(2), GradientDescent(linesearch!=hz_linesearch!
))

LoadError: UndefVarError: linesearch not defined
while loading In[10], in expression starting on line 1
```

It took students a long time to realize the problem could be fixed with

```
GradientDescent((linesearch!)=Optim.hz_linesearch!)
```

KristofferC:
I have gotten confused by this as well. No point in having it in the keyword imorgon.

pkofod:
> having it in the keyword imorgon.

or the day after that... :)

KristofferC:
Swedish autocomplete hehe.

jiahao:
@pkofod I agree that the whitespace convention you describe will avoid the problem of whitespace-sensitive parsing. However, I don't agree that it is a "solution". The convention in Julia is very much to provide keyword arguments without the whitespace; see, for example, [`test/statistics.jl`](https://github.com/JuliaLang/julia/blob/792ded802743830044198d2f525bded1f1c0bfb6/test/statistics.jl#L80) in Base.

pkofod:
Okay, wasn't aware. Anyway, we shouldn't have an API that is easy to get wrong by missing a space. So convention or not, I vote for removing the !.

pkofod:
Sooo... What is the best way to deprecate a character in a keyword? https://discourse.julialang.org/t/deprecating-usage-of-keywords/978

KristofferC:
I posted an idea.

pkofod:
In master due to #304 - I'll tag soon. Thanks for opening this @jiahao

PR #304, opened 2016-11-11
Options and Keywords clean-up

### LINKED ISSUE #235: No method for `optimize(f, x, OptimizationOptions(...))`
No method for `optimize(f, x, OptimizationOptions(...))`,  `optimize(f, g!, x, OptimizationOptions(...))` or for `optimize(f, g!, h!, x, OptimizationOptions(...))`. These should just default to Nelder-Mead, BFGS, and Newton as usual.

pkofod:
Ref #243 also fixed #235 

- [x] Add tests for fix in #235 
- [x] See what do to deprecation-wise
- [x] Deprecate `linesearch!` and `precondprep!` in favor of `linesearch` and `precondprep`
- [x] <s>Use "manual" deprecation to deprecate the keyword methods so there is no deprecation warning given when keywords are not used.</s> parse and pass keywords in a better way

**Summary**:
This PR removes ! in keyword argument names to avoid the situation where a user write `kw!=kw_value` and gets an error. In this PR the user would write `kw=kw_value` and be fine. We don't want an interface that depends on white space, when white space is generally not relied upon in Julia.

Additionally, this PR seeks to remove the exported `OptimizationOptions` in favor of the unexported `(Optim.)Options`. This change also seeks to remove all keyword arguments that used to be passed to `optimize(args...;kwargs...)`, such that it is required for a user to construct the `Options` and pass it to `optimize` as a positional argument. This seems more in line with the general use of dispatch in Julia, and allows us to parameterize Options later if we wish, without having to maintain two different APIs. This interface might seem annoying at first (you *have* to write `Optim.Options(` to set an option), but since users had to write `method=...` before to be able to use the other keywords, the overall keystroke count is basically the same.


pkofod:
Does anyone object to this change? I don't want to spend time on the deprecations if people are against it anyway.

anriseth:
I think the proposed change is good

pkofod:
Took care of #285  while cleaning up options in the latest commit.

docs/src/dev/contributing.md
```diff
@@ -18,12 +18,12 @@ that does the actual work. Say you want to contribute a solver called
 
 ```
 immutable Minim{T} <: Optimizer
-    linesearch!::Function
+    linesearch::Function
```
anriseth: I think we should still have the `linesearch!` in the type fields. The issue we want to fix is the problem that arises in the keywords.

src/accelerated_gradient_descent.jl
```diff
@@ -8,11 +8,11 @@
 
 
 immutable AcceleratedGradientDescent <: Optimizer
-    linesearch!::Function
+    linesearch::Function
```
anriseth: Keep it `linesearch!` ?

src/accelerated_gradient_descent.jl
```diff
@@ -61,7 +61,7 @@ function update_state!{T}(d, state::AcceleratedGradientDescentState{T}, method::
 
     # Determine the distance of movement along the search line
     state.alpha, f_update, g_update =
-      method.linesearch!(d, state.x, state.s, state.x_ls, state.g_ls, state.lsr, state.alpha, state.mayterminate)
+      method.linesearch(d, state.x, state.s, state.x_ls, state.g_ls, state.lsr, state.alpha, state.mayterminate)
```
anriseth: Keep it `linesearch!`? This is a method that should tell the person reading the code that some of the objects in its argument will be changed.

src/deprecate.jl
```diff
@@ -18,8 +18,8 @@ for name in names(LineSearches)
     eval(:(@deprecate $name LineSearches.$name))
 end
 
-@deprecate hz_linesearch! LineSearches.hagerzhang!
-@deprecate mt_linesearch! LineSearches.morethuente!
-@deprecate interpolating_linesearch! LineSearches.strongwolfe!
-@deprecate backtracking_linesearch! LineSearches.backtracking!
-@deprecate interpbacktracking_linesearch! LineSearches.interpbacktracking!
+@deprecate hz_linesearch LineSearches.hagerzhang!
```
pkofod: this is a mistake...

pkofod:
> I think we should keep the type objects in the optimizers as linesearch! still. The issue was to have the ! in kwargs

Yes, it is indeed a problem to have the ! in kwargs only, but isn't it odd to have a keyword for a field being different than the actual field name? I guess it's okay if spelled out somewhere, it would probably just surprise most people if they looked at the type definition, saw a field, and then couldn't set that field with the keyword spelled the same.

timholy:
FYI, going forward it might be better to declare these types like this:
```jl
immutable AcceleratedGradientDescent{LS<:Function} <: Optimizer
    linesearch!::LS
end
```
The reason is that in julia 0.5 and above, each function has its own type. Using the type parameter means this type is fully type-stable, and you can even dispatch like this:
```jl
foo(o::Optimizer{typeof(hagerzhang!)}) = 1
foo(o::Optimizer) = 2
```

anriseth:
> Yes, it is indeed a problem to have the ! in kwargs only, but isn't it odd to have a keyword for a field being different than the actual field name? I guess it's okay if spelled out somewhere, it would probably just surprise most people if they looked at the type definition, saw a field, and then couldn't set that field with the keyword spelled the same.

Both options complicate things I guess. My perspective is the following: To understand what a solver does, I will look at the code in the `optimize` method. And if I see `method.linesearch!`, I understand that the call to line search may change my arguments. If it only says `method.linesearch`, I would expect it not to. But the line searches will change the LineSearchResults object, and potentially df (if they include evaluation counts).

pkofod:
> FYI, going forward it might be better to declare these types like this:
> 
> immutable AcceleratedGradientDescent{LS<:Function} <: Optimizer
>     linesearch!::LS
> end
> 
> The reason is that in julia 0.5 and above, each function has its own type. Using the type parameter means this type is fully type-stable, and you can even dispatch like this:
> 
> foo(o::Optimizer{typeof(hagerzhang!)}) = 1
> foo(o::Optimizer) = 2

Thank you for the comment. I just realized that fact yesterday, I only thought it applied to anonymous functions for some reason. I should probably revisit some of the changes made to nelder mead over the summer, where I think I might have over-complicated some things for this very reason.

pkofod:
> Both options complicate things I guess. My perspective is the following: To understand what a solver does, I will look at the code in the optimize method. And if I see method.linesearch!, I understand that the call to line search may change my arguments. If it only says method.linesearch, I would expect it not to. But the line searches will change the LineSearchResults object, and potentially df (if they include evaluation counts).

I'll add the ! to the field names, and only change the keywords.

pkofod:
precondprep should suffer the same fate (note to self)

KristofferC:
Might be worth thinking if deprecated keywords should just be warned about once per session.

pkofod:
> Might be worth thinking if deprecated keywords should just be warned about once per session.

True. On the other hand, I would prefer if people fixed it right away ;)

pkofod:
@timholy @johnmyleswhite is this change OK with you? I'm thinking about the `OptimizationOptions -> Optim.Options` and removal of the keyword args.

pkofod:
Removed the "controversial" keywords deprecation, so I'm just going to merge this. Keywords can be deprecated by changing two lines, so we can still do it by next tag if someone feels like we should. The new "low maintenance" way of doing it is fine by me at least.

PR #311, opened 2016-12-06
fminbox(): save convergence status to results

alyst:


pkofod:
Just to be 100%, the problem is that it will exit with the convergence declarations of the second to last iteration unless it converges in the first iteration, correct? Thanks!

alyst:
Thanks!

Issue #198, opened 2016-05-15
Inappropriate StackOverflowError with unexpected array type


veita:
When an `Array{Float64,2}` instead of a `Array{Float64,1}` is passed to the `optimize` function a `StackOverflowError` is thrown. A more helpful exception with a hint to the true reason for the error would be nice.

``` julia
using Optim

function f(x::Vector)
    return (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
end

optimize(f, [0.0 0.0])
```

```
LoadError: StackOverflowError:
  in call at C:\Users\alex\.julia\v0.4\Optim\src\types.jl:193
  in optimize at C:\Users\alex\.julia\v0.4\Optim\src\optimize.jl:146
  in optimize at C:\Users\alex\.julia\v0.4\Optim\src\optimize.jl:150
 while loading D:\prj\julia\optimizebug.jl, in expression starting on line 7
```

Julia version 0.4.5 (2016-03-18 00:58 UTC)

johnmyleswhite:
At one point we were making an effort to allow passing in arrays of arbitrary order. Would be good to see if we can restore some of that if it got lost in the API redesign.

veita:
Additional note:

```julia
optimize(f, [0.0 0.0]')
```
also fails with a `StackOverflowError` error.

Evizero:
I suppose it is because the fallback function is called over and over again since no specialized method applies. Probably lines like this are the culprit: https://github.com/JuliaOpt/Optim.jl/blob/master/src/nelder_mead.jl#L53 . I do think I had a reason for narrowing it down to vectors here, especially because I didn't consistently do so: https://github.com/JuliaOpt/Optim.jl/blob/master/src/gradient_descent.jl#L36, but I can't think of a fundamental one. I guess that decision should be revisited

Regardless, a Stackoverflow is not a graceful way to deal with an unsupported parameter setting.

pkofod:
I don't see why this shouldn't be possible, and it should certainly have tests, as I was sure we allowed for this. Unless we can find a good reason, I think we (I, op, anyone) should restore the functionality. I've never had an obvious need for it myself, though I can imagine a few uses.

pkofod:
Although I really wish it would be handled by something like #163 . If op doesn't do it, I'll try to make a pr asap.

pkofod:
Okay, so I looked a bit into this.

[...]

PR #312, opened 2016-12-06
Fix problem dimensions inside the method

alyst:
Modify the methods to respect the problem dimensions, so that `df.f()`, `df.fg!()`, `df.g!()` are called with the correct `x`, `grad` and `hessian` argument types. E.g. currently `Fminbox` fails when I submit the problem with `x` being a matrix.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -6,26 +6,26 @@ iteration_limit_reached(r::OptimizationResults) = r.iteration_converged
 trace(r::OptimizationResults) = length(r.trace) > 0 ? r.trace : error("No trace in optimization results. To get a trace, run optimize() with store_trace = true.")
 
 function x_trace(r::UnivariateOptimizationResults)
-	tr = trace(r)
-	!haskey(tr[1].metadata, "minimizer") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
+    tr = trace(r)
+    !haskey(tr[1].metadata, "minimizer") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
     [ state.metadata["minimizer"] for state in tr ]
 end
 function x_lower_trace(r::UnivariateOptimizationResults)
-	tr = trace(r)
-	!haskey(tr[1].metadata, "x_lower") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
+    tr = trace(r)
+    !haskey(tr[1].metadata, "x_lower") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
     [ state.metadata["x_lower"] for state in tr ]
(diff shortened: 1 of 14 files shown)


pkofod:
Wonderful! I've been meaning to get this working again - also ref #198, but havn't had the time. This is really appreciated.

I've restarted travis.

Two things before I merge.

1) Could you add an entry to the documentation about this? Just a simple section where you show that either input will return the same thing. The relevant file is https://github.com/JuliaOpt/Optim.jl/blob/ccc51d73621988265f77b5e42cf2393f444491ec/docs/src/user/minimization.md and you can add it just before box minimization.

2) We need some test to make sure this doesn't regress. I guess it would suffice to loop over the methods optimizing some simple problem (you can take one from the unconstrained problems module) solving it with the vector- and matrix-input and checking the results are identical. This will ensure a) that everything can run and b) it returns the (exact [besides randomized methods]) same thing.

Oh, and a tiny thing. Can you add a line to NEWS.md saying that x is no longer restricted to be a vector ref #198 .

Once again, thanks.

alyst:
@pkofod I'm a little bit busy these days, but your suggestions make total sense, I will update the PR in a week or two.

pkofod:
> @pkofod I'm a little bit busy these days, but your suggestions make total sense, I will update the PR in a week or two.

No worry, this is all voluntary. When you update this PR, I'll make sure to review it quickly.

pkofod:
I'll open a new PR in a bit, with the proposed changes. I'll keep your changes under your name, so you'll still get credit in the git history. Thanks for doing this.

pkofod:
thank you for the effort, but this is now superseded by #337

alyst:
@pkofod for the reference, #345 implemented (I don't know whether independently or not) parts of #312 that addresses using proper dimensions, but not all. E.g. the Hessian in Newton method. Also I have to note that while #345 was merged, it didn't introduce the test cases either.

pkofod:
Sure, but that PR was to remove warnings, it didn't claim to fix the bug you brought up. I just wanted the warnings on Travis to go away.

alyst:
@pkofod next time I would be more modest in the PR description ;)

pkofod:
Earlier your could do this. Then there were changes, and a bug was introduced. Then you opened this pr with a fix, but no tests. If I had merged and closed the issue, then the bug could silently creep in again. Providing tests is not really a horrible thing to ask for. I really do appreciate all contributions, but merging stuff without proper tests is what made it possible to introduce this  bug in the first place. In the mean time the old style constructors were removed on master, and I changed the lines that blocked Travis. I did not have this problem in mind at all when doing that.

pkofod:
Looking through my notes: I didn't add the hessian due to the backslash operator/factorization, which was part of the background for comment #5

PR #317, opened 2016-12-12
Test on nightly

tlnagy:
Currently tests aren't run on nightly so this can break downstream packages (like Gadfly)


pkofod:
It is nice to be able to follow along as v0.6 progresses, and I am aware of the macro-problem. I'll tag a new version shortly. Do note, thought, that supporting 3 versions is probably out of scope given the number of (and time constraints of) the contributors here.

PR #322, opened 2016-12-15
Fix some bugs of the tracing for Nelder-Mead algorithm.

hayatoikoma:
This pull request is intended to fix the tracing for Nelder-Mead algorithm. Currently, `show_trace` for Nelder-Mead is showing the same function values for all iterations because `state.f_lowest` is not updated. In addition to the fix of this bug, I have updated the variable names as suggested in this comment.
> f_x and f_x_previous should be renamed to nm_x, nm_x_previous and f_x should hold
-# f_lowest instead!

On a side note, `f_lowest` and `f_x` should hold the same values during the iterations, but I am updating only `f_lowest` by following how `trace!` is currently implemented. I have also removed `f_x_previous` because it was not used.

I hope this update reflects the intention of the comment.

### DIFF
--- a/src/nelder_mead.jl
+++ b/src/nelder_mead.jl
@@ -37,7 +37,6 @@ end
 FixedParameters(; α = 1.0, β = 2.0, γ = 0.5, δ = 0.5) = FixedParameters(α, β, γ, δ)
 parameters(P::FixedParameters, n::Integer) = (P.α, P.β, P.γ, P.δ)
 
-
 immutable NelderMead{Ts <: Simplexer, Tp <: NMParameters} <: Optimizer
     initial_simplex::Ts
     parameters::Tp
@@ -104,9 +103,6 @@ function Base.show(io::IO, t::OptimizationState{NelderMead})
     return
 end
 
-# f_x and f_x_previous should be renamed to nm_x, nm_x_previous and f_x should hold
-# f_lowest instead!
-
 type NelderMeadState{T, N}
     @add_generic_fields()
     m::Int64
@@ -118,7 +114,7 @@ type NelderMeadState{T, N}
     x_reflect::Array{T}
     x_cache::Array{T}
     f_simplex::Array{T}
-    f_x_previous::T
+    nm_x::T
     f_lowest::T
     i_order::Vector{Int64}
     α::T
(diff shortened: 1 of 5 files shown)


pkofod:
> Currently, show_trace for Nelder-Mead was showing the same function values for all iterations because state.f_lowest was not updated.

Well, that is certainly not intentional. Thanks for fixing this.

This means that in the current master, `Optim.f_trace(res)` is going to return a vector of identical number. I think it might make sense to add a test that checks that if the trace is stored, the values actually differ across iterations.

hayatoikoma:
I just added the test as you suggested.

The test is to check if `Optim.f_trace(res)` and `Optim.g_norm_trace(res)` change during iterations and make sure that `Optim.f_trace(res)` decreases. Do you think that it is a valid test?

pkofod:
> just added the test as you suggested.
> 
> The test is to check if Optim.f_trace(res) and Optim.g_norm_trace(res) change during iterations and make sure that Optim.f_trace(res) decreases. Do you think that it is a valid test?

I think it's fine. I am just reviewing it one last time, and then I'll merge. Thanks for doing this.

pkofod:
There is a tiny thing. It's a bit funny that we have `f_lowest` and `f_x` playing the same role. We just just remove `f_lowest` and use `f_x` instead. That would of course require people who read the code knows what `f_x` generally mean in this code base. Since the expected overhead in insignificant, I'm fine with it. A way to fix it would of course be to change all other occurrences of `f_x` in the package to `f_lowest` as this is more informative. Not for this PR though. Thanks for the contribution.

hayatoikoma:
Thank you for merging it.

Yes, that's the thing I mentioned in the first comment.
> On a side note, f_lowest and f_x should hold the same values during the iterations, but I am updating only f_lowest by following how trace! is currently implemented.

As you say, `f_lowest` may be more informative here. However, completely removing `state.f_x` in `nelder_mead.jl` may be cumbersome because `state.f_x` is called in `optimize` to store the result.

The current code looks fine to me, but do you prefer to remove `f_lowest` from this code? I can take care about it if it is a better thing to do.

pkofod:
> The current code looks fine to me, but do you prefer to remove f_lowest from this code? I can take care about it if it is a better thing to do.

Nope, leave it in. I don't care if there's a little overhead.

hayatoikoma:
OK. I'll leave it, then.
Thank you for clarifying it.

Issue #332, opened 2017-01-07
Optim.jl missing a function on Windows 32-bit


ChrisRackauckas:
Optim.jl is making one of my tests fail on Windows 32-bit 

https://ci.appveyor.com/project/ChrisRackauckas/diffeqparamestim-jl/build/job/jy632frdrd1weenv

```julia
ERROR: LoadError: MethodError: no method matching Optim.BFGSState{T}(::String, ::Int32, ::Array{Float64,1}, ::Float64, ::Int32, ::Int32, ::Int32, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Float64, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,2}, ::Array{Float64,2}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Float64, ::Bool, ::LineSearches.LineSearchResults{Float64})
Closest candidates are:
  Optim.BFGSState{T}{T}(::String, !Matched::Int64, ::Array{T,N}, ::T, !Matched::Int64, !Matched::Int64, !Matched::Int64, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::T, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::Array{T,N}, ::T, ::Bool, ::LineSearches.LineSearchResults{T}) at C:\Users\appveyor\.julia\v0.5\Optim\src\bfgs.jl:20
  Optim.BFGSState{T}{T}(::Any) at sysimg.jl:53
 in initial_state(::Optim.BFGS{LineSearches.#hagerzhang!,Optim.##38#40}, ::Optim.Options{Void}, ::Optim.DifferentiableFunction, ::Array{Float64,1}) at C:\Users\appveyor\.julia\v0.5\Optim\src\bfgs.jl:41
 in optimize(::Optim.DifferentiableFunction, ::Array{Float64,1}, ::Optim.BFGS{LineSearches.#hagerzhang!,Optim.##38#40}, ::Optim.Options{Void}) at C:\Users\appveyor\.julia\v0.5\Optim\src\optimize.jl:176
[...]

PR #333, opened 2017-01-07
Enable AppVeyor

ChrisRackauckas:
Optim.jl has failures on Windows which are undetected by the devs because Windows CI is not enabled. As a large package in the Julia package ecosystem which has no binary dependencies, it should be easy to test and support Windows. This PR adds an appveyor.yml file to do just that, but does not enable CI because I can't do that (you have to be in the org). I would be willing to take the CI burden if you want (Orgs cannot have CI for AppVeyor). (I would have to be added to the org to do that though.)

For current failures, see https://github.com/JuliaOpt/Optim.jl/issues/332

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,35 @@
+environment:
+  matrix:
+    - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
+    - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
+matrix:
+  allow_failures:
+    - JULIAVERSION: "julianightlies/bin/winnt/x86/julia-latest-win32.exe"
+    - JULIAVERSION: "julianightlies/bin/winnt/x64/julia-latest-win64.exe"
+branches:
+  only:
+    - master
+    - /release-.*/
+
+notifications:
+  - provider: Email
+    on_build_success: false
+    on_build_failure: false
+    on_build_status_changed: false
+
+install:
+# Download most recent Julia Windows binary
+  - ps: (new-object net.webclient).DownloadFile(
+        $("http://s3.amazonaws.com/"+$env:JULIAVERSION),
+        "C:\projects\julia-binary.exe")
+# Run installer silently, output to C:\projects\julia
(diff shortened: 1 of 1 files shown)


pkofod:
Good idea.

@blegat Sorry for not knowing everyone in the JuliaOpt organization, but you're not really a regular around Optim.jl, so while this change is uncontroversial, you may want to wait a bit before merging pull requests. If I saw a PR at StructJuMP.jl I would make sure to get a :+1: from a regular contributor before hitting "merge" even if I could.

I wasn't going to mention it, but I saw you also pushed a commit directly to master without going through the usual pull request channels. For a non-regular contributor it is quite unusual to push directly to master, and should be avoided.

pkofod:
Running now, but obviously failing on 32bit (that's the whole point of these PRs) https://ci.appveyor.com/project/blegat/optim-jl

blegat:
@pkofod Sorry about that, I wouldn't push commits or merge pull request like that if it wasn't README changes or CI-related. In this case, I merged the PR and added the appveyor's badge because it was needed for @ChrisRackauckas to continue investigating the issue and I thought in the cases you didn't want Optim to have appveyor you could remove it one the issue was fixed. Again, sorry if you felt that this wasn't right. If you want to host appveyor on your account, you can change the README's badge and think of my hosting as a temporary one.

ChrisRackauckas:
>@blegat Sorry for not knowing everyone in the JuliaOpt organization, but you're not really a regular around Optim.jl, so while this change is uncontroversial, you may want to wait a bit before merging pull requests. If I saw a PR at StructJuMP.jl I would make sure to get a 👍 from a regular contributor before hitting "merge" even if I could.

Well, there does tend to be an exception for maintenance. Updating test versions, fixing README typos, etc. I think it's accepted that everyone in the org shares those duties. I would think enabling Windows testing would fall under that (if it's uncontroversial. For a package with no binary deps, it's uncontroversial)

KristofferC:
> I think it's accepted that everyone in the org shares those duties.

No one is saying otherwise or that this was a controversial PR. However, someone with little involvement in a package previous development should simply not merge PRs without consulting other more active contributors. Pushing to master is taking it one step further. I don't think this has to be debated at length since it should be fairly obvious.

PR #335, opened 2017-01-08
AppVeyor with show trace test.

pkofod:


pkofod:
Go figure, it's a NaN not Inf

PR #336, opened 2017-01-08
BigFloat compatibility for univariate optimization, fixes #271

### LINKED ISSUE #271: typecasts needed for BigFloat compatibility
The Optim.Brent() and Optim.GoldenSection() methods currently don't work with BigFloat.

The problem is very easy to fix: rel_tol and abs_tol need typecasts at the very end, inside the  UnivariateOptimizationResults() constructor.

andrewclausen:
The univariate optimization code unintentionally assumed that the choice
variable was a Float64.  Specifically, the interfaces accommodated an
AbstractFloat, but then would crash when reporting the results because of
missing typecasts.

This patch introduces approriate typecasts and test cases.


pkofod:
Thanks for making this PR. However, I think the better thing to do is to drop the Float64 type in the results type. You can either change it to `::T` or just remove it altogether. If you pass in bigfloats, I don't think you would want a Float64 result back.

src/brent.jl
```diff
@@ -162,8 +162,8 @@ function optimize{F <: Function, T <: AbstractFloat}(
                                          iteration,
                                          iteration == iterations,
                                          converged,
-                                         rel_tol,
-                                         abs_tol,
+                                         Float64(rel_tol),
```
pkofod: Revert these and remove the `::Float64` in the results type in `src/types.jl`

src/golden_section.jl
```diff
@@ -103,8 +103,8 @@ function optimize{F<:Function, T <: AbstractFloat}(f::F, x_lower::T, x_upper::T,
                                          iteration,
                                          iteration == iterations,
                                          converged,
-                                         rel_tol,
-                                         abs_tol,
+                                         Float64(rel_tol),
```
pkofod: and these

src/optimize.jl
```diff
@@ -261,7 +261,7 @@ function optimize{F<:Function, T <: AbstractFloat}(f::F,
     if show_trace
         @printf "Iter     Function value   Gradient norm \n"
     end
-    optimize(f, Float64(lower), Float64(upper), method;
+    optimize(f, T(lower), T(upper), method;
```
pkofod: simply remove `T` here, as it's already there in the method signature

test/brent.jl
```diff
@@ -5,4 +5,11 @@ let
         @assert Optim.converged(results)
         @assert norm(Optim.minimizer(results) - prob.minimizers) < 1e-7
     end
+
```
pkofod: I think it's simpler if we have 
```julia
for (name, prob) in Optim.UnivariateProblems.examples
    for T in (Float64, BigFloat)
        results = optimize(prob.f, convert(Array{T}, prob.bounds)..., method = Brent())

        @assert Optim.converged(results)
        @assert norm(Optim.minimizer(results) - prob.minimizers) < 1e-7
    end
end
```
than repeating the test loops twice

src/brent.jl
andrewclausen: What about consistency between UnivariateOptimizationResults and MultivariateOptimizationResults?

Is there a good reason why MultivariateOptimizationResults uses ::Float64 for x_tol et al?

src/optimize.jl
andrewclausen: Oops, I meant to typecast rel_tol and abs_tol.

andrewclausen:
I made all the changes you requested, but I don't know how to make a new Pull Request (or revise the old one).  Can you pull?

(I'm new to github, but I've been using git for 10 years.)

KristofferC:
Pushing to the same branch (like you seem to have done here) will automatically update the Pull Request.

src/brent.jl
pkofod: Nope, those should be removed as well. However, I am not entirely certain that all multivariate solvers will have methods for BigFloat from a-z, so maybe that can be left for another PR.

pkofod:
Everything looks fine. Thanks! AppVeyor is failing for another reason that I have yet to fix (some barrier function issue on 32bit windows).

pkofod:
Thanks for addressing this issue.

Issue #329, opened 2016-12-26
Finite differences and autodiff - bake into *DifferentiableFunction?


pkofod:
I was having a first go at fixing a lot of the issues surrounding (coming up with an improved for) *DifferentiableFunctions, and I realized that maybe we want to remove the `autodiff` keyword from `Optim.Options`. Currently you can use finite differences by `DifferentiableFunction(f)`, but you have to pass a function and use `Optim.Options(..., autodiff=true, ...)` to use ForwardDiff. Wouldn't it make more sense to have something like `DifferentableFunction(f; autodiff = false)`? Going one step further, we might want to make an even more general interface so you could pass functions from any autodiff package (you might need to do some anonymous function magic to conform to the setup).

Edit: of course it could just as well go the other way: move the finite difference decision to `Optim.Options` if that is easier for people.

anriseth:
We're constructing a more "intelligent" objective function here, and to me it makes sense (and provides more flexibility?) to let the DifferentiableFunction object decide how the objective and derivatives should be approximated. This is not only for derivative approximation, but also approximations of expensive functions, expected values etc. This might go against some algorithms in the literature though, where the approximation method is part of the solver, and potentially "dynamic".

ChrisRackauckas:
>We're constructing a more "intelligent" objective function here, and to me it makes sense (and provides more flexibility?) to let the DifferentiableFunction object decide how the objective and derivatives should be approximated.

No, it can turn out to be less flexible since many times the reason to turn off autodifferentiation is because you know that your function cannot handle Julia's type system (say, your objective function is the loss between data and the solution to an ODE via Sundials). If the system tries to be too smart, it will error, meaning you'd have to place it in a try/catch block which can hinder performance. So it's best left as a boolean flag.

pkofod:
> No, it can turn out to be less flexible since many times the reason to turn off autodifferentiation is because you know that your function cannot handle Julia's type system (say, your objective function is the loss between data and the solution to an ODE via Sundials). If the system tries to be too smart, it will error, meaning you'd have to place it in a try/catch block which can hinder performance. So it's best left as a boolean flag.

[...]

Issue #306, opened 2016-11-14
OEP: more general state for user-supplied functions


timholy:
It would be lovely to be able to support sparse Hessians, Jacobians (for constrained optimization), etc. It would be good to move to an API where one can supply HessState and JacState objects of any kind. The user-supplied functions would update the state at the current point, and then such state objects would have to support `J*x` and `H\x` type of operations. Obviously `H` *could* be a dense `n`-by-`n` matrix, but it wouldn't have to be.

pkofod:
I will try to fit this in when I have a shot at the dispatch based function interface. I intend to target this after #304 , as quite a few open issues are related to this.

pkofod:
Currently #337 simply opens up the fields for storing gradients and hessians. This means that if you use the full constructor (with all the fields), you can input your own gradient type as storage as long as you've implemented
```
copy
copy!
getindex
vecdot(s::Array, g::G)
```
for your type `G`. For the Hessian you can in principle do the same, but currently you'll run into a wall when it gets passed to the `NewtonTrustRegion` solver, and in `Newton` you'd need to support a method that can be passed to `PositiveFactorizations`. I do intend to make a switch, such that you can simply do good old renegade Newton steps: `-H\g`. Then you just need your state to be supported by a `\` method.

I'm going to leave Newton alone for now, but I will try to make some stupid examples with the gradient, such as
```
type MyStupidGradient
    g
end
```
just to make sure it works all the way through. Some changes might have to be made over at `LineSearches.jl`, but I will find out :)

pkofod:
Should be possible with #337. If you create a `OnceDifferentiable` "manually" that is you also provide the `g` yourself, then this is possible. (the gradient storage is parametric now, as is the Hessian)

Issue #305, opened 2016-11-14
OEP: function-objects should keep track of their own call counts


timholy:
(OEP = Optim Enhancement Proposal) Rather than keeping track of `f_calls` and `g_calls` in the algorithm, it seems much cleaner to have `DifferentiableFunction` store these quantities, and every time you call `value(d, x)` and `gradient!(g, d, x)` the counters get updated. That way there is no chance you're not keeping track of each evaluation (e.g., #219).

pkofod:
I agree, and I know others do as well. We've had other suggestions for `value` and `gradient!`, but I actually like both.

pkofod:
I'd throw my time after this asap (as in asap, not open source work asap), but (and correct me if I'm mistaken) I've seen @johnmyleswhite oppose something similar earlier. Any opinions? If I remember correctly, you said you wanted to go in the opposite direction API wise. Care to elaborate? Or correct me?

Issue #287, opened 2016-09-25
Changes to DifferentiableFunction API


ahwillia:
Now that PR https://github.com/JuliaOpt/Optim.jl/pull/266 looks nearly complete, I'm hoping we can revisit issue https://github.com/JuliaOpt/Optim.jl/issues/156 and a few other additions or changes I've been thinking about. I know @johnmyleswhite mentioned that he would like to change the API in the past (not sure if that was meant in a broader sense or not).

**Minor / Low Priority:**
- As I brought up in https://github.com/JuliaOpt/Optim.jl/issues/156 - I think it would be nice to switch the order of the arguments so that `fg!(x, storage)` becomes `fg!(storage, x)` (making the mutated argument come first seems more consistent with Julian conventions and autodiff packages).
- I kind of like changing the names to `OnceDifferentiable` and `TwiceDifferentiable` for brevity (and clarity in the case of `DifferentiableFunction` which is a bit ambiguous). This could be handled with a depwarn, but maybe isn't worth it. We could also consider merging these into a single type (see some options below)

**Higher Priority:**
- `DifferentiableFunction` and `TwiceDifferentiableFunction` should know about each other somehow. Some ideas:

``` julia
abstract AbstractDifferientiable
type OnceDifferentiable <: AbstractDifferentable ...  end
type TwiceDifferentiable <: AbstractDifferentiable ... end
```

``` julia
typealias Differentiable Union{OnceDifferentiable, TwiceDifferentiable}
```

``` julia
# Merge OnceDifferentiable and TwiceDifferentiable ?
# This could be a problem for dispatching on one vs the other. But there are tricks to fix it.
type Differentiable
    f::Function
    g!::Function
    fg!::Function
    h!::Function
end
# Constructor for a once differentiable function 
Differentiable(f, g!, fg!) = Differentiable(f, g!, fg!, (x,s)->error("Not implemented") )
```
- Specifying a differentiable function with autodiff options. The autodiff packages in Julia are coming along and I'd like the ability to do something along these lines:

``` julia
f(x) = ... # my objective function
DifferentiableFunction(f, method=:autograd) # uses AutoGrad.jl
DifferentiableFunction(f, method=:forward) # uses ForwardDiff.jl
DifferentiableFunction(f, method=:central_diff) # uses finite central differencing
```

I'm not attached to using symbols - we could also dispatch on types if that is preferred.

anriseth:
[...]

PR #337, opened 2017-01-08
[RFC/WIP] Rework *DifferentiableFunction

pkofod:
The main point is to create dispatch based objective evaluations, and use them in a smarter way. For example, `f_calls+=1` now follows a `value(df, x)` call. This actually caught a few places where we forgot to increment these counters even if a call was made. It's nowhere near done, but I'm just putting it up here for people to see. I didn't put [RFC] because I don't consider it anywhere near done, so I don't expect people to spend time reviewing it. However, I do accept comments and discussion at this point as well, if people want to dive in.
- [x] figure out how to get proper calls when doing finite differences (I *could* add a type parameter to the types, but that seems a bit excessive to count the number of calls... maybe a mutable instead of a number (1 element array)) ref #219
- [x] take advantage of the finite difference code for the hessian if the gradient is available (lowers the number of f calls)
- [x] allow for other "automatic" differentiation mechanisms through the keyword (maybe go back to symbols)
- [x] avoid too many calculations by checking input
- [x] exploit `last_x_h`
- [x] loosen gradient and hessian fields slightly such that users can specify their own gradient and hessian types (need to deal with `g_previous` etc)
- [x] Document changes
- [x] restrict differentiable constructor with no gradient to vector
- [x] go over type annotations. Some were left open as an experiment, some will have to be parametrized and so on. Remember, there's a WIP tag on there :)
- [x] find out what is wrong with "Large Polynomial" (which should have a diagonal hessian) and finite differences... -> there must be a bug in Calculus
- [x] better name than "method" for keyword
**Checking input `x` before calculating**
I now have *a* way to avoid recalculating `f` for the same `x`. I'm not sure if this is too much, or if it addresses the things relevant to LineSearches.jl. This PR is introducing quite some machinery, and I am not exactly sure if it's a bit too much... But let's see where it takes us!

fixes/closes: #329 #306 #305 #287 #241 #219 #163


src/objective_types.jl
```diff
@@ -0,0 +1,113 @@
...
+    g_x
+    f_calls
+    g_calls
+end
+Differentiable{T}(f, g!, fg!, x_seed::Array{T}) = Differentiable(f, g!, fg!, zero(T), similar(x_seed), [0], [0])
+function Differentiable{T}(f::Function, x_seed::Array{T}; g_method = FinDiff())
+    n_x = length(x_seed)
+    f_calls = [0]
+    g_calls = [0]
+    function g!(x::Array, storage::Array)
```
KristofferC: Why not something like:

```jl
function g!(x::Array, storage::Array)
    f_calls = 0
    Calculus.finite_difference!(x -> (f_calls += 1; f(x)), x, storage, g_method.method)
    f_calls[1] .+= f_calls
    return
end
```

to avoid having to predict how many times the gradient estimator evaluates the function. For automatic differentiation this is not always so obvious.

pkofod: I will certainly do this! Thanks for that comment :)

pkofod:
In this PR you have to give the `*Differentiable` constructor an `x` seed, as also mentioned in elsewhere in an issue somewhere. I can personally live with that, and my advice to users would simply be to use `optimize(f, g!, ...)` and don't bother constructing the `*Differentiable` types yourself unless you want to have a special `g` or `H` storage.

anriseth:
Should this be part of Optim, or some separate package?
This could be useful for LineSearches (and I guess NLsolve?) as well.

pkofod:
I think I'll put it in here first, but if it fits with nlsolve and linesearches we can take it out and have it as a dependency.

anriseth:
I think LineSearches needs this, or will have to replicate the `value` and `gradient` functions in order to work with Optim.

pkofod:
I'm not removing or renaming the old fields, so I think it still works? Certainly the line searches still work in the tests here

anriseth:
Okay, as long as we can still communicate number of f- and g-calls back to Optim, it should be fine.

KristofferC:
I think Optim.jl will keep track of that itself.

pkofod:
Currently, line searches spits out the additional calls to f and g and then we increment the counter in optim. I would love for lineaearches to use the value and value_grad methods. So I am for the idea put forth, I just want this pr to be a bit more complete before doing it. It does require nlsolve to agree though. If not, we can just keep it as it is

anriseth:
Ah, I see `do_linesearch` still keeps track of function and gradient evaluations.

Will this PR make it easier for us to prevent multiple evaluations of functions or gradients at the same points? (As discussed in https://github.com/anriseth/LineSearches.jl/issues/10 and #288)

pkofod:
If course, there's the simpler way where we just use the anon function trick as in the finite differences constructor in all incoming functions, and then lineaearches doesn't even have to worry about calls at all. Then we're back to `d.f(x)` and family again

pkofod:
Yes, I will get to that soon. It will check internally if the last x is input again.

test/nelder_mead.jl
```diff
@@ -2,7 +2,7 @@
 	# Test Optim.nelder_mead for all functions except Large Polynomials in Optim.UnconstrainedProblems.examples
 	for (name, prob) in Optim.UnconstrainedProblems.examples
 		f_prob = prob.f
-		res = Optim.optimize(f_prob, prob.initial_x, NelderMead(), Optim.Options(iterations = 10000))
+		@show res = Optim.optimize(f_prob, prob.initial_x, NelderMead(), Optim.Options(iterations = 10000))
```
anriseth: Did you intend to commit the `@show`?

src/objective_types.jl
```diff
@@ -0,0 +1,213 @@
...
+    last_x
+    f_calls
+    g_calls
+end
+function Differentiable{T}(f, g!, fg!, x_seed::Array{T})
+    g_x = similar(x_seed)
+    g!(x_seed, g_x)
+    Differentiable(f, g!, fg!, f(x_seed), g_x, copy(x_seed), [1], [1])
+end
+function Differentiable{T}(f::Function, x_seed::Array{T}; method = :finitediff)
```
Evizero: Is it important that `typeof(f) <: Function` ? Would the code work if the type of `f` simply implements `(f::MyType)(x) = ...`? if so we could consider not be overly restrictive here

test/nelder_mead.jl
pkofod: oops, thanks..

src/objective_types.jl
pkofod: Not at all, and it will be removed. I know @ChrisRackauckas would appreciate them gone as well :)

docs/src/algo/precondition.md
```diff
@@ -39,7 +39,7 @@ using ForwardDiff
 plap(U; n = length(U)) = (n-1)*sum((0.1 + diff(U).^2).^2 ) - sum(U) / (n-1)
 plap1 = ForwardDiff.gradient(plap)
 precond(n) = spdiagm((-ones(n-1), 2*ones(n), -ones(n-1)), (-1,0,1), n, n)*(n+1)
-df = DifferentiableFunction(x -> plap([0; X; 0]),
+df = Differentiable(x -> plap([0; X; 0]),
```
anriseth: Does this also need the `x_seed` now?

docs/src/user/tipsandtricks.md
```diff
@@ -81,7 +81,7 @@ using Optim
 initial_x = ...
 buffer = Array(...) # Preallocate an appropriate buffer
 last_x = similar(initial_x)
-df = TwiceDifferentiableFunction(x -> f(x, buffer, initial_x),
+df = TwiceDifferentiable(x -> f(x, buffer, initial_x),
```
anriseth: `x_seed`?

docs/src/algo/precondition.md
pkofod: Yes, that's why the `docs` box is not ticked yet :) but thanks for looking through the changes. This was just a search and replace :)

pkofod:
is this "check `last_x`" along the lines of what you had in mind @anriseth ?

docs/src/user/tipsandtricks.md
pkofod: note to self: maybe we should start adding some actual doctests

anriseth:
> is this "check last_x" along the lines of what you had in mind @anriseth ?

Thank you, it looks like a good way to deal with the double evaluation issues as far as I'm concerned. Are there situations where people would want to define such objective objects without an initial evaluation at `x_seed`? (E.g. in NLsolve or JuliaML, if we put use this in a OptimTests.jl, @KristofferC @Evizero )

pkofod:
well it depends on the general interface. Generally, I'm thinking that people in optim provide f g and h for example, and then this object is used internally. Then there is automatically a seed: the first x used for evaluation.

src/objective_types.jl
```diff
@@ -0,0 +1,221 @@
...
+        g!(x, storage)
+        return f(x)
+    end
+    g_x = similar(x_seed)
+    g!(x_seed, g_x)
+    return TwiceDifferentiable(f, g!, fg!, h!, f(x_seed),
+                                       g_x, Array{T}(n_x, n_x), copy(x_seed), [1], [1], [0])
+end
+
+function value(obj, x)
```
anriseth: Are there situations where one would want to call f(x) without updating last_x?

pkofod: > If not, maybe we would have to store last_x_f, last_x_g and last_x_h for the evaluation of f,g and h respectively.

I certainly need to handle this!

src/bfgs.jl
```diff
@@ -1,10 +1,10 @@
 # Translation from our variables to Nocedal and Wright's
 # JMW's dx <=> NW's s
 # JMW's dg <=> NW' y
-immutable BFGS{L<:Function, H<:Function} <: Optimizer
-    linesearch!::L
-    initial_invH::H
-    resetalpha::Bool
+immutable BFGS <: Optimizer
```
ScottPJones: Just curious, why was this changed from being parameterized, to leaving them all as type Any?
(I usually recommend to always use concrete types for type/immutable fields, parameterized if necessary)

ChrisRackauckas: Probably because they don't want to recompile for each new function/ But the linesearch! function, is that dependent on the user's function? If not and this is usually the same function (some default), then it should be more strictly typed. Anyways, resetalpha should still be typed to a Bool.

src/objective_types.jl
```diff
@@ -0,0 +1,221 @@
...
+end
+
+type TwiceDifferentiable
+    f
+    g!
+    fg!
+    h!
+    f_x
+    g_x
+    h_storage
```
ScottPJones: You have this parameterized in the constructor, maybe it should also be parameterized here, for better performance?

src/bfgs.jl
johnmyleswhite: I'd like to stop this comment thread now because @ScottPJones should probably be banned from the JuliaOpt org's repos in the same way that he is currently banned from the JuliaLang org's repos. For me (as the original creator of this specific project), it is very important to respect the Julia community stewards' decision to ban Scott.

pkofod:
> Just let the function ones be free so that all callable types work?
``` 
type Type{T}
    f::T
end
```
like that?

pkofod:
This now fixes #198 . I've added tests for matrices and simple tensors (rgb pictures for example?). Zeroth and first order methods only. Hopefully it will never regress again!

pkofod:
I think that maybe one commit it going to be a bit too condensed for all these changes, so I'm thinking that maybe I should rewrite history a bit before merging this. I think most of the functionality is there now, but I still need to document and add tests. When that is done, I'll replace [WIP] with [RFC].

pkofod:
> What is the meaning of :analytic here? Is it not that you would also pass prob.g!'? tooptimize`?

yes, that was a mistake. Those were not tested before for some reason (and apparently not even in this commit...oops)

pkofod:
@anriseth I'm going to rewrite history and get a clean sequence of commits. If you want to move to the `NLSolverBase.jl` approach in `LineSearches.jl` we can do it now? Then:
* I will clean up
* I will register the base package
* Optim will depend on the base package after this PR
* I will tag with a LineSearches max version in require
* We can change LineSearches to use the new methods
* Tag LineSearches
* Update Optim, lose the version cap on LineSearches
* Tag Optim
-> new system in place (NLsolve will follow along in sync of course)

anriseth:
Sounds good to me @pkofod
Would you say the introduction of an NLbase dependency warrants a LineSearches v0.2?

pkofod:
Yes, changes to REQUIRE needs at least a `:minor` tag

pkofod:
So, I'm doing PRs with subsets of the changes here, in case people hadn't noticed. The point of merging these separately is to tag a v0.7.5 that has so many of the changes from here as possible, and then only do a final PR that breaks stuff (and can't really be deprecated) in a v0.8.0 where we bump versions in other packages and announce it wide and far: discourse, twitter, readme, gitter, etc... Heck, I say we even change the order such that the mutated gradient and hessian is the first argument as well - let's create chaos around the JuliaVerse!

pkofod:
So, let's get down to business. I am considering if we should still allow
```julia
TwiceDifferentiable(f, g!, h!)
```
or if we should only have the new
```julia
TwiceDifferentiable(f, g!, h!, initial_x)
```
If we want to allow the first syntax still, then we need to allow for `TwiceDifferentiable` to be uninitialized, because we now have all sorts of stuff stored in the `*Differrentiable` types.

My personal opinion is that I don't really see any need to keep the old syntax. Users gain  <s>little</s>nothing from writing
```
df = Differentiable(f, g!)
optimize(df, initial_x, Method())
```
over
```
optimize(f, g!, initial_x, Method())
```
*unless* they want to specify `fg!`. One way to get around the dispatch problems surrounding `h!` vs `fg!` being passed would be to make a new syntax
```
optimize((f, g!, fg!), initial_x, Method())
```
syntax. That is, if you want to pass f, `g!` and it's related `fg!` you will pass them as a tuple. This would give the following combinations:
```
optimize(f, initial_x, Method(), Optim.Options())
optimize(f, g!, initial_x, Method(), Optim.Options())
optimize(f, g!, h!, initial_x, Method(), Optim.Options())
optimize((f, g!, fg!), initial_x, Method(), Optim.Options())
optimize((f, g!, fg!), h!, initial_x, Method(), Optim.Options())
```
It could also be something like the `f_and_g!_only` functions in NLsolve. Or we could just group `g!` and `fg!`
```
optimize(f, initial_x, Method(), Optim.Options())
optimize(f, g!, initial_x, Method(), Optim.Options())
optimize(f, g!, h!, initial_x, Method(), Optim.Options())
optimize(f, (g!, fg!), initial_x, Method(), Optim.Options())
optimize(f, (g!, fg!), h!, initial_x, Method(), Optim.Options())
```
I actually prefer this one, as the tuple is very clear visually.

Anyway. I think it's nice that this PR introduces an actual use of the *Differentiable types, but I think the value is more internally oriented than to the users. That is why I suggest we actually stop mentioning the types *at all* in the user facing documentation. It gives no value to them, so it's really just an implementation/booking/calculation avoiding mechanism used internally.

Would love people's opinion here! (I believe the relevant people are already present, so I'm not going to ping.)

KristofferC:
`TwiceDifferentiable(f, g!, h!, initial_x)` looks weird to me. It seems odd to pass the initial value to the function object. To me, it seems that should just go into the solver functions like before.

pkofod:
> TwiceDifferentiable(f, g!, h!, initial_x) looks weird to me. It seems odd to pass the initial value to the function object. To me, it seems that should just go into the solver functions like before.

Sure, it especially looks weird if you do
```
optimize(TwiceDifferentiable(f, g!, h!, initial_x), initial_x, Newton())
```
However, for `TwiceDifferentiable` to allocate gradient and Hessian storage, a seed is needed. Maybe that's a design mistake then. The `*Differentiable` types could be mixed with the storage in a wrapper type, but that also seems like it's over the top to me.

I guess my question is: what value does a user get from writing
```
df = TwiceDifferentiable(f, g!, h!)
optimize(df, initial_x, Method())
```
or even
```
optimize(TwiceDifferentiable(f, g!, h!), initial_x, Method())
``` 
over 
```
optimize(f, g!, h!, initial_x, Method())
```
? I mean, why would a user prefer it? To collect the functions for use multiple places? Then they might as well just do 
```
df = (f, g!, h!)
optimize(df..., initial_x, Method())
```
no need to invent a type for that purpose.

KristofferC:
Ah, so they only to use as a buffer? Couldn't it just default upon creating `df` as empty arrays in the gradient and jacobian fields, and then at the first call to `gradient(df)` they swap themselves out for the correct size arrays?

pkofod:
> Ah, so they only to use as a buffer? Couldn't it just default upon creating df as empty arrays in the gradient and jacobian fields, and then at the first call to gradient(df) they swap themselves out for the correct size arrays?

Yes, but what if you want to use `Float32` then? We cannot initialize with a `Vector{Float64}` then, if we want the type to be paramerized on the eltype. We also don't know if `initial_x <: Array{T, 3}` for example.

anriseth:
I think a setup where *Function is just an internal is sufficient. (At least for LineSearches and Optim).
The user only needs to supply callables that compute f,g,h,fg, and I guess a Tuple works well for that.

KristofferC:
> Yes, but what if you want to use Float32 then?

Good point. `ForwardDiff` does similar to this in that you pass in a vector so it can allocate and know the eltype so it has some prior art.

However, I do agree that a user should not have to create `df` themselves and just pass in `f, g!, (fg!)` in which case `df` is just created internally and then this whole conversaytion is moot.

anriseth:
>  better name than "method" for keyword

I like "solver" (if you're talking about the `method` keyword in `optimize`)

pkofod:
> I like "solver" (if you're talking about the method keyword in optimize)

This was actually for for autodiff-keyword :) but I'm open to the solver keyword as well (which we can simply manually deprecate).

pkofod:
Tests are running significantly longer, but I guess it might be because of compilation time. Will investigate.

pkofod:
Sort of need people's opinion on something. The strict typing of the `NonDifferentiable`, `OnceDifferentiable`, `TwiceDifferentiable` (for example, https://github.com/JuliaNLSolvers/Optim.jl/pull/337/files#diff-ab516b956239be1e33f110faaa3a8b4cR11 ) is a possible cause for concern.

If you run a test set repeatedly on master and on this branch, then master is much faster the first time, but this branch will be slightly faster and use less memory the second time and after that. However, if you change all the `optimize` calls to be called on a `x->prob.f(x)` anonymous function, then master is almost exactly the same performance, but this branch will be quite a lot slower.

So this all comes down to the following: how do people use Optim? Should we provide a non-strict version by default and then an opt-in `OnceDifferentiableStrict` for the cases where strict typing is preferred? Or just keep the non-strict version (this was the case before). Personally, I'm not really sure, so I guess a "non-breaking" change would be better, that is revert to `::Function` until we know more.

Comments are welcome.

anriseth:
Let me try to understand the implications here:
I often run thousands of calls to optimize, with functions of the same form but different parameters, e.g.
```jl
#avals some vector
f(x,a) = (x[1]-a)^2
for k = 1:1000
   obj(x) = f(x,avals[k])
   optimize(obj, [0.0], GradientDescent())
end
```
Would this be slowed down with the new PR? And if so, how would I fix that as a user calling `optimize`?

ChrisRackauckas:
>Would this be slowed down with the new PR? And if so, how would I fix that as a user calling optimize?

No. This case will be slow down by the PR, because it will compile a special version of `optimize` for each new `obj` function. You can avoid this by making your function a call overloaded type and just changing the parameters in it though.

anriseth:
> You can avoid this by making your function a call overloaded type and just changing the parameters in it though.

Exciting! Do you mean something along the lines of:
```jl
type myobj{T}
  a::T
end
(ob::myobj{T}){T}(x) = (x[1]-ob.a)^2
f = myobj(avals[1])
df = OnceDifferentiable(f, [0.0])
for k = 1:1000
   f.a = avals[k]
   optimize(df, [0.0], GradientDescent())
end
```
I initially tried to do `optimize(f,[0.0],GradientDescent())`, but that would require looser typing of `optimize`.

ChrisRackauckas:
>I initially tried to do optimize(f,[0.0],GradientDescent()), but that would require looser typing of optimize.

Yes, I agree the typing of `optimize` should be looser to more easily accomodate this, though with the current typing you can do this with one change:

```julia
type myobj{T} <: Function
  a::T
end
```

pkofod:
I could remove restrictions there. Did it to avoid optimize being compiled over and over.

ChrisRackauckas:
The restrictions don't do anything for compilation or performance.

KristofferC:
Specializing on the function type is unnecessary since for everything except completely trivial functions it will not inline anyway.

anriseth:
Please correct me if I'm wrong here, I am still trying to understand the pros/cons of restricting to `::Function`.

I thought reverting "Back to Function" with ac27e76 decreases the generality of `Optim` without improving performance?
If I understand correctly, restricting `f`,`g`, etc. in `optimize` and `*Function` to be of type `::Function` does not help Julia create specialised lower-level code, nor does it reduce the amount of compilation that must be done.

KristofferC:
I consider the options to be:

```jl
type OnceDifferentiable{F, ...}
    f::F
    ...
```

or

```jl
type OnceDifferentiable{, ...}
    f
    ...
```

Neither of these restrict the arguments to `Function`. The difference is that for the first one, all methods that gets the `OnceDifferentiable` type passed into are recompiled when the type `F` changes (for example when using different functions). This can be useful if you want `f` to be inlined or dynamic dispatch.  Dynamic dispatch is likely insignificant here compared to the amount of work that is being done per iteration so I would suggest going with the second version.

pkofod:
> I thought reverting "Back to Function" with ac27e76 decreases the generality of Optim without improving performance?

I wanted to see the changes in CI times. I've removed them completely now. Should probably do this on line searches as well.

pkofod:
It's hitting something weird on v0.5, that makes the Newton tests run much longer. I think the next major tag will be either v0.6 only, or the last v0.5 tag, so I'm not going to be fixing it. Obviously some bug has been fixed somewhere, as the tests go from ~190s to 20-30s from v0.5 to v0.6.

anriseth:
Shall we move from `g!(x, out)` to `g!(out, x)` as well, while we're already at breaking changes?

pkofod:
> Shall we move from g!(x, out) to g!(out, x) as well, while we're already at breaking changes?

Sure, but maybe in a separate PR. Then I can merge this this week, create a new pr for the gradient position change (to avoid confusion), merge and then tag. Sounds good?

anriseth:
> Sure, but maybe in a separate PR. Then I can merge this this week, create a new pr for the gradient position change (to avoid confusion), merge and then tag. Sounds good?

Makes sense. 
Are you going to separate out the *Function changes in this PR to `NLSolversBase.jl`, or merge directly to `Optim`?

pkofod:
> Makes sense.
> Are you going to separate out the *Function changes in this PR to NLSolversBase.jl, or merge directly to Optim?

I was just going to merge this actually, and then copy+paste the code to NLSolversBase.jl, just to have the change in the git history (so it is easier to follow what happened prior to moving code to another package).

pkofod:
I'm pushing some bug fixes tonight, and adding `x_calls_limit` for f, g, and h. Now is the time to speak up if you don't like these changes, as I'll merge soon.

src/objective_types.jl
```diff
@@ -0,0 +1,275 @@
...
+        g!(x, storage)
+        return f(x)
+    end
+    return OnceDifferentiable(f, g!, fg!, x_seed)
+end
+function OnceDifferentiable{T}(f, x_seed::Vector{T}; method = :finitediff)
+    n_x = length(x_seed)
+    f_calls = [1]
+    g_calls = [1]
+    if method == :finitediff
```
anriseth: If we add these methods here, maybe we can call `*Differentiable(f,x_init;method)` directly from the `optimize` functions instead of having separate code for generating `g!`, `fg!` (and `h!`) there? I'm referring to e.g. https://github.com/JuliaNLSolvers/Optim.jl/blob/9d90b5aa3d2047555c965fd5e3b22efd4bd09973/src/optimize.jl#L98

pkofod: good catch, I forgot that was also done there.

test/gradient_descent.jl
```diff
@@ -2,18 +2,21 @@
     for use_autodiff in (false, true)
```
anriseth: Is `use_autodiff` used anywhere now?

pkofod:
Right, so just to update everyone, this is approaching its end. With this PR I am no longer targeting v0.4 (yet to be changed) and I won't carry the deprecations forwards either. This means that this PR is heavily breaking, so tagging will have to be done with care (limits in metadata). I removed all deprecations as well. 

V0.4 was so long ago, v0.6 is coming soon, and shortly after we have v1.0. Post v1.0 I will be very happy to be very careful in backwards compatibility, but as far as I'm concerned, we're approaching a sprint for v1.0 of JuliaLang, and some things will have to be done a little faster. The main reason is, that we don't have 1-2 people sitting here full time, making sure that everything works across v0.2-v0.6, mac, linux, windows, amiga, and raspberry pi.

PR #338, opened 2017-01-09
Change all tests to testsets and suppress some information by default.

pkofod:
Just something I've been wanting to do for some time. I think we should also move all the identical loops over problems into the `optimize` test set, and let individual files for each solver be there only if we have unit tests for related methods or very specific problems. NewtonTrustRegion is a perfect example with tests for the subproblem solver.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,0 +1,1 @@
+BaseTestNext
--- a/test/accelerated_gradient_descent.jl
+++ b/test/accelerated_gradient_descent.jl
@@ -1,5 +1,5 @@
 # TODO expand tests here
-let
+@testset "Accelerated Gradient Descent" begin 
     f(x) = x[1]^4
     function g!(x, storage)
         storage[1] = 4 * x[1]^3
--- a/test/api.jl
+++ b/test/api.jl
@@ -1,5 +1,5 @@
 # Test multivariate optimization
-let
+@testset "API" begin
     rosenbrock = Optim.UnconstrainedProblems.examples["Rosenbrock"]
     f = rosenbrock.f
     g! = rosenbrock.g!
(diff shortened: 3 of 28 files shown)


Evizero:
The 0.4 tests fails could likely be fixed with the following in `runtests.jl` instead of https://github.com/JuliaOpt/Optim.jl/blob/master/test/runtests.jl#L6 

```julia
if VERSION >= v"0.5-"
    using Base.Test
else
    using BaseTestNext
    const Test = BaseTestNext
end
```

and a `BaseTestNext` in `REQUIRE`

Evizero:
unsure about the win32 fail.

pkofod:
> unsure about the win32 fail.

That is some weird `Fminbox` failure that I have yet to figure out. Somewhere the objective (including the barrier) returns `NaN`, and it is while calculating the barrier contribution I believe, as it never actually reaches the `fg!` call. I don't have a 32bit install, so it's kind of slow to debug (through a PR and watching the output of AppVeyor).

pkofod:
> The 0.4 tests fails could likely be fixed with the following in runtests.jl instead of https://github.com/JuliaOpt/Optim.jl/blob/master/test/runtests.jl#L6
> 
> if VERSION >= v"0.5-"
>     using Base.Test
> else
>     using BaseTestNext
>     const Test = BaseTestNext
> end
> 
> and a BaseTestNext in test/REQUIRE

I'll try it out tomorrow. Thanks. It's actually impressive how little we have to do to support v0.4 through v0.6. Julia sure is stabilizing! (at least the parts we use).

pkofod:
Great. v0.4 now passes. It's running again due to a merge conflict I had to handle, and I also added some test sets I had missed.

tkelman:
You should use a more specific VERSION cutoff than that. Base.test did not support testsets for all 0.5 prerelease versions.

pkofod:
@tkelman okay. Is there an easy way to get the build number, or should I just build the specific commit and see?

tkelman:
check other packages, they mostly have it right. or run contrib/commit-name.sh

PR #340, opened 2017-01-12
Add allow_f_increases to Options (default: false) and fminbox (defaul…

pkofod:
…t: true).

It will be interesting to see if tests pass (forget 32bit on windows still) :) I simply did the boolean version for now.


pkofod:
Hm, I knew that iterations test would come back to haunt me once, but it's a bit annoying that we get numerically different results (even if ever so slightly) between platforms.

anriseth:
Great, thank you.

Differences between platforms are definitely annoying, I've had problems with that before too. (On one system, my patch made huge performance improvements, on another it made it worse).

pkofod:
We'll see if it's the only failure on mac... For now I just commented it out

pkofod:
@johnmyleswhite is this :+1: ?

src/fminbox.jl
```diff
@@ -111,6 +111,7 @@ function optimize{T<:AbstractFloat}(
         x_tol::T = eps(T),
         f_tol::T = sqrt(eps(T)),
         g_tol::T = sqrt(eps(T)),
+        allow_f_increases = true,
```
Evizero: should we document the type? is this always a `Bool`?

src/types.jl
```diff
@@ -2,6 +2,7 @@ abstract Optimizer
 immutable Options{TCallback <: Union{Void, Function}}
     x_tol::Float64
     f_tol::Float64
+    allow_f_increases
```
Evizero: Here we probably should specify a concrete type, since all other members do

src/types.jl
```diff
@@ -148,6 +151,7 @@ function Base.show(io::IO, r::MultivariateOptimizationResults)
         @printf io "   * |x - x'| < %.1e: %s\n" x_tol(r) x_converged(r)
         @printf io "   * |f(x) - f(x')| / |f(x)| < %.1e: %s\n" f_tol(r) f_converged(r)
         @printf io "   * |g(x)| < %.1e: %s\n" g_tol(r) g_converged(r)
+        @printf io "   * f(x) > f(x'): %s\n" r.f_increased
```
Evizero: may also worth hiding behind an API?

src/fminbox.jl
pkofod: Right now it can only be boolean, yes.

src/types.jl
pkofod: Yes, I can type it out as boolean... I think I left it untyped, as I wasn't sure if I wanted to add a limit or an on/off switch. Right now I think the on/off approach is sufficient.

src/types.jl
pkofod: Sure... speaking of API and looking at that code, I need to go over the code that uses `g_converged` and family as variable names...

Evizero:
> I need to go over the code that uses g_converged and family as variable names

Here some locations that seem relevant (i filtered a few out by hand)

```
src$ find . -name "*.jl" | xargs awk '/[^."]._(converged|calls)[^(,:]/ { print FILENAME ":" FNR, $0 }' 
|---------------------------------------+-----------------------------------------------------------------------------------------------|
|  ./levenberg_marquardt.jl:46          |     x_converged = false                                                                       |
|  ./levenberg_marquardt.jl:47          |     g_converged = false                                                                       |
|  ./levenberg_marquardt.jl:52          |     f_calls = 0                                                                               |
|  ./levenberg_marquardt.jl:53          |     g_calls = 0                                                                               |
|  ./levenberg_marquardt.jl:56          |     f_calls += 1                                                                              |
|  ./levenberg_marquardt.jl:78          |             g_calls += 1                                                                      |
|  ./levenberg_marquardt.jl:132         |         f_calls += 1                                                                          |
|  ./levenberg_marquardt.jl:166         |             g_converged = true                                                                |
|  ./levenberg_marquardt.jl:168         |             x_converged = true                                                                |
|  ./levenberg_marquardt.jl:170         |         converged = g_converged | x_converged                                                 |
|  ./optimize.jl:182                    |     x_converged, f_converged = false, false                                                   |
|  ./optimize.jl:183                    |     g_converged = if typeof(method) <: NelderMead                                             |
|  ./optimize.jl:186                    |         g_converged = false                                                                   |
|  ./newton_trust_region.jl:243         |     f_calls, g_calls = 1, 1                                                                   |
|  ./newton_trust_region.jl:246         |     h_calls = 1                                                                               |
|  ./newton.jl:34                       |     f_calls, g_calls = 1, 1                                                                   |
|  ./newton.jl:37                       |     h_calls = 1                                                                               |
|  ./golden_section.jl:48               |     f_calls = 1 # Number of calls to f                                                        |
|  ./golden_section.jl:74               |             f_calls += 1                                                                      |
|  ./golden_section.jl:85               |             f_calls += 1                                                                      |
|  ./golden_section.jl:109              |                                          f_calls)                                             |
|  ./particle_swarm.jl:86               |     f_calls = 0                                                                               |
|  ./particle_swarm.jl:89               |     f_calls += 1                                                                              |
|  ./types.jl:176                       |     a.iteration_converged = iteration_limit_reached(b)                                        |
|  ./brent.jl:49                        |     f_calls = 1 # Number of calls to f                                                        |
|  ./brent.jl:123                       |         f_calls += 1                                                                          |
|  ./brent.jl:168                       |                                          f_calls)                                             |
|  ./simulated_annealing.jl:41          |     f_calls = 1                                                                               |
|---------------------------------------+-----------------------------------------------------------------------------------------------|
```

src/types.jl
```diff
@@ -69,6 +71,7 @@ type MultivariateOptimizationResults{T,N,M} <: OptimizationResults
     f_tol::Float64
     g_converged::Bool
     g_tol::Float64
+    f_increased
```
Evizero: This is also a known `Bool`,  right? I guess since this type is mutable it makes less of a difference though

pkofod: huh, I thought I added it here as well... good to have you watching over me ! I'll add it

test/types.jl
```diff
@@ -35,8 +35,8 @@ import Compat.String
             @test startswith(lines[8], "   * |x - x'| < ")
             @test startswith(lines[9], "   * |f(x) - f(x')| / |f(x)| < ")
             @test startswith(lines[10], "   * |g(x)| < ")
-            @test startswith(lines[11], "   * Reached Maximum Number of Iterations: ")
-            @test startswith(lines[12], " * Objective Function Calls: ")
+            @test startswith(lines[12], "   * Reached Maximum Number of Iterations: ")
```
Evizero: did you leave `lines[11]` out on purpose?

johnmyleswhite:
Are we returning `x` or `x_previous` after finding that `f_increased`? Can we make sure we always return the smaller of the two?

pkofod:
> Are we returning x or x_previous after finding that f_increased? Can we make sure we always return the smaller of the two?

Good point, I'll make sure we return the best of the two.

pkofod:
I also removed the odd `Float64` restriction on the minimum.

pkofod:
So, there's kind of a "bad code style" going on here. It relates to the existing code, but let us take the new code as an example. I'm using `f_increased` as a variable name inside `optimize`, but we also have a function with that name. Should we rename the variable names? I mean... It works because we never use the API-functions in the code itself, but it still seems like something a linter would throw big fat red warnings about.

Issue #34, opened 2013-08-27
Maximum Step Size


SamChill:
I'm not sure where the best place to discuss Optim is so I'll post this here.

In my field, computational chemistry, it is common to enforce a maximum step size constraint at each optimization step, when working with atomic systems. This is very important when the gradient is large because a large step can be taken without this constraint. A large step will often lead away from the current basin of attraction to a different one. Finding the minimum associated with the current geometry is often why the optimization is being performed. Also, if the initial force is large enough, all of the atoms might be placed so far apart that the energy (function value) goes to zero. This certainly is a minimum, but it is a trivial one.

I would like to know if anyone has any thoughts on adding a maximum step size to Optim. The way that it is typically implemented would be to calculate the length of the proposed step size and if it is larger than the maximum, set it to the maximum.

One reason that a maximum step size makes sense in atomic systems is that the second derivates very rapidly with changes in bond lengths. This means that curvature information learned at the current position should only be used locally. For example, a typical maximum step size might be between 0.1 and 0.3 Angstroms (several percent of an equilibrium bond length).

I am currently writing a paper that benchmarks various optimization algorithms and implementations in atomic systems and I would like to include Optim, however, as it currently stands, Optim immediently pushes all of the atoms extremely far apart in the first step, due to the large initial gradient.

pkofod:
As mentioned, the functionality is already here, so I am closing this.

PR #341, opened 2017-01-17
bug fix for newton_trust_region

jeff-regier:
This PR fixes a bug in `newton_trust_region.jl`. In the loop around line 128, we want to increment lambda by 1% of the **difference** between the largest and smallest eigenvalues. Without this PR, we're effectively incrementing lambda by 1% of the **sum** of the largest and smallest eigenvalues.   That can lead to a `Base.LinAlg.PosDefException(5)` later in the code if the algebraically smallest eigenvalue is larger in magnitude than the algebraically largest eigenvalue, as reported by @rgiordan in jeff-regier/Celeste.jl#529

### DIFF
--- a/src/newton_trust_region.jl
+++ b/src/newton_trust_region.jl
@@ -124,8 +124,9 @@ function solve_tr_subproblem!{T}(gr::Vector{T},
             p_lambda2 = p_sq_norm(lambda, min_H_ev_multiplicity + 1)
             if p_lambda2 > delta_sq
                 # Then we can simply solve using root finding.
-                # Set a starting point between the minimum and largest eigenvalues.
-                lambda = lambda_lb + 0.01 * (max_H_ev - lambda_lb)
+                # Set a starting point greater than the minimum based on the
+                # range between the largest and smallest eigenvalues.
+                lambda = lambda_lb + 0.01 * (max_H_ev - min_H_ev)
             else
                 hard_case = true
                 reached_solution = true
--- a/test/newton_trust_region.jl
+++ b/test/newton_trust_region.jl
@@ -193,4 +193,11 @@ end
     	end
     end
 end
+
+
+@testset "PR #341" begin
+    # verify that no PosDef exception is thrown
+    Optim.solve_tr_subproblem!([0, 1.], [-1000 0; 0. -999], 1e-2, ones(2))
+end
+
 end


rgiordan:
I think a more appropriate fix for this problem would be to change ```lambda_lb``` to be no larger in magnitude than ```max_H_ev```, something like a line 114:

```lambda_lb = -min(abs(min_H_ev), abs(max_H_ev)) + max(1e-8, 1e-8 * (max_H_ev - min_H_ev))```

```lambda_lb``` is defined as the most negative allowable value for ```lambda```  Line 128 sets the starting value of ```lambda``` to be partway through its allowable range, which is from ```lambda_lb``` to ```max_H_ev```, not from ```min_H_ev``` to ```max_H_ev```.

**Comment by rgiordan** [comment]
Oh, never mind, I misunderstood and retract my comment.  To save other people from my confusion, the comment in the code should be updated from

```
# Set a starting point between the minimum and largest eigenvalues.
```

to something like

```
# Set a starting point greater than the minimum based on the range between the largest and smallest eigenvalues.
```

**Comment by jeff-regier** [comment]
Oh good, I was about to stay that if `lambda = lambda_lb < -min_H_ev`, then `H_ridge` won't be positive definite, even if `abs(lambda_lb) == max_H_ev`.You'll get a posdef exception on line 161. So a lower bound on lambda should be at least `-min_H_ev` (plus a tiny positive integer).

**Comment by pkofod** [comment]
Thanks.. I'll have a look later!

**Comment by jeff-regier** [comment]
Great! Btw, this PR also fixes a lot of warnings on nightly---so many warnings that unit the tests fail on travis because the log gets too big. The old array creation syntax must have been deprecated in the last 3 days, because I see the Optim.jl unit tests all passed last time it was run.

The appveyor error looks unrelated to this PR.

**Comment by pkofod** [comment]
> Great! Btw, this PR also fixes a lot of warnings on nightly---so many warnings that unit the tests fail on travis because the log gets too big. The old array creation syntax must have been deprecated in the last 3 days, because I see the Optim.jl unit tests all passed last time it was run.
> 
> The appveyor error looks unrelated to this PR.

AppVeyor is some weird bug in fminbox on 32bit windows.. I havn't had time to fix it, because I don't have such an installation, and I havn't had time to VM it.

The array deprecation is very recent. I suspected this would happen, as I saw that PR the other day. Didn't know it was merged already. It'll be fixed. I'm not too worried about v0.6 tests passing for this to be merged.

**Comment by pkofod** [comment]
If you reset that last commit I think you won't have any conflicts (I solved this deprecation issue in another pr).

**Comment by jeff-regier** [comment]
I resolved the conflicts though editor, though reset may have been better. Is this OK as is? Would you mind tagging a new release with this bug fix?

**Comment by pkofod** [comment]
I can see that you hit the error during a benchmark. Is it possible to include a reduced examples that would his this posdef exception before this fix?

**Comment by jeff-regier** [comment]
How can I recreate the error? It looks like the tests all passed on Travis.

**Comment by pkofod** [comment]
Sorry for being cryptic. I meant I can see over at Celeste that the error was found in one of your benchmarks. I was just wondering if you could add the relevant part of that benchmark here as a test, unless it runs for six days or is 2000 lines of code .

**Comment by jeff-regier** [comment]
It'd be a pretty big matrix, probably not something you'd hard code into test file. We should be able to trigger the condition with just a 2x2 matrix, with very negative eigenvalues, except that we also have to get past the hard case check. This works:
```
julia> Optim.solve_tr_subproblem!([0, 1.], [-1000 0; 0. -999], 1e-2, ones(2))
p_sq_norm(0.0,1) = 1.002003004005006e-6
hard_case_candidate = true
p_lambda2 = 0.9999999799999846
lambda = 980.0100000099
ERROR: Base.LinAlg.PosDefException(1)
 in _chol!(::Array{Float64,2}, ::Type{UpperTriangular}) at ./linalg/cholesky.jl:55
 in chol!(::Hermitian{Float64,Array{Float64,2}}) at ./linalg/cholesky.jl:124
 in chol(::Hermitian{Float64,Array{Float64,2}}) at ./linalg/cholesky.jl:145
 in #solve_tr_subproblem!#34(::Float64, ::Int64, ::Function, ::Array{Float64,1}, ::Array{Float64,2}, ::Float64, ::Array{Float64,1}) at /home/ubuntu/.julia/v0.5/Optim/src/newton_trust_region.jl:165
 in solve_tr_subproblem!(::Array{Float64,1}, ::Array{Float64,2}, ::Float64, ::Array{Float64,1}) at /home/ubuntu/.julia/v0.5/Optim/src/newton_trust_region.jl:64
```

pkofod:
That would be great, can you put it in a `@testset "PR #341" begin end` block` in `newton_trust_region.jl`?

pkofod:
Great. Everything passes. Good to have a heavy user of one of the solvers in here! I'll tag relatively soon.

PR #344, opened 2017-01-18
Fix f convergence

rgiordan:
I tried to make a unit test, but my attempts to construct functions that reliably forced different types of convergence were getting a little clunky, so I thought I'd just submit the fix.  If you really want tests, LMK and we can talk about how best to do it.

### DIFF
--- a/src/utilities/assess_convergence.jl
+++ b/src/utilities/assess_convergence.jl
@@ -15,7 +15,7 @@ function assess_convergence(x::Array,
     # Absolute Tolerance
     # if abs(f_x - f_x_previous) < f_tol
     # Relative Tolerance
-    if abs(f_x - f_x_previous) < min(f_tol * (abs(f_x) + f_tol), eps(abs(f_x)+abs(f_x_previous)))
+    if abs(f_x - f_x_previous) < max(f_tol * (abs(f_x) + f_tol), eps(abs(f_x)+abs(f_x_previous)))
         f_converged = true
     end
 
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
I think we can merge this as it seems to be a clear bug fix, but we should be able to unit test this by passing in dummy information that doesn't generated by an actual attempt at optimization.

Can I ask why we use `abs(f_x) + f_tol` instead of just `abs(f_x)`? And why we're testing the `eps` branch? It's not obvious to me what situation checking against `eps` helps in?

rgiordan:
I'm just changing the existing code from min to max.

My $0.02, I think you use ```abs(f_x) + f_tol``` so that ```f_x = 0``` doesn't always result in convergence.  Hopefully someone else can answer for ```eps``` though.

rgiordan:
FWIW, I briefly tried writing a test that passed in dummy information but had problems with the line search failing and terminating the optimization (since it wasn't a real function).  Probably one of you will have a good idea to get around this.

anriseth:
> My $0.02, I think you use abs(f_x) + f_tol so that f_x = 0 doesn't always result in convergence. 

`abs(f_x_previous-f_x) < f_tol*abs(f_x)` would not yield convergence just because f_x is zero?

pkofod:
:+1: to @anriseth s explanation. I Will say though that its somewhat of a hack, and we should just have f_atol and f_rtol imo

anriseth:
@pkofod Sorry I just edited my answer, although the extra +f_tol is still sort of a hack to work around the missing f_atol 

@rgiordan you can try to run the tests with `LineSearches.basic!` instead of the default linesearch.

pkofod:
Based on conversations with Tom Breloff, I have some ideas for how to handle all these stopping criteria more flexibly in the future anyway, so as long as the bug is gone, and we can use the `ftol` as before, I don't think we need to worry about adding `atol` just not. But... if someone has time and wants to improve this right now, they're certainly also welcome.

pkofod:
Also - woot, this seems to fix the appveyor 32bit failure. I'm not exactly sure why, but...

PR #352, opened 2017-01-26
Refactor linesearch.

### LINKED ISSUE #320: perform_linesearch_step
Lots of duplicate code in the algorithms that call linesearch. Maybe we should create a `perform_linesearch_step` function?

Ref https://github.com/JuliaOpt/Optim.jl/pull/275#issuecomment-266857553

pkofod:
fixes #320

- [x] include clearing
- [x] move alphaguess functionality
- [x] check dphi0 for those that need it

### DIFF
--- a/src/bfgs.jl
+++ b/src/bfgs.jl
@@ -78,9 +78,6 @@ function update_state!{T}(d, state::BFGSState{T}, method::BFGS)
     push!(state.lsr, zero(T), state.f_x, dphi0)
 
     # Determine the distance of movement along the search line
-    if method.resetalpha == true
-        state.alpha = one(T)
-    end
     lssuccess = perform_linesearch(state, method, d)
 
     # Maintain a record of previous position
--- a/src/bfgs.jl
+++ b/src/bfgs.jl
@@ -59,7 +59,6 @@ end
 
 
 function update_state!{T}(d, state::BFGSState{T}, method::BFGS)
-    lssuccess = true
     # Set the search direction
     # Search direction is the negative gradient divided by the approximate Hessian
     A_mul_B!(state.s, state.invH, state.g)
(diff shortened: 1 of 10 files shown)


KristofferC:
Nice code deletion

src/bfgs.jl
```diff
@@ -79,24 +78,10 @@ function update_state!{T}(d, state::BFGSState{T}, method::BFGS)
...
-    catch ex
-        if isa(ex, LineSearches.LineSearchException)
-            lssuccess = false
-            state.f_calls, state.g_calls = state.f_calls + ex.f_update, state.g_calls + ex.g_update
-            state.alpha = ex.alpha
-            Base.warn("Linesearch failed, using alpha = $(state.alpha) and exiting optimization.")
-        else
-            rethrow(ex)
-        end
+    if method.resetalpha == true
```
pkofod: this should just be in the function

src/newton.jl
```diff
@@ -69,29 +68,12 @@ function update_state!{T}(d, state::NewtonState{T}, method::Newton)
     push!(state.lsr, zero(T), state.f_x, dphi0)
 
     # Determine the distance of movement along the search line
-    try
-        if method.resetalpha == true
```
pkofod: this was deleted, but will be in `perform_linesearch` next commit

src/utilities/perform_linesearch.jl
```diff
@@ -0,0 +1,23 @@
+perform_linesearch(state, method, d) = perform_linesearch(state, method, d, state.alpha)
+function perform_linesearch(state, method, d, alphaguess)
+    # Determine the distance of movement along the search line
+    try
+        if method.resetalpha == true
```
anriseth: What happens if 'method' doesn't have the 'resetalpha' field?

pkofod: Well, that is what you call a good question. It won't work. I'll fix it when at a pc.

pkofod:
Alright, so now a linesearch step is one line in the `update_state!` calls. Right now `alphaguess` and `checked_dphi0` are in the `perform_linesearch` file, but could be in the respective method files.

pkofod:
@anriseth when you have some time I'd be happy if you could have one last look.

anriseth:
> @anriseth when you have some time I'd be happy if you could have one last look.

This looks good, I have no particular comments.

PR #353, opened 2017-01-27
v0.6 warning fixes

anriseth:
I fixed a warning that came up with Optim on v0.6:
```jl
WARNING: Method definition optimize(Any, Array{T, N} where N, M<:Optim.Optimizer, Optim.Options{TCallback} where TCallback<:Union{Void, Function}) in module Optim at /home/riseth/.julia/v0.6/Optim/src/optimize.jl:52 overwritten at /home/riseth/.julia/v0.6/Optim/src/optimize.jl:169.
```

In addition, I removed lots of Levenberg deprecation warnings that cluttered up the test output. I guess we should just remove the Levenberg code at some point?

### DIFF
--- a/src/levenberg_marquardt.jl
+++ b/src/levenberg_marquardt.jl
@@ -54,7 +54,7 @@ function levenberg_marquardt{F<:Function, G<:Function, T}(f::F, g::G, initial_x:
 
     fcur = f(x)
     f_calls += 1
-    residual = sumabs2(fcur)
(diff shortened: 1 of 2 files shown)


pkofod:
> I guess we should just remove the Levenberg code at some point?

Sure, we will remove it when the next major tag is made (after the Differentiable-PR is merged). I thought we just removed it, but I can see my PR was merged at LsqFit november 17 ... time flies!

PR #355, opened 2017-01-29
NonDifferentiableFunction -> NonDifferentiable, DifferentiableFunctio…

pkofod:
…n -> OnceDifferentiable, TwiceDifferentiableFunction -> TwiceDifferentiable.

Taken from #337 . The point of merging these separately is to tag a v0.7.5 that has so many of the changes from #337 as possible, and then only do a final PR that breaks stuff (and can't really be deprecated) in a v0.8.0 where we bump versions in other packages and announce it wide and far: discourse, twitter, readme, gitter, etc...


pkofod:
I don't see why this shouldn't work, but I will wait until the v0.6 problems are sorted out until I merge. I don't want to debug v0.6 in another PR if this PR is to blame!

pkofod:
Actually, this might be due to some v0.6 change that we're hitting... :microscope:

PR #357, opened 2017-02-02
RFC: Support ReverseDiff

anriseth:
This adds ReverseDiff automatic differentiation support.

I am not sure whether I have set it up in the most robust way, some feedback on how I call ReverseDiff will be appreciated ( @jrevels ?) 

I have chosen to change `autodiff::Bool` to be a symbol, and allow the options `autodiff == :finite`, `:forward` or `:reverse`. Anybody opposed to this approach?

TODO:
- [ ] Deprecate `autodiff::Bool` keywords

 @jeff-regier

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -134,7 +134,7 @@ function optimize{F<:Function, T, M <: Union{FirstOrderSolver, SecondOrderSolver
         error("The autodiff value $(options.autodiff) is not supported. Use :finite, :forward or :reverse.")
     end
 
-    optimize(d, initial_x, method, options).
+    optimize(d, initial_x, method, options)
 end
 
 function optimize{M<:Union{Newton,NewtonTrustRegion}}(d::OnceDifferentiable,
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -88,13 +88,13 @@ function optimize{F<:Function, T, M <: Union{FirstOrderSolver, SecondOrderSolver
                   initial_x::Array{T},
                   method::M,
                   options::Options)
-    if !options.autodiff
+    if options.autodiff == :finite
         if M <: FirstOrderSolver
             d = OnceDifferentiable(f)
         else
-            error("No gradient or Hessian was provided. Either provide a gradient and Hessian, set autodiff = true in the Options if applicable, or choose a solver that doesn't require a Hessian.")
+            error("No gradient or Hessian was provided. Either provide a gradient and Hessian, set autodiff = :forward in the Options if applicable, or choose a solver that doesn't require a Hessian.")
         end
-    else
+    elseif options.autodiff == :forward
         gcfg = ForwardDiff.GradientConfig(initial_x)
         g! = (x, out) -> ForwardDiff.gradient!(out, f, x, gcfg)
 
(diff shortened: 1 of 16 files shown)


src/optimize.jl
```diff
@@ -111,33 +111,47 @@ function optimize{F<:Function, T, M <: Union{FirstOrderSolver, SecondOrderSolver
             h! = (x, out) -> ForwardDiff.hessian!(out, f, x, hcfg)
             d = TwiceDifferentiable(f, g!, fg!, h!)
         end
-    end
+    elseif options.autodiff == :reverse
```
anriseth: @jrevels Is this way of implementing  `g!` and `fg!` and `h!` sensible?

mlubin:
``autodiff=true`` is dead. Long live ``autodiff=true``

src/optimize.jl
anriseth: (I mean in the lines 114-132)

jrevels:
First off, thanks for this!

Getting into the nitty gritty: Calling `ReverseDiff.compile_*` in non-top-level scope is generally a bad idea.  I've been meaning to emphasize this in the documentation more, but haven't gotten around to it - sorry about that.  On Julia v0.6, this actually won't even be usable (or at least shouldn't be usable) because of ReverseDiff compilation depends on `eval` (since it essentially translates runtime values into a compiled function).

Instead, you can use the `ReverseDiff.gradient!(result, input, cfg)` style API methods (making sure, of course, to preallocate reuse `cfg` where possible). 

To provide a better explanation: You're guaranteeing certain properties about a function when you compile it using ReverseDiff. These properties are in the ReverseDiff docs, but might be a bit scattered (again, my bad). For example, one of the guarantees is that the function won't generate a tape that is too long to be compiled in a reasonable amount of time. Another is that the function doesn't have any value-dependent branches. If you're using `ReverseDiff.compile_*` in a non-top-level scope, it's probably because you don't actually know what your objective function is, and thus, you can't guarantee that it's safe to compile.

anriseth:
>Getting into the nitty gritty: Calling ReverseDiff.compile_* in non-top-level scope is generally a bad idea. I've been meaning to emphasize this in the documentation more, but haven't gotten around to it - sorry about that. On Julia v0.6, this actually won't even be usable (or at least shouldn't be usable) because of ReverseDiff compilation depends on eval (since it essentially translates runtime values into a compiled function).

Ah, yes I was wondering if `compile_` was a bit too optimistic. 

> Instead, you can use the ReverseDiff.gradient!(result, input, cfg) style API methods (making sure, of course, to preallocate reuse cfg where possible).

Is it safe to preallocate cfg for ReverseDiff, in the same way that Optim does forForwardDiff? https://github.com/JuliaOpt/Optim.jl/blob/master/src/optimize.jl#L98

anriseth:
In any case, this PR is not necessarily the way we want to go with AD, based on the most recent discussions in #329.

pkofod:
> In any case, this PR is not necessarily the way we want to go with AD, based on the most recent discussions in #329.

Well, we might still want to provide "wrappers" as @jrevels mentioned in #329 , and this code is very relevant there! So your work will definitely not have been for nothing.

pkofod:
thanks, closed by #384

PR #363, opened 2017-02-14
Fix depwarns on 0.6

yuyichao:


anriseth:
Seems like there are some issues with this fix in `LineSearches.jl` for v0.4.

yuyichao:
It doesn't look related....

pkofod:
Thanks for doing this

PR #366, opened 2017-02-18
Update deprecate.jl

pkofod:


pkofod:
@tkelman This should fix both the problems in NLsolve (ref https://github.com/EconForge/NLsolve.jl/pull/86 ) and Celeste. The splatting dots were missing in three deprecations.

tkelman:
if this fixes matters, then a new release here should be fine and we can hold off on adding bounds until we see whether it's really needed

pkofod:
Okay, Will merge and tag later

PR #375, opened 2017-02-22
Simplify some tests.

pkofod:
The idea is to remove a lot of the direct use of `*Differentiable` as we are moving towards discouraging this format for most users.

### DIFF
--- a/test/newton.jl
+++ b/test/newton.jl
@@ -11,13 +11,11 @@
         storage[1, 1] = 12.0 * (x[1] - 5.0)^2
     end
 
-    d = TwiceDifferentiable(f_1, g!_1, h!_1)
-
     # Need to specify autodiff!
     @test_throws ErrorException Optim.optimize(OnceDifferentiable(f_1, g!_1), [0.0], Newton())
     Optim.optimize(OnceDifferentiable(f_1, g!_1), [0.0], Newton(), Optim.Options(autodiff = true))
 
-    results = Optim.optimize(d, [0.0], Newton())
+    results = Optim.optimize(f_1, g!_1, h!_1, [0.0], Newton())
     @test_throws ErrorException Optim.x_trace(results)
     @test Optim.g_converged(results)
     @test norm(Optim.minimizer(results) - [5.0]) < 0.01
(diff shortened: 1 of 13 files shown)


test/newton.jl
```diff
@@ -70,9 +68,9 @@
         		ddf = OnceDifferentiable(prob.f, prob.g!)
```
anriseth: Do you need `ddf` now? Or should we just pass `prob.f, prob.g!, prob.h!` on line 69?

pkofod: Good question. I think I left it in just to check that all "valid" input variations would work. Maybe there should be more comments in the tests.

anriseth: Or we could have a separate testset for tests where we pass in *Differentiable types?

pkofod: That would work. I'll update this or to reflect that

pkofod: Or = pr (can't edit on phone)

test/newton.jl
```diff
@@ -50,29 +47,25 @@
...
+        res = optimize(prob.f, prob.g!, prob.h!, [0., 0.], Newton())
         @test norm(Optim.minimizer(res) - prob.solutions) < 1e-9
     end
 
     @testset "Optim problems (ForwardDiff)" begin
         for (name, prob) in Optim.UnconstrainedProblems.examples
         	if prob.istwicedifferentiable
-        		ddf = OnceDifferentiable(prob.f, prob.g!)
-        		res = Optim.optimize(ddf, prob.initial_x, Newton(), Optim.Options(autodiff = true))
         		@test norm(Optim.minimizer(res) - prob.solutions) < 1e-2
```
anriseth: This line no longer makes sense (res is not defined yet)

pkofod: I know, its filing locally ;)

pkofod:
I wonder if something breaks stuff on v0.6. Check this out
```

anriseth:
The v0.4 results are the same as v0.6.

pkofod:
> The v0.4 results are the same as v0.6.
> 
> EDIT: And the difference between v0.6 and v0.5 only happens for finite differencing, not for autodiff.

Strange, exactly the same packages are installed (same versions) no matter the JuliaLang vers. I guess I need to install v0.6 to inspect.

pkofod:
> EDIT: And the difference between v0.6 and v0.5 only happens for finite differencing, not for autodiff.

It's pretty weird. I get differences in results, but I don't get different gradients if I check them out manually at various `x`s across julia versions. More prints!

pkofod:
Oh my... I think I finally found out why the tests might be failing sometimes. It's the initialization of `y` in accelerated gradient descent that sometimes would contain a nan, such that `y_previous` would contain a nan, such that even if `y` is overwritten, `x` would still be corrupted in the end. I think that particular solver needs to be bug checked thoroughly, because it has had some weird performance (optimization wise) behavior in the past as well.

PR #377, opened 2017-02-25
Allowed x_upper == x_lower for univariate methods.

tpapp:
In this case, the iteration finished immediately. Also added tests,
and further tests for getting errors with x_upper < x_lower.


pkofod:
Thanks

PR #381, opened 2017-03-09
bump ForwardDiff version upper bound

jrevels:
The latest version of ForwardDiff is needed to support Julia v0.6 without deprecation warnings, but it drops v0.4 support. Since bumping the lower bound here would mean dropping v0.4 support from Optim, I only bumped the upper bound.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.4
 Calculus
-ForwardDiff 0.3.0 0.4.0
+ForwardDiff 0.3.0 0.5.0
 PositiveFactorizations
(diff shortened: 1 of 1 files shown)


pkofod:
I intend to move away from v0.4 on the next major tag, so... up to you

jrevels:
I'll leave the v0.4-dropping to somebody who knows their way around Optim better than myself :)

If you could make a tag after merging this I'd be super grateful! Thanks!

pkofod:
https://github.com/JuliaLang/METADATA.jl/pull/8266#event-993945737

PR #328, opened 2016-12-23
BFGS and Newton: reset initial linesearch step to one

anriseth:
Resetting the initial step for the linesearch to one after each iteration in Newton and BFGS improves convergence: 
Newton CUTEst tests with maxvar = 5:
f_err_cutest
BFGS CUTEst tests with maxvar = 5:
f_err_bfgs_cutest

BT3 is backtracking with cubic interpolation, added in anriseth/LineSearches.jl#20. The behaviour for quadratic bactracking is similar.
LBFGS already has a variant of this implemented, so I haven't added it there.
Ref #91

PR #388, opened 2017-03-16
LineSearches + NLSolversBase usage

anriseth:
Just for reference, this branch is needed to work with https://github.com/JuliaNLSolvers/LineSearches.jl/pull/37


test/api.jl
```diff
@@ -112,8 +112,8 @@
    @test isapprox(Optim.minimum(res), 0.0020622412076141045; rtol=1e-3)
    @test isapprox(Optim.minimizer(res), [0.9719007353489979,0.9410235857510793]; rtol=1e-3)
    @test Optim.iterations(res) == 10
-   @test Optim.f_calls(res) == 47
-   @test Optim.g_calls(res) == 47
+   @test Optim.f_calls(res) == 37
```
pkofod: Seems like this change actually makes sense!

pkofod:
Is `hessian!` clashing with the diff packages?

test/api.jl
anriseth: Yep, for problems with many nonlinear iterations, this will make a big difference on expensive functions:)
We should update the Rosenbrock example in the README after the transition as well, together with #328, the number of f and g-calls go down by about 100 on v0.5 :)

anriseth:
> Is hessian! clashing with the diff packages?

It complained about `Calculus` actually.

test/api.jl
pkofod: That's a pretty cool improvement.

pkofod:
> It complained about Calculus actually.

I type annotated `hessian!` in Base, this should solve that right?

Also, could you put all the constructors back in? I'm removing them from NLSolversBase. I don't think Base should depend on all those packages, I'd rather have those dependencies in higher level packages (Optim and NLsolve in particular).

Also, it's `value_gradient!` now, just to be consistent across our own NLSolversBase functions, but also the diff packages.

anriseth:
> If we want the gradient checker, it could be in a NLSolversTools or something like that. Then we might even place the "autodiff" constructors there as well, to keep Optim cleaner, but I'm not too sure about that.

I like that idea

anriseth:
I'll move the *Diff-constructors back here for now. Whenever an NLSolversTools is created, we can move them there if that seem to suit.

pkofod:
There is such a package already :)

pkofod:
I think we might just want the autodiff constructors in Optim still, until we figure out exactly how we want to do the tools-thing: OptimTools and NLsolveTools, or NLSolversTools? Probably the former, but I'm not sure. Also, the manual addition of NLSolversBase can be removed now that it is tagged and published.

pkofod:
Guess we're waiting on https://github.com/JuliaLang/METADATA.jl/pull/8415 https://github.com/JuliaLang/METADATA.jl/pull/8412 . Then we can revert the travis changes, and we should be good to go, no?

anriseth:
Sorry, I messed up the master-merges that you had pushed to this branch. Managed to forget to pull before I pushed the new changes.

REQUIRE
```diff
@@ -1,7 +1,8 @@
...
-ForwardDiff 0.3.0 0.5.0
-ReverseDiff 0.0.3
 PositiveFactorizations
 Compat 0.18.0
-LineSearches 0.1.2 1.0.0-
+LineSearches 1.0.0 2.0.0-
+Calculus
+NLSolversBase
+ForwardDiff
+ReverseDiff
```
tkelman: won't these still have minimum version requirements?

anriseth: Oh, yes. Thank you.

pkofod:
@anriseth is this ready?

anriseth:
I believe so, yes

pkofod:
Great, thank you for the PR @anriseth , and thank you for the assistance @tkelman

REQUIRE
```diff
@@ -4,5 +4,5 @@ Compat 0.18.0
 LineSearches 1.0.0 2.0.0-
 Calculus
 NLSolversBase
-ForwardDiff
-ReverseDiff
+ForwardDiff 0.3.0 0.5.0
```
tkelman: was this in anticipation of future breaking changes? worth testing whether Optim works with ForwardDiff 0.5 now that it's been released

anriseth: Yes it was. I'll see what Travis says about 0.5

tkelman: If a ForwardDiff-0.5-compatible version doesn't get released soon, then on julia 0.6 Pkg may choose to downgrade Optim all the way back to v0.5.0, which was the most recent version not to depend on ForwardDiff at all, in order to upgrade ForwardDiff to 0.5.

KristofferC: Yeah I saw this happen.

PR #392, opened 2017-04-07
Change order of storage and evaluation points.

pkofod:


pkofod:
Needs LineSearches to be tagged in order to pass

pkofod:
I will deprecate this in `NLSolversBase` by checking if `g!` changes `x` or `g` in the constructor, and defining `(g, x)->g!(x, g)` in case the users did it the wrong way around, throw a warning, and then continue.

PR #394, opened 2017-04-15
Make state an argument to optimize such that it can be reused

andreasnoack:


pkofod:
Thanks. I was going to get around to this, as someone over at stackexchange wanted to access `invH` from BFGS. This will allow them to do exactly that.

The only question is, with the state-instance as a positional argument, would it make sense to introduce an `optimize!`?

pkofod:
I'll merge it, but do note that this will probably be changed to an `optimize!` function before this change gets tagged. As in: `optimize!(obj, initial_x, method, options = Options(), state = initial_state(method, options, d, initial_x)`

andreasnoack:
We are departing from the rule for other functions that are obviously stateful, e.g. `write` and since it doesn't modify any non-state variables, I think it is okay to leave out the exclamation mark. I think it is expected that a state variable will be modified when passed to a function.

pkofod:
I get your point. Has there been a broader discussion about this, or is it just "happening "?

andreasnoack:
I'm pretty sure it has been discussed somewhere in Base but it is a while back and I don't remember which function.

PR #396, opened 2017-04-19
Simplify state by moving name string into a method and removing the

andreasnoack:
size attribute. Since the default arguments macro would only have
a single field, I've also removed the macro.

### DIFF
--- a/src/fminbox.jl
+++ b/src/fminbox.jl
@@ -100,14 +100,17 @@ function precondprepbox!(P, x, l, u, mu)
     end
 end
 
-immutable Fminbox <: Optimizer end
+immutable Fminbox{T<:Optimizer} <: Optimizer end
+Fminbox() = Fminbox{ConjugateGradient}() # default optimizer
 
-function optimize{T<:AbstractFloat}(
+Base.summary{O}(::Fminbox{O}) = "Fminbox with $(summary(O()))"
+
+function optimize{T<:AbstractFloat,O<:Optimizer}(
         df::OnceDifferentiable,
         initial_x::Array{T},
         l::Array{T},
         u::Array{T},
-        ::Fminbox;
+        ::Fminbox{O};
         x_tol::T = eps(T),
         f_tol::T = sqrt(eps(T)),
         g_tol::T = sqrt(eps(T)),
@@ -123,13 +126,12 @@ function optimize{T<:AbstractFloat}(
         mu0::T = convert(T, NaN),
         mufactor::T = convert(T, 0.001),
         precondprep = (P, x, l, u, mu) -> precondprepbox!(P, x, l, u, mu),
-        optimizer = ConjugateGradient,
         optimizer_o = Options(store_trace = store_trace,
                                           show_trace = show_trace,
                                           extended_trace = extended_trace),
         nargs...)
 
-    optimizer == Newton && warning("Newton is not supported as the inner optimizer. Defaulting to ConjugateGradient.")
+    O == Newton && warning("Newton is not supported as the inner optimizer. Defaulting to ConjugateGradient.")
     x = copy(initial_x)
     fbarrier = (gbarrier, x) -> barrier_box(gbarrier, x, l, u)
     fb = (gbarrier, x, gfunc) -> function_barrier(gfunc, gbarrier, x, df.fg!, fbarrier)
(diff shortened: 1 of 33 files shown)


pkofod:
How can the mac travis runs get a 
```
Constrained: Error During Test

  Got an exception of type ErrorException outside of a @test

  type BFGSState has no field n
```
but the linux ones doesn't, and neither does AV?

pkofod:
:+1:  overall, and it's much cleaner than `state.n` - I think I went a little overboard with what was worth calculating up front :)

Is there a reason you avoided `name` for example? As in `name(method)` instead of `Optim.method(method))`?

andreasnoack:
I wasn't sure about what to call the method but saw that `api.jl` already defined `method`. Actually, I think the right thing to do here might be to overload `Base.summary`.

pkofod:
> I wasn't sure about what to call the method but saw that api.jl already defined method. Actually, I think the right thing to do here might be to overload Base.summary.

`summary` could work, but tbh `method` could also work, as this is not exported

andreasnoack:
I might have gone a little too far here so I can more the last part to a separate PR if necessary. Instead of storing the optimization method information as strings, I've changed the types to store an `Optimizer`. Only in the show/printing method, the type is converted to a string via `summary`. A consequence of the current version is that the way you specify the optimization procedure in `Fminbox` has changed. Now it is a type parameter to `Fminbox`.

anriseth:
> A consequence of the current version is that the way you specify the optimization procedure in Fminbox has changed. Now it is a type parameter to Fminbox.

I think this is an improvement :)

andreasnoack:
@pkofod Regrading https://github.com/JuliaNLSolvers/Optim.jl/pull/396#issuecomment-295156394 then it is pretty puzzling. I realized that I haven't changed the `state.n` occurrences in `perform_linesearch.jl` but does this mean that Mac and Linux take different code paths?

pkofod:
> @pkofod Regrading #396 (comment) then it is pretty puzzling. I realized that I haven't changed the state.n occurrences in perform_linesearch.jl but does this mean that Mac and Linux take different code paths?

Yes, they can certainly take different paths. Sometimes, an iteration will throw off the Mac travis runs, for example if the problem isn't well-conditioned. It's been observed in other PRs as well.

andreasnoack:
Okay. I've updated the last uses of `state.n` so the tests pass now. Any thoughts on the `String`->`Type` changes?

pkofod:
> Any thoughts on the String->Type changes?

I like it. Just one thing. Isn't the `summary` method missing for `Fminbox`? Or am I missing something here.

src/fminbox.jl
```diff
@@ -100,14 +100,17 @@ function precondprepbox!(P, x, l, u, mu)
     end
 end
 
-immutable Fminbox <: Optimizer end
+immutable Fminbox{T<:Optimizer} <: Optimizer end
+Fminbox() = Fminbox{ConjugateGradient}() # default optimizer
 
-function optimize{T<:AbstractFloat}(
+Base.summary{O}(::Fminbox{O}) = "Fminbox with $O"
```
pkofod: Won't this print something like
```
"Fminbox with Optim.ConjugateGradient{T,Tprep<:Union{Function,Void},L<:Function}"
```
as opposed to
```
"Fminbox with Conjugate Gradient"
```
?
(sorry for nitpicking)

andreasnoack: No problem, it's fine. I've fixed it but I'm actually wondering if `Fminbox` should take instances instead of types as the optimization method argument. It seems more natural to specify the method specific parameters through the instance instead of through extra arguments to `optimize` for `Fminbox`. In practice, we could just add an `Optimizer` field to `Fminbox`. However, that is a PR for another day.

pkofod:
> In practice, we could just add an Optimizer field to Fminbox.

That would make sense. To be honest, `Fminbox` is in need of a lot of love. For a while, it's just been "kept alive" with no real improvements.

src/fminbox.jl
```diff
@@ -100,14 +100,17 @@ function precondprepbox!(P, x, l, u, mu)
     end
 end
 
-immutable Fminbox <: Optimizer end
+immutable Fminbox{T<:Optimizer} <: Optimizer end
+Fminbox() = Fminbox{ConjugateGradient}() # default optimizer
```
pkofod: Hm, this change should probably have been deprecated/updated its documentation. The old version doesn't work, and it isn't documented how to use the new interface. For example, `Fminbox` ( the type) along with the `optimizer` keyword should have created the default `Fminbox{O}` instance and thrown a warning. I'll see if I can get around to it later.

pawbz: Thank you :) That helps

PR #397, opened 2017-04-19
Reorganize optimize, and allow callables other than Functions as f g …

pkofod:
…or h.


src/optimize/multivariate/interface.jl
```diff
@@ -0,0 +1,112 @@
...
+end
+
+function optimize{M <: Union{FirstOrderSolver, SecondOrderSolver}}(f, initial_x::AbstractArray, method::M, options::Options)
+    d = ifelse(M <: FirstOrderSolver, OnceDifferentiable(f, initial_x), TwiceDifferentiable(f, initial_x))
+    optimize(d, initial_x, method, options)
+end
+function optimize{M <: Union{FirstOrderSolver, SecondOrderSolver}}(f::Function, initial_x::AbstractArray, method::M, options::Options)
+    d = ifelse(M <: FirstOrderSolver, OnceDifferentiable(f, initial_x), TwiceDifferentiable(f, initial_x))
+    optimize(d, initial_x, method, options)
+end
```
Evizero: why the two versions? ambiguities ?

Evizero: (once with `f` and once with `f::Function`, while the rest stays the same)

pkofod:
Seems like I need to look at this at v0.6 as well due to the type system changes

src/optimize/multivariate/interface.jl
pkofod: You mean why even have the `::Function` version and not leave it unspecified altogether?

Evizero: yes. Just curious.

pkofod: Don't really have a good answer. For some reason I just kept the `::Function` annotated ones, and added the ones with wider signatures. Removed them.

pkofod:
`multivariate/interface.jl` should be *much* easier to read now (at least it is to me...)

PR #398, opened 2017-04-19
Fix array.jl tests

pkofod:


pkofod:
The problem here is with PSO and some problem/method combinations that are problematic (and well-known). I'll fix it after the other PR

PR #400, opened 2017-04-21
More informative wording in case optimization was stopped by an increasing objective

pkofod:


pkofod:
*something* went wrong with git here (shouldn't started afresh from a clean master), but I'll squash anyway.

PR #401, opened 2017-04-23
Update README output

anriseth:
Seems like our algorithms have improved :+1:


pkofod:
Not bad

anriseth:
> Not bad. I guess a lot of these come from the refactoring of d.f(x) and friends, although that won't explain the 16 vs 21 major iterations.

The majors should be due to #328

pkofod:
It probably is. There has been no other changes to bfgs, so. Great stuff!

PR #403, opened 2017-04-26
Clean up folder structure.

pkofod:
I know this is a matter of taste, but I just couldn't stand the "one folder for all files" approach anymore.


src/Optim.jl
```diff
@@ -55,36 +55,39 @@ module Optim
...
 
     # preconditioning functionality
-    include("precon.jl")
+    include("multivariate/precon.jl")
 
     # Gradient Descent
-    include("gradient_descent.jl")
-    include("accelerated_gradient_descent.jl")
-    include("momentum_gradient_descent.jl")
+    include("multivariate/solvers/first order/gradient_descent.jl")
```
anriseth: Not a biggie, but I'm not a fan of spaces in folder/file names. Would you be okay with using `first_order` etc.?

pkofod: Sure! I was actually just thinking if it could maybe cause problems for some users/systems/setups.

pkofod:
Probably we can do more, for example for some of the `types.jl`-stuff.

PR #405, opened 2017-04-26
Clean up folder structure. (#403)

pkofod:
* Clean up folder structure.
* Don't print just because the extended trace is asked for.


KristofferC:
Out of curiosity, what was wrong with the previous one?

pkofod:
The `objective` in `cg.jl` conflicted with the `global objective` in `constrained.jl` on v0.5 builds.

PR #406, opened 2017-05-08
Prepare for a Switching LineSearches.jl to dispatch

cortner:
In all the first order methods I've simply removed the requirement that a `linesearch!` must be a function. This allows to plug in a type, which can be made callable. That will allow for a smooth transition from line search functions to line search types. 

In a second step, one can introduce a type `AbstractLinesearch` in `Linesearches.jl`, and then `linesearch!` could be made a `Union{Function, AbstractLinesearch}`. But I'd leave this as a second step, or even remove the possibility of it being a function altogether.

CC @pkofod  , @anriseth


pkofod:
This had to be done regardless, so :+1:

PR #407, opened 2017-05-08
Optimisation Results

cortner:
I added just a little more information to `MultivariateOptimisationResults`. Currently, they only tell whether or not the iteration was successful but not by how much the tolerance was missed. So I just added printouts for |x-x'| and |g|. This now looks like this:
```
 * Algorithm: Conjugate Gradient
 * Starting Point: [0.0,0.0, ...]
 * Minimizer: [-1.069462778281003e-15,-1.0648308460784488e-15, ...]
 * Minimum: -6.451059e+01
 * Iterations: 4
 * Convergence: true
   * |x - x'| < 1.0e-32: false 		(|x-x'| = 1.668135812371574e-5)
   * |f(x) - f(x')| / |f(x)| < 1.0e-32: false
   * |g(x)| < 1.0e-05: true 		(|g(x)| = 1.1704999897781426e-7)
   * stopped by an increasing objective: false
 * Objective Calls: 13
 * Gradient Calls: 10
```

Notes: 
* Should probably to the same for the f-tolerance, but I don't care much for that one to begin with
* formatting might not be ideal (see above)
* this is a bit of a hack (e.g. see exclusion of `NelderMead` via `try` + `catch`) 

I can try to fix these issues myself, maybe with some helpful discussion and update the PR, or I'd also be happy if it just gets merged and cleaned up subsequently. But I feel strongly that this is missing information that ought to be included.


anriseth:
I agree, those values are quite useful.

johnmyleswhite:
If you're concerned about formatting, maybe it would be more readable with additional line breaks:

```
 * Algorithm: Conjugate Gradient
 * Starting Point: [0.0,0.0, ...]
 * Minimizer: [-1.069462778281003e-15,-1.0648308460784488e-15, ...]
 * Minimum: -6.451059e+01
 * Iterations: 4
 * Convergence: true
   * |x - x'| < 1.0e-32: false
       * |x-x'| = 1.668135812371574e-5
   * |f(x) - f(x')| / |f(x)| < 1.0e-32: false
       * |f(x) - f(x')| / |f(x)| = FOO
   * |g(x)| < 1.0e-05: true
       * |g(x)| = 1.1704999897781426e-7
   * stopped by an increasing objective: false
 * Objective Calls: 13
 * Gradient Calls: 10
```

pkofod:
Indentation and line break seems sufficient to make it readable. :+1: on the change in general

cortner:
* This PR update adds f-tolerance but I think this will show a problem with how f-tolerance is implemented: when `|f - f_previous| < eps` then this is also considered as "converged", but this test does not show up in the results printing.
* I haven't removed the `try`, `catch` in optimize.jl:l71. In all my "application tests" this runs ok, but an exception gets thrown here several times while running `Pkg.test("Optim")`. I don't really understand what happens here.

pkofod:
You need to update the tests that check for the output of an optimization result.

cortner:
those are the failures?

cortner:
this ok? (there are some other failures but those are due to LineSearches)

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -218,7 +218,10 @@ function optimize{T<:AbstractFloat,O<:Optimizer}(
         f_increased && !allow_f_increases && break
     end
     return MultivariateOptimizationResults(Fminbox{O}(), initial_x, minimizer(results), df.f(minimizer(results)),
-            iteration, results.iteration_converged, results.x_converged, results.x_tol, results.f_converged,
-            results.f_tol, results.g_converged, results.g_tol, results.f_increased, results.trace, results.f_calls,
+            iteration, results.iteration_converged,
+            results.x_converged, results.x_tol, vecnorm(x - xold),
+            results.f_converged, results.f_tol,
```
pkofod: there's a term missing after `results.f_tol`, which is causing the failure in CI

cortner:
ah - sorry about this. I hope this fixes it. (I think I was using the wrong version of LineSearches before for testing and so got confused which tests failed because of the changes I made.)

pkofod:
Seems like something is off in LineSearches 2.1.0, we need to fix it and patch https://ci.appveyor.com/project/pkofod/optim-jl/build/1.0.296/job/wkqv94i0an8yqork @anriseth (not saying you need to do it, just pinging)

cortner:
Right so with 2.0 the tests look ok

pkofod:
> Right so with 2.0 the tests look ok
Are you telling me so, or is it meant more like a question ? :) Can you post an example of how it looks now?

cortner:
```
 * Algorithm: BFGS
 * Starting Point: [0.0,0.0]
 * Minimizer: [0.9999999926033423,0.9999999852005353]
 * Minimum: 5.471433e-17
 * Iterations: 16
 * Convergence: true
   * |x - x'| < 1.0e-32: false
     |x - x'| = 3.47e-07
   * |f(x) - f(x')| / |f(x)| < 1.0e-32: false
     |f(x) - f(x')| / |f(x)| = NaN
   * |g(x)| < 1.0e-08: true
     |g(x)| = 2.33e-09
   * stopped by an increasing objective: false
 * Objective Calls: 53
 * Gradient Calls: 53
```

pkofod:
Is this worse or better?
```
 * Algorithm: BFGS
 * Starting Point: [0.0,0.0]
 * Minimizer: [0.9999999926033423,0.9999999852005353]
 * Minimum: 5.471433e-17
 * Iterations: 16
 * Convergence: true
   * |g(x)|   = 2.33e-09 < g_tol = 1.0e-08: true
   * |x - x'| = 3.47e-07 < x_tol = 1.0e-32: false
   * |f(x) - f(x')| / |f(x)| = NaN < f_tol = 1.0e-32: false
   * stopped by an increasing objective: false
 * Objective Calls: 53
 * Gradient Calls: 53
```
in my head it reads like: the norm of the gradient is equal to this number and we are checking if it is smaller than g_tol that is this number and it is true

cortner:
i'm ok with your first option. harder to read but all the information is there

cortner:
we could also *properly* redesign this, e.g. with tables and colours, but I just don't have that time right now and since it wouldn't be a breaking change it can be done anytime as a patch, no?

pkofod:
> we could also properly redesign this, e.g. with tables and colours, but I just don't have that time right now and since it wouldn't be a breaking change it can be done anytime as a patch, no?

I think we should, and I don't prioritize it myself either. I'm not asking you to change anything here, I was just throwing out some ideas (very late in the process) we could consider at another point. 

I will tag a minor version as soon as nlsolversbase and linesearches is sorted out

cortner:
I'm happy to update to your version of the output - just let me know which you prefer, I don't really mind either way.

pkofod:
> I'm happy to update to your version of the output - just let me know which you prefer, I don't really mind either way.

Well, I did it to avoid two occurrences of `|x-x'||  etc, but I think it becomes a bit too dense tbh.

pkofod:
> we could also properly redesign this, e.g. with tables and colours,

I think we need a larger redesign than that for v2.0.0. We need to modularize the package even more, but with sensible defaults that will be very much like what we have now. There's a plan for it, but I just havn't gotten around to it.

cortner:
Ok, let's play a bit 
```
Results of Optimization Algorithm (BFGS)
 * Minimizer: [0.9999999926033423,0.9999999852005353]
 * Minimum: 5.471433e-17
 * Statistics:   Iterations | Objective Calls | Gradient Calls 
                                   16  |                   53  |                 53
 * Convergence: true
             residual |    | x - x'|       |    |f-f'|/|f|     |       |g|  
           achieved |      3.47e-07 |        NaN      |     2.33e-09
                target |     1.0e-32     |     1.0e-32   |     1.0e-08
 * Failure  |  increasing objective   |   Maximum Number of Iterations
                 |                   false             |    false 
```

pkofod:
> Ok, let's play a bit

whitespace seems off in the browser, but I like where you're going with it

cortner:
anyways, all success could be printed in green and all failure in red

anriseth:
> Seems like something is off in LineSearches 2.1.0, we need to fix it and patch 

Should be fixed in https://github.com/JuliaNLSolvers/LineSearches.jl/pull/51
The tests in LineSearches are not good enough it seems. I should remind myself to always run the tests in Optim (and maybe NLsolve) as well.

pkofod:
https://github.com/JuliaLang/METADATA.jl/pull/9438 and we're good to go (after a rerun of CI)

pkofod:
> The tests in LineSearches are not good enough it seems. I should remind myself to always run the tests in Optim (and maybe NLsolve) as well.

We should add cross-package tests the moment Optim is tagged

pkofod:
Thanks

PR #409, opened 2017-05-08
Move initial condition inside box

anriseth:
I prefer to warn the user and move their initial guess inside the box, rather than error and exit.

The updated values are rather arbitrary.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,5 @@
 # Optim v1.0.0
-* Fminbox: If an initial guess is not in the interior of the box, the guess is moved inside the box and a warning is produced, as opposed to exting with an error.
+* Fminbox: If an initial guess is on the boundary of the box, the guess is moved inside the box and a warning is produced, as opposed to crashing with an error.
 * Significant changes to the Non-, Once-, and TwiceDifferentiable setup; these now hold temporaries relevant to the evaluation of objectives, gradients, and Hessians. They also hold f-, g-, and h_calls counters
 * Refactor tests
 * Drop v0.4 support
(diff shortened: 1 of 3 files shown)


pkofod:
Any chance you could just mention it briefly? It is a behavior chance after all. But other than that, fine by me.

anriseth:
Do you mean mention in NEWS?

pkofod:
That's what I mean, I missed a few words there :)

KristofferC:
FWIW I disagree with this. If the user provides obviously faulty data then a function should just error out quickly and not guess what was intended. It is just as likely that the bounds where entered incorrectly as the initial condition.

anriseth:
Okay. How about when `x[i] == l[i]` or `x[i] == u[i]`, that is when the
user sets the initial guess to be *on* the boundary?
This happens to me quite often when I call Fminbox as a small part of a
bigger simulation, and is the main problem I wanted to fix with this PR.

KristofferC:
That sounds like a good idea.

pkofod:
New version is :+1:

There's a `@test_warn` macro on v0.6 at least, that lets you test for an exact string, or if a regex matches the warning message.

anriseth:
This is ready to merge. Newton trust region is killing Travis on Julia 0.5, but the tests on constrained optimisation passed:)

pkofod:
In a sense this is still correcting behavior that could indicate that the user did someething wrong, and didn't intend for the code to be run with wrong starting values. However, I do appreciate that to many users, "on the boundary" seems like a legal place to start, it is just that our barrier function does not like it. Unless someone disagrees, we can go ahead and merge this.

anriseth:
> In a sense this is still correcting behavior that could indicate that the user did someething wrong, and didn't intend for the code to be run with wrong starting values. However, I do appreciate that to many users, "on the boundary" seems like a legal place to start, it is just that our barrier function does not like it.

Let's hope the warning is big enough for them to realise that they've done something bad

PR #411, opened 2017-05-17
switched all line search functions to line search types

cortner:
this PR is the `Optim.jl` counterpart to [PR #43 in LineSearches](https://github.com/JuliaNLSolvers/LineSearches.jl/pull/43)

PR #413, opened 2017-05-20
Add tests for f,g,h counters

anriseth:
Requires #411

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.5
 PositiveFactorizations
 Compat 0.18.0
-LineSearches 2.0.0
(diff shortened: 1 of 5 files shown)


pkofod:
This should probably change LineSearches require to 2.1.0

anriseth:
I've bumped the version. 
Tests will fail until https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/10 is merged. So until then, we could either wait with this PR, or I can "fix" the tests by adding extra counts

pkofod:
Could you do a version that works now, with the "future" versions commented out?

pkofod:
Ready?

PR #416, opened 2017-05-21
WIP: CG/Krylov Trust Region method

### LINKED ISSUE #356: Hessian-vector products
Have you all considered adding optimization methods that make use of Hessian-vector products, but that don't explicitly form Hessians? I've been thinking about writing a version of `newton_trust_region` that does that, essentially using conjugate gradient iterations to multiply the gradient by the inverse Hessian. Is that something you'd be interested in including in Optim.jl?

jeff-regier/Celeste.jl#380

jeff-regier:
@pkofod  Here's a PR for a trust region method that uses Hessian-vector products. I thought I'd post it now since I'm not sure how you want Optim's types to handle functions that are twice differentiable but only support Hessian-vector products. In the PR, I added a new function type `TwiceDifferentiableHV` but I thought you might be able to suggest a better way.

Also, please let me know what other changes and additions to make to get the PR merged.

Once complete, this PR closes #356.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -80,6 +80,7 @@ module Optim
     # Newton
     include("multivariate/solvers/second_order/newton.jl")
     include("multivariate/solvers/second_order/newton_trust_region.jl")
+    include("multivariate/solvers/second_order/krylov_trust_region.jl")
 
     # Constrained optimization
     include("multivariate/solvers/constrained/fminbox.jl")
--- a/src/multivariate/solvers/second_order/krylov_trust_region.jl
+++ b/src/multivariate/solvers/second_order/krylov_trust_region.jl
@@ -1,0 +1,219 @@
+immutable TwiceDifferentiableHV
+    f::Function
+    fg!::Function
+    hv!::Function
+end
+
+
+immutable KrylovTrustRegion{T <: Real} <: Optimizer
+    initial_radius::T
+    max_radius::T
+    eta::T
+    rho_lower::T
+    rho_upper::T
+    cg_tol::T
+end
+
+
+KrylovTrustRegion(; initial_radius::Real = 1.0,
+                    max_radius::Real = 100.0,
+                    eta::Real = 0.1,
+                    rho_lower::Real = 0.25,
+                    rho_upper::Real = 0.75,
+                    cg_tol::Real = 0.01) =
+                    KrylovTrustRegion(initial_radius, max_radius, eta,
+                                  rho_lower, rho_upper, cg_tol)
+
+
+type KrylovTrustRegionState{T,N,G}
+    x::Vector{T}
+    g::G
+    f_x_previous::T
+    x_previous::Vector{T}
+    s::Vector{T}
+    interior::Bool
+    accept_step::Bool
+    radius::T
+    m_diff::T
+    f_diff::T
+    rho::T
+    r::Vector{T}  # residual vector
+    d::Vector{T}  # direction to consider
+    Hd::Vector{T} # hessian-vector product
+    cg_iters::Int64
+end
+
+
+function initial_state{T}(method::KrylovTrustRegion, options, d, initial_x::Array{T})
+    n = length(initial_x)
+    # Maintain current gradient in gr
+    @assert(method.max_radius > 0)
+    @assert(0 < method.initial_radius < method.max_radius)
+    @assert(0 <= method.eta < method.rho_lower)
+    @assert(method.rho_lower < method.rho_upper)
+    @assert(method.rho_lower >= 0)
+
+    g = similar(initial_x)
+    f_x = d.fg!(initial_x, g)
+
+    KrylovTrustRegionState("Krylov Trust Region",
+                         length(initial_x),
+                         copy(initial_x), # Maintain current state in state.x
+                         f_x, # Store current f in state.f_x
+                         1, # track f calls in state.f_calls
+                         1, # track g calls in state.g_calls
+                         0,  # track hessian-vector products
+                         g, # Store current gradient in state.g
+                         0., # f_x_previous
+                         similar(initial_x), # x_previous
+                         similar(initial_x), # Maintain current search direction in state.s
+                         true,  # interior
+                         true,  # accept step
+                         method.initial_radius,
+                         0., # model change
+                         0., # observed f change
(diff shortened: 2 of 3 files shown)


KristofferC:
I think everything written as e.g. `z[:] += tau * d` could be written with the fusing dot syntax as `z .+= tau .* d`. This should avoid forming any temporaries and work totally in place.

pkofod:
> @pkofod Here's a PR for a trust region method that uses Hessian-vector products. I thought I'd post it now since I'm not sure how you want Optim's types to handle functions that are twice differentiable but only support Hessian-vector products. In the PR, I added a new function type TwiceDifferentiableHV but I thought you might be able to suggest a better way.

I'll have a look! Great to see you getting around to this. Have you played around with it in Celeste?

pkofod:
You may not have noticed, but we moved the `*Differentiable`-stuff to https://github.com/JuliaNLSolvers/NLSolversBase.jl so that is where `TwiceDifferentiableHV` should go. I'm not sure I can think of a better solution than yours off the top of my head... I like that you placed all methods in the same file. We should really move all the trace code for the other solvers to their respective files as well.

jeff-regier:
It's running on Celeste's `jcr/cg2` branch now, with Hessian-vector products computed by ForwardDiff.jl. It works well at getting the right result. It can take a lot of Hessian-vector products though, for our application. I'm working on ways of making it more efficient. (The implementation in this PR is pretty much textbook N&W.) There's an extension to the N&W approach that uses Lanczos iterations as well as CG iterations that I'm exploring.

jeff-regier:
What about adding an `hv!` field to `TwiceDifferentiable` in NLSolversBase, that gets populated in the constructor automatically if the caller specifies `h!` but not `hv!`? And if a constructor is called with `hv!` but not `h!`, then `h!` is set to a function that throws a "not implemented" exception if called?

That way there's just one `TwiceDifferentiable` type, differentiability is really a property of the underlying function, not the ways of observing the function.

An alternative might be to do something with type parameters, to avoid any possibility of the `h!` method throwing an exception, but I'm not exactly sure how that would work.

pkofod:
> What about adding an hv! field to TwiceDifferentiable in NLSolversBase, that gets populated in the constructor automatically if the caller specifies h! but not hv!? And if a constructor is called with hv! but not h!, then h! is set to a function that throws a "not implemented" exception if called?
> 
> That way there's just one TwiceDifferentiable type, differentiability is really a property of the underlying function, not the ways of observing the function.
> 
> An alternative might be to do something with type parameters, to avoid any possibility of the h! method throwing an exception, but I'm not exactly sure how that would work.

Had not seen your response. Do you mean something like keeping fields for `h!` and `hv!` and then use a keyword to specify which one it is? 
```julia
...
hv! = ...
td = TwiceDifferentiable(..., hv!, ; hessianvector = true)
```
?

jeff-regier:
Yes, that sounds like a good way to do it. I hadn't thought through whether we'd use a keyword argument like `hessianvector = true`, but I suppose that makes sense.

pkofod:
> Yes, that sounds like a good way to do it. I hadn't thought through whether we'd use a keyword argument like hessianvector = true, but I suppose that makes sense.

Just to understand your earlier comment correctly, you would say that this works but didn't prove to be particularly fast compared to your existing approach in Celeste, or?

jeff-regier:
Yes, that's the case. For Celeste we're only optimizing 44 parameters at once so forming a dense hessian isn't all that slow. We'll nonetheless probably switch to using a krylov method eventually because it's so tedious to modify any code that computes the Hessian.

pkofod:
Alright, I take it you are fine that we wait with the merge until we converge on the best way to specify that hessian-vector products are to be used. In the mean time, I'll try to find some good test cases with large hessians

jeff-regier:
Yes, no rush for me, I can wait until hessian-vector products are in NLSolversBase. @rgiordan wanted to use this code eventually for his work, but I'm not sure how soon he needs it.

rgiordan:
Yes, if I'm going to use Julia for my research projects I'll need something that does second-order optimization using Hessian-vector products.  If this isn't checked in I can look into either trying to maintain a  version myself for private use or use port from C or Python.

anriseth:
> In the mean time, I'll try to find some good test cases with large hessians

> Yes, if I'm going to use Julia for my research projects I'll need something that does second-order optimization using Hessian-vector products. 

Sounds to me like @rgiordan may have some test cases for you @pkofod ?

rgiordan:
I do have use cases!  When you're doing perturbation analysis on an optimization problem with a large number of latent variables (e.g. large GLMMs fit with variational inference), you need to be at a high-quality optimum in terms of the gradient being very close to zero and the Hessian actually being positive definite.  I've found that it's necessary to use actual second-order optimization to get there -- quasi-second order methods like BFGS doesn't do a good job.  And if you have a lot of latent variables, calculating the exact Hessian can be prohibitive, at least with autodiff.

I have a paper draft for motivation and python code to share, but since it's still unpublished I'd rather share it privately.  @pkofod , if you're interested, please email me (rgiordano@berkeley.edu).

pkofod:
Just had a talk with @rgiordan , and as I told him, I basically think we might as well just merge this as is, and then we can try the design out before moving anything to a base package that affects other packages than Optim.

There are some changes I want to do here to make it up-to-date with other stuff in Optim, but I can do that myself after a merge. What's most important, I think, is to have a working implementation that we can play around with. We'll just list it as experimental, or just not list it at all ;)

jeff-regier:
There isn't a lot left to do---just rebasing, enabling the unit tests (the were running locally but now they are commented out because of breaking changes), and adding some basic documentation. But unfortunately I don't think I'm going to have time to work on it during the next couple of months. I will get to it then. Or I'm glad for someone else to take over this PR.

pkofod:
> I will get to it then. Or I'm glad for someone else to take over this PR.

I will see if I can find the time, would you maybe have time to answer a few basic questions if necessary?

jeff-regier:
Yes, glad to answer any questions. Thanks for any help getting this merged.

rgiordan:
Awesome, thanks everyone!  I'll give it a spin.

jeff-regier:
Thanks for seeing it through @pkofod !

PR #418, opened 2017-05-23
Use copy when storing Nelder-Mead trace

lstagner:
I was using the Nelder-Mead solver and the trace was only storing the final value. This was because the solver was reusing memory. This small changes fixes the issue.

### DIFF
--- a/src/utilities/trace.jl
+++ b/src/utilities/trace.jl
@@ -2,7 +2,7 @@ function trace!(tr, d, state, iteration, method::NelderMead, options)
     dt = Dict()
     if options.extended_trace
         dt["centroid"] = copy(state.x_centroid)
-        dt["step_type"] = copy(state.step_type)
+        dt["step_type"] = state.step_type
     end
     update!(tr,
     iteration,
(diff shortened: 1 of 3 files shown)


timholy:
`deepcopy` for generic objects is really slow, might want to benchmark the implications of this change.

pkofod:
Thank you for the bug report and PR! I think this is the wrong way to go about it though. As Tim mentions, deepcopy can be expensive, but more than that we simply shouldn't generate `dt` incorrectly. The problem lies at line 4 and 5 here https://github.com/JuliaNLSolvers/Optim.jl/blob/810253d19b2ad63175f3857592efe8634bcfe58b/src/utilities/trace.jl#L5 . Those should have copy around them! I'll be more than happy to merge a version of this where you drop the deepcopy, and add copies around the two offending lines. Please also add a test such that this doesn't regress in the future.

Thanks!

pkofod:
Okay, only around the centroid then :) no copy method for string in v0.6 due to the string changes.

lstagner:
This should be good to go.

pkofod:
Thanks

PR #420, opened 2017-05-25
No upper bound on ForwardDiff

anriseth:
We were constraining ForwardDiff to 0.4, but 0.5 should work as well.
Do we want to have upper bounds on packages in anticipating breakages? I think we should either have it on all, or none

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,5 +4,5 @@ Compat 0.18.0
 LineSearches 2.1.0
 Calculus
 NLSolversBase
-ForwardDiff 0.3.0 0.5.0
+ForwardDiff 0.3.0 
 ReverseDiff 0.0.3


anriseth:
We could also push this to a branch starting from the 0.7.8 tag, so we can tag a non-breaking 0.7.9 that does not mess up things for the Forward Diff update

pkofod:
A v0.7.9 would prob be the best option

anriseth:
> A v0.7.9 would prob be the best option

Aite. Is there anything I can/need to do for that? Or maybe it is easier for anyone with write access to just create their own upstream branch and tag that, and then we can ignore this PR.

anriseth:
Thanks for sorting out 0.7.* @pkofod. Can you merge this so master works well with ForwardDiff as well, please?

pkofod:
The problem here is that v0.5 only works on Julia v 0.6 so i Think we might need to tag a version soon without it and Them a V1.0.0 with this change for v0.6 only

PR #421, opened 2017-05-30
Move to v0.6 and drop ReverseDiff for now

pkofod:
Sorry @anriseth but we have to wait for ReverseDiff to get up to speed wrt ForwardDiff :)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,5 +3,5 @@ PositiveFactorizations
 Compat 0.18.0
 LineSearches 2.1.0
 Calculus
-NLSolversBase
-ForwardDiff
+NLSolversBase 2.1.0
+ForwardDiff 0.5.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6.0-pre
 PositiveFactorizations
 Compat 0.18.0
 LineSearches 2.1.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,5 +4,4 @@ Compat 0.18.0
 LineSearches 2.1.0
 Calculus
 NLSolversBase
-ForwardDiff 0.3.0 0.5.0
-ReverseDiff 0.0.3
+ForwardDiff
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,14 @@
 environment:
   matrix:
-    - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
-    - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
+    - JULIAVERSION: "julialang/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
+    - JULIAVERSION: "julialang/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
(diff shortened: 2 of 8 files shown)


REQUIRE
```diff
@@ -1,8 +1,7 @@
-julia 0.5
+julia 0.6.0-pre
 PositiveFactorizations
 Compat 0.18.0
 LineSearches 2.1.0
 Calculus
 NLSolversBase
-ForwardDiff 0.3.0 0.5.0
-ReverseDiff 0.0.3
+ForwardDiff
```
KristofferC: Should this be lower bounded to 0.5.0?

pkofod: It should indeed, thanks.

pkofod:
the goal here is to get https://github.com/JuliaLang/METADATA.jl/pull/9585 through

appveyor.yml
```diff
@@ -1,11 +1,14 @@
...
-    - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
+    - JULIAVERSION: "julialang/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
+    - JULIAVERSION: "julialang/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
+    - JULIAVERSION: "julianightlies/bin/winnt/x86/julia-latest-win32.exe"
+    - JULIAVERSION: "julianightlies/bin/winnt/x64/julia-latest-win64.exe"
 matrix:
   allow_failures:
-    - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
-    - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
+    - JULIAVERSION: "julialang/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
```
tkelman: what's wrong on win32?

pkofod: Two conjugate gradient tests fail. We've had this happen with osx as well. I can't point you to the exact line.

Tbh, I don't think I have access to a win32 installation... But maybe I can find a computer somewhere with one.

tkelman: you can run 32 bit julia on 64 bit windows

tkelman: or likewise 32 bit linux julia on 64 bit linux, if the problem is integer-literal-size-dependent

pkofod: > or likewise 32 bit linux julia on 64 bit linux, if the problem is integer-literal-size-dependent

I'll try 32 bit linux to see if the problem is there as well. When I said I don't have a 32bit windows, a more precise statement would have been, I don't have a windows installation

pkofod: great, I can reproduce locally, will investigate.

pkofod: The Hosaki tests do not converge but are improving up until iteration 1000, although very slowly.

pkofod: So I'm not at a computer but I think there's convergence in x. I would bet that increasing the number of iterations by 10 would cause win32 to converge

pkofod: https://ci.appveyor.com/project/pkofod/optim-jl/build/1.0.336/job/j5p6b1qm39rcnlfy#L105 seems to work after https://github.com/JuliaNLSolvers/Optim.jl/pull/423

tkelman: this is also undoing #410, should go back to the newer url's

PR #423, opened 2017-05-30
Fix win32 tests.

pkofod:


test/multivariate/solvers/first_order/cg.jl
```diff
@@ -1,7 +1,8 @@
  @testset "Conjugate Gradient" begin
 	# Test Optim.cg for all differentiable functions in Optim.UnconstrainedProblems.examples
 	run_optim_tests(ConjugateGradient(),
-                    convergence_exceptions = (("Powell", 1), ("Powell", 2), ("Polynomial", 1)))
+                    convergence_exceptions = (("Powell", 1), ("Powell", 2), ("Polynomial", 1)),
+                    f_increase_exceptions = (("Hosaki"),))
```
tkelman: strange that 32 bit would change behavior here? I wouldn't think integer size should make a difference?

pkofod: I don't think it is. The changes in `f` are very small towards the end, and for some reason the Hosaki objective has a tiny increase in `f` that stops the procedure without this option.

PR #424, opened 2017-05-30
Allow nightly to fail on Travis

anriseth:


pkofod:
good idea. I think the cholesky call can be compatted? (if it's added)

PR #425, opened 2017-06-02
Update appveyor.yml

pkofod:


pkofod:
> thanks, no need to re-tag for this since it's ci-only

sure thing, thanks for noticing!

PR #429, opened 2017-06-11
Add type parameter in OnceDifferentiable constructor with forward aut…

pkofod:
…odiff.


pkofod:
typo in test, should be fixed

Issue #248, opened 2016-07-18
Allow Dual Numbers


matthieugomez:
I'd like  `optimize` to work with dual numbers. This would be handy when optimizing a function that optimizes something. 
Would it be difficult to write all the algorithms in a way that work with dual numbers? At least for the case of derivative free univariate optimization, it seems like the only reason it does not work now is
 more type assertions than necessary.

KristofferC:
You need to be careful when differentiating iterative processes because the relationship between converging the residual and the derivative is not necessarily clear. For Newton methods I believe it should be ok though.

matthieugomez:
Could we make a list of methods that should be ok and try to make them work?
@jrevels @lbenet

mlubin:
It would be nice if this worked, but note there are better ways to differentiate through optimization: http://link.springer.com/chapter/10.1007%2F978-3-540-68942-3_7

pkofod:
> Could we make a list of methods that should be ok and try to make them work?
> @jrevels @lbenet

I've never had to do something like this, but I get why it would be neat. Can you post an example you would wish worked?

The only problem I can see is that we use stuff like explicit `axpy!` and other BLAS calls, often with a substantial performance gain. Some of this is can be dispatched away I guess, but this is not at the top of my personal todo-list. A PR is of course more than welcome.

matthieugomez:
In economics, we usually start by solving a model for a set of parameters. In a second step, we may want to find parameters so that the model solution approximates the data. So the code looks like a minimization of a minimization problem.

But, looking at the univariate code for, say, GoldenSection, I don't actually see how it would work with automatic differentiation. There is no way to use automatic differentiation to differentiate a binary search right? (i.e. the derivative of the minimizer is computed to be 0.0)

anriseth:
@matthieugomez for the HJB equations, have you considered using the methods described in https://cs.uwaterloo.ca/~paforsyt/hjb.pdf and https://arxiv.org/abs/1503.05864 ?
For example, the Piecewise Constant Policy Timestepping reduces the HJB equations to a system of linear PDEs that can be solved in parallel.

matthieugomez:
@anriseth wow that's pretty cool, thank you so much!. Did you have any experience with this? I have a public repository to work on solving HJBs in Julia (https://github.com/matthieugomez/HJBFiniteDifference.jl). I'm very interested in pull requests in case you have some code.

anriseth:
[...]

PR #430, opened 2017-06-14
Allow dualnumbers in univariate optimisation

anriseth:
Small step for #248


pkofod:
Yes,this is the use case I could imagine... Still waiting for @ranjanan  to get back to me :)

anriseth:
It's a bit more tricky for multivariate optimisation. There are issues with
the tracing that I couldn't figure out at the time of doing this pr.

Issue #399, opened 2017-04-21
Multidimensional arrays


pkofod:
Fix up (L-)BFGS NelderMead, and PSO code such that we can pass multidimensional arrays as input.

cortner:
How about making the input an abstract type? At the end of the day the state of the optimisation just needs to be an element of a vector space.

cortner:
Also see @ChrisRackauckas comments in #326. I really see no barrier to admitting a general abstract type as the state of optimisation - only advantages.

pkofod:
How abstract? `AbstractArray`?

cortner:
Not even an AbstractArray. You actually need elements of a Hilbert space, i.e., you need `x + y`, `t * x`, `dot(x, y)` and I think that is it already. 

This is really not a necessity, BUT: it might actually make the code more readable, and it could have some unexpected applications. E.g., if I minimise over a space of functions, then I don't have to represent functions in subsequent optimisation steps w.r.t. the same basis.

ChrisRackauckas:
>This is really not a necessity, BUT: it might actually make the code more readable, and it could have some unexpected applications. E.g., if I minimise over a space of functions, then I don't have to represent functions in subsequent optimisation steps w.r.t. the same basis.

Yes, ApproxFun would be a very interesting application which takes the generality to the max here.

>Not even an AbstractArray. You actually need elements of a Hilbert space, i.e., you need x + y, t * x, dot(x, y) and I think that is it already.

From BFGS, it looks like you need `A_mul_B!`, `scale!`, indexing, `*`, `+`, and `vecdot`. So if the indexing operation was replaced with a broadcast, then this all boils down to `*`, `+`, `vecdot`, and broadcast overloads for the type (guaranteed by indexing, but necessarily needing indexing). I believe `Fun` types may already satisfy this.

cortner:
everything in `BFGS`-type algorithms can in principle be reduced to the operations I mentioned - even if the current implementation possibly doesn't. The problem here is that that LBFGS stores the outer product `x * y'` instead of the operator it defines

ChrisRackauckas:
>everything in BFGS-type algorithms can in principle be reduced to the operations I mentioned - even if the current implementation possibly doesn't. 

[...]

PR #431, opened 2017-06-16
Fix input for (L)BFGS and PSO

pkofod:
ref #399

### DIFF
--- a/src/multivariate/solvers/first_order/bfgs.jl
+++ b/src/multivariate/solvers/first_order/bfgs.jl
@@ -30,7 +30,7 @@ type BFGSState{T,N,G}
     dg::Array{T,N}
     u::Array{T,N}
     invH::Matrix{T}
-    s::Array{T,N}
+    s::Vector{T}
     @add_linesearch_fields()
 end
 
@@ -47,7 +47,7 @@ function initial_state{T}(method::BFGS, options, d, initial_x::Array{T})
               similar(initial_x), # Store changes in gradient in state.dg
               similar(initial_x), # Buffer stored in state.u
               method.initial_invH(initial_x), # Store current invH in state.invH
-              similar(initial_x), # Store current search direction in state.s
+              vec(similar(initial_x)), # Store current search direction in state.s
               @initial_linesearch()...) # Maintain a cache for line search results in state.lsr
 end
 
@@ -57,7 +57,7 @@ function update_state!{T}(d, state::BFGSState{T}, method::BFGS)
 
     # Set the search direction
     # Search direction is the negative gradient divided by the approximate Hessian
-    A_mul_B!(state.s, state.invH, gradient(d))
+    A_mul_B!(state.s, state.invH, vec(gradient(d)))
     scale!(state.s, -1)
 
     # Maintain a record of the previous gradient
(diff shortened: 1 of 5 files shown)


src/multivariate/solvers/first_order/bfgs.jl
```diff
@@ -72,11 +72,10 @@ function update_state!{T}(d, state::BFGSState{T}, method::BFGS)
     copy!(state.x_previous, state.x)
 
     # Update current position
-    @simd for i in 1:n
-        @inbounds state.dx[i] = state.alpha * state.s[i]
-        @inbounds state.x[i] = state.x[i] + state.dx[i]
+    @simd for i = 1:n
+        state.dx[i] .= state.alpha .* state.s[i]
+        state.x[i] .= state.x[i] .+ state.dx[i]
```
anriseth: Will it not work to do `state.x .= state.x .+ state.dx`? (Would dot-notation not work for arbitrary vector-types?)

anriseth: Sorry, I got confused. I mean on the `state.s`: I thought that's meant to be a vector, whilst `state.x` may be a multidimensional array?

pkofod: This is a mistake! I had originally tried to change it to broadcast, but that didn't like mixing vectors and matrices. 

I mean, `state.x` is allowed to be multidimensional. If `x` is passed as a matrix we want `Optim.minimizer(res)` to be a matrix as well.

We could store `s` as the same type as `x`, and throw `vec`'s around `state.s` whereever it shows up?

anriseth:
Would reshape on state.s work?

src/multivariate/solvers/first_order/bfgs.jl
anriseth: Would reshape on state.s work?
Alternatively reshape on `Optim.minimizer(res)`, and then just have `state.x` as a vector internally

pkofod:
> Would reshape on state.s work?

That's what my `vec` suggestion does, so yes! and I'll do just that

src/multivariate/solvers/first_order/bfgs.jl
pkofod: yeah, ofc you could just do `minimizer = similar(initial_x); copy!(minimizer, state.x)` in the end, that's another possibility.

test/multivariate/array.jl
```diff
@@ -9,36 +9,37 @@
...
-                _minimizer = Optim.minimizer(res)
-                @test typeof(_minimizer) <: Array{Float64, 3}
-                @test size(_minimizer) == (2,2,1)
+        for m in (AcceleratedGradientDescent, BFGS, LBFGS, ConjugateGradient,  GradientDescent, MomentumGradientDescent, ParticleSwarm, SimulatedAnnealing)
+            res = optimize(f, g!, eye3, m())
+            _minimizer = Optim.minimizer(res)
+            @test typeof(_minimizer) <: Array{Float64, 3}
+            @test size(_minimizer) == (2,2,1)
+            if !(m in (SimulatedAnnealing, ParticleSwarm))
+                println(typeof(m))
```
pkofod: yes yes, this is debugging text, will remove... :)

PR #435, opened 2017-06-22
Manifold optimization

antoine-levitt:
Follow-up from https://github.com/JuliaNLSolvers/Optim.jl/issues/433. This implements the last option https://github.com/JuliaNLSolvers/Optim.jl/issues/433#issuecomment-309685983. Having a manifold and tangent vector point is a nice idea (and it works pretty well), but I got stuck at the typing of points and vectors (they are not arrays, nor abstractarrays, so what should Optim type them as?). Anyway this is not too bad, allows more fine-grained on when projections and retractions are done, and seems to work well. I fake an objective function for compatibility with line searches. Ideally, LineSearches should be redesigned to take as input a 1D function, but I'm not confortable enough with the API to do that.

This is a pretty basic implementation, and not optimal by any means: I just project the gradient on the tangent space, and retract the iterate after each update. There is possibly either too much or too little projections. A proper Riemannian optimization person would possibly be horrified (no fancy vector transport or anything like that), but it works well enough in my tests, and this PR implements the infrastructure to make additional modifications possible. Only the spherical and Stiefel manifolds are implemented right now, but adding other manifolds/different retractions should be very easy. In the future, it would probably be better to split into a separate Manifolds package, with only the Manifold and Flat types living in Optim (or NLSolversBase?).

I tried to make the modifications as unintrusive as possible compared to the unconstrained case. In the default case of a flat manifold, everything is a no-op, so there shouldn't be any change to the generated code.

There are some test failures on the counters, but they seem to be present on master already...


pkofod:
Thanks! Will come back and review at a later point

pkofod:
Very busy these days, sorry.. I haven't forgotten this, and I appreciate the effort!

antoine-levitt:
It's OK, I'm not in any hurry ;-)

antoine-levitt:
I've been using this in real life in a pretty involved example (minimizing a functional on collections of unitary matrices of different sizes), so functionally at least this works.

src/Manifolds.jl
```diff
@@ -0,0 +1,157 @@
...
+## * More manifolds from ROPTLIB
+## * {x, Ax = b}
+## * Intersection manifold (just do the projection on both manifolds iteratively and hope it converges)
+
+abstract type Manifold
+end
+
+
+type ManifoldObjective{T<:NLSolversBase.AbstractObjective} <: NLSolversBase.AbstractObjective
+    manifold :: Manifold
```
pkofod: the space around `::` seems quite unusual

src/Manifolds.jl
```diff
@@ -0,0 +1,157 @@
...
+project_tangent!(M::Flat, g, x) = g
+
+# {||x|| = 1}
+struct Sphere <: Manifold
+end
+retract!(S::Sphere, x) = normalize!(x)
+project_tangent!(S::Sphere,g,x) = (g .= g .- real(vecdot(x,g)).*x)
+
+# N x n matrices such that X'X = I
+# TODO: add more retractions, and support arbitrary inner product
```
pkofod: Any chance you would want to open an issue for this? Just so we don't lose track of the `TODO`

src/Manifolds.jl
```diff
@@ -0,0 +1,157 @@
...
+abstract type Stiefel <: Manifold end
```
pkofod: I know it's not done all over the code base, but a simple reference or explanation of what the "Stiefel manifold" is would be nice.

pkofod:
There are some conflicts you need to handle (the dot calls I believe) before this can be merged.

cortner:
will look at it asap.

src/Manifolds.jl
cortner: That's an important but pretty special manifold, no? What is the justification for having it as part of Optim?

pkofod:
> will look at it asap.

Thanks, I think I've made @antoine-levitt wait long enough :)

src/Manifolds.jl
antoine-levitt: Will document it. The main justification is that it is the one I need in my application ;-) More seriously, it's the basic manifold to do this kind of algorithms on: it was the original motivation for the theory, many other manifolds (sphere, O(n), U(n)) are special cases, it's probably the most used in applications (at least that I know of) outside of the sphere, and it's a good template for implementation of other manifolds.

There could be a Manifolds package living outside Optim, but it's a pretty short file so I would think this is fine, and people implementing other manifolds can just PR on Optim?

cortner: Good point about the special cases. Ok maybe leave this for now and discuss moving outside of Optim only when  somebody complains

src/Manifolds.jl
antoine-levitt: Done, see https://github.com/JuliaNLSolvers/Optim.jl/issues/448

docs/src/algo/complex.md
```diff
@@ -0,0 +1,8 @@
+# Complex optimization
+Optimization of functions defined on complex inputs is supported by simply passing a complex `x0` as input. All zeroth and first order optimization algorithms are supported.
+
+The gradient of a complex-to-real function is defined as the only vector `g` such that `f(x+h) = f(x) + real(g' * h) + O(h^2)`. This is sometimes written `g = df/d(z*) = df/d(re(z)) + i df/d(im(z))`.
+
+Because in general the gradient is not a holomorphic function of `z`, the Hessian is not a well-defined concept and second-order optimization algorithms are not applicable directly. To use second-order optimization, convert to real variables.
```
cortner: I don't understand this statement. Or rather, as I read it, it equally applies to the real case. I suggest to either give a little more detail?

docs/src/algo/complex.md
```diff
@@ -0,0 +1,8 @@
+# Complex optimization
+Optimization of functions defined on complex inputs is supported by simply passing a complex `x0` as input. All zeroth and first order optimization algorithms are supported.
+
+The gradient of a complex-to-real function is defined as the only vector `g` such that `f(x+h) = f(x) + real(g' * h) + O(h^2)`. This is sometimes written `g = df/d(z*) = df/d(re(z)) + i df/d(im(z))`.
+
+Because in general the gradient is not a holomorphic function of `z`, the Hessian is not a well-defined concept and second-order optimization algorithms are not applicable directly. To use second-order optimization, convert to real variables. 
+
+For first-order methods, preconditioners are assumed to be complex matrices, which is slightly restrictive for the reason mentioned above. If you need a more general operator, convert to real variables or create a custom type for P and overload `A_ldiv_B!(pg, P, g)` and `dot(x, P, y)`.
```
cortner: I don't understand this again: this is how it was always implemented?

docs/src/algo/manifolds.md
```diff
@@ -0,0 +1,37 @@
...
+
+manif = Optim.Sphere()
+Optim.optimize(f, g!, x0, Optim.ConjugateGradient(manifold=manif))
+```
+
+## Supported solvers and manifolds
+All first-order optimization methods are supported.
+
+The following manifolds are currently supported:
+* Flat: Euclidean space, does nothing
```
cortner: "does nothing" is maybe a bit informal. Maybe better "this is the standard and default case requiring no projection"

src/Manifolds.jl
```diff
@@ -0,0 +1,157 @@
...
+# The Geometry of Algorithms with Orthogonality Constraints
+# Alan Edelman, Tomás A. Arias, and Steven T. Smith
+# SIAM. J. Matrix Anal. & Appl., 20(2), 303–353. (51 pages)
+
+# Optimization Algorithms on Matrix Manifolds 
+# P.-A. Absil, R. Mahony, R. Sepulchre 
+# Princeton University Press, 2008
+
+
+abstract type Manifold
```
cortner: I suggest to add docstrings at least to all the new types

cortner:
@antoine-levitt what is the purpose of `real_to_complex`? I am guessing this just converts real. vectors to complex vector or no-op if you start with complex vectors and also no-op if the problem is real to begin with. But I didn't find the definition. Wherever it is defined, maybe add a docstring to it?

docs/src/algo/complex.md
antoine-levitt: What is not clear exactly in the statement above? The point of this paragraph is that the Hessian of a C^n -> R function is a 2n x 2n matrix, which does not map onto n x n complex matrices.

docs/src/algo/complex.md
antoine-levitt: You're right, it's more confusing than anything, will remove. It was just saying that passing a matrix P as a preconditionner is not the most general thing one can do (because of the 2n x 2n thing above), but it does not really matter that much.

antoine-levitt:
It's in the companion PR in NLSolversBase: https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/14

antoine-levitt:
@ChrisRackauckas moving the discussion here from https://github.com/JuliaNLSolvers/Optim.jl/pull/440, 
> This only works if you specify the gradient, right?
Yes, and it does crash when not passed a gradient. Finite differences is https://github.com/johnmyleswhite/Calculus.jl/issues/115, autodiff is https://github.com/JuliaDiff/ForwardDiff.jl/issues/157

docs/src/algo/complex.md
cortner: why can't the hessian of a C^n -> R function be a C^{n x n} matrix? Is the issue that the function is complex -> real rather than complex -> complex? Do you then mean by "in general" that it is *never* a holomorphic function of z or do you mean *sometimes* it is not a holomorphic function of z? I am now wondering whether I got tripped up by your usage of the word "in general".

antoine-levitt: It certainly *can* be a C^{n x n} matrix (e.g. |z|^2) but it can also not be, when the gradient (a C^n -> C^n function) is not holomorphic. Is the new formulation clearer?

docs/src/algo/complex.md
```diff
@@ -0,0 +1,6 @@
+# Complex optimization
+Optimization of functions defined on complex inputs (C^n to R) is supported by simply passing a complex `x0` as input. All zeroth and first order optimization algorithms are supported. For now, only explicit gradients are supported.
+
+The gradient of a complex-to-real function is defined as the only vector `g` such that `f(x+h) = f(x) + real(g' * h) + O(h^2)`. This is sometimes written `g = df/d(z*) = df/d(re(z)) + i df/d(im(z))`.
+
+The gradient of a C^n to R function is a C^n to C^n map. Even if it is differentiable when seen as a function of R^2n to R^2n, it might not be complex-differentiable. For instance, take f(z) = Re(z)^2. Then g(z) = 2 Re(z), which is not complex-differentiable (holomorphic). Therefore, the Hessian of a C^n to R function is not well-defined as a n x n complex matrix (only as a 2n x 2n real matrix), and therefore second-order optimization algorithms are not applicable directly. To use second-order optimization, convert to real variables.
```
cortner: This is clear to me now. But the sentence 

    Therefore, the Hessian of a C^n to R function is not well-defined as a n x n complex matrix (only as a 2n x 2n real matrix), and therefore second-order optimization algorithms are not applicable directly. To use second-order optimization, convert to real variables. 

still reads as if this is *always* true, but it is only true in this example (and many others). Maybe you could add the "rarely" somewhere? But I think my original point was valid: if g were complex-differentiable then you can define the hessian as a C^{n x n} matrix. You are just arguing that this is "rare"? Correct? But is it more "rare" than the objective f being complex-differentiable? 

I have no experience this this at all so will take your word - I just want the documentation to be clear. If you can add a reference that would help as well.

antoine-levitt: I added a "in general".

I don't know how rare it is. Some simple functions do have a C^{n x n} matrix, e.g. |z|^2, z' A z, others don't. I have encountered two cases where that happens: |z|^4 (the nonlinearity in the Gross-Pitaevskii equation) and Im(ln(z)) (used in the computation of Wannier functions).

I don't have a good reference for this. Even the definition of the gradient is somewhat non-standard (even though it's cleary the right thing to do for C^n to R optimization). Since half of quantum mechanics is minimizing complex-to-real energies (the other half being perturbation), the physics literature is littered with "dE/dz^*" with no details, so this might be covered in a physics textbook somewhere.

antoine-levitt:
Any remaining obstacles to merging this?

pkofod:
Not much I would say! Tests only fail because of the base PR not being merged, right?

antoine-levitt:
I think so, yes (at least they work locally)

cortner:
I have no reasons to delay merging

pkofod:
https://github.com/JuliaLang/METADATA.jl/pull/10875 yayay, we'll be there soon! Sorry for taking so long. It's really a cool addition. I have just been swamped with other stuff!

pkofod:
Just as a fyi, I'm not hesitating to merge, I'm just waiting for some metadata stuff to be sorted out for nlsolversbase...

anriseth:
It was not as easy as we hoped to tag a new NLSolversBase. It seems like we are getting closer, this PR is good to go: https://github.com/JuliaLang/METADATA.jl/pull/11123

Then remains (correct me if I'm wrong):
- Tagging NLSolversBase.jl 3.0
- Fixing merge conflicts here
- Tests pass && merge

pkofod:
@antoine-levitt I know this is not your fault, but there were some problems with getting NLSolversBase tagged, but there should be nothing stopping us now! Let's get this merged, do you have time to get it up to speed?

antoine-levitt:
I'm always confused with git and merges, but I think it's done. Let's see what CI has to say.

pkofod:
> edit: still doesn't import the correct NLSolversBase apparently?

yeah, I know.. I'm on it.. Will fix it later

pkofod:
Well well well, that only took three months! :) Thanks for the PR and your patience.

I'll just say this here, but it goes quite generally. If you have some interesting examples please share them. I'm collecting a bunch of "minitutorials" for Optim as Jupyter notebooks, and I think it would be cool to show this off on a/some nice example(s).

antoine-levitt:
Awesome, thanks!

The example I put in the tests is quite nice and self contained, it computes the first eigenvalues of a matrix by optimizing the Rayleigh quotient over orthogonality constraints. I have other examples but way more involved.

PR #437, opened 2017-06-22
Move assess_convergence and trace! methods from utilities to solvers

guimarqu:
Considering that the convergence assessment and the trace update are specific to each solver, I suggest to move those methods from utilities to solver files.

I noticed that a trace method was used for almost all 1st order solver, also a convergence assessment method. So I keep them in utilities with another name and they are called by the `trace!` or `assess_convergence` specific to the solver.

### DIFF
--- a/test/multivariate/solvers/second_order/newton.jl
+++ b/test/multivariate/solvers/second_order/newton.jl
@@ -15,7 +15,9 @@
     Optim.optimize(NonDifferentiable(f_1, initial_x), [0.0], Newton())
     Optim.optimize(OnceDifferentiable(f_1, g!_1, initial_x), [0.0], Newton())
 
-    results = Optim.optimize(f_1, g!_1, h!_1, [0.0], Newton())
+    options = Optim.Options(store_trace = true, show_trace = false,
+                            extended_trace = true)
+    results = Optim.optimize(f_1, g!_1, h!_1, [0.0], Newton(), options)
 
     @test_throws ErrorException Optim.x_trace(results)
     @test Optim.g_converged(results)
--- a/test/multivariate/solvers/second_order/newton.jl
+++ b/test/multivariate/solvers/second_order/newton.jl
@@ -18,8 +18,7 @@
     options = Optim.Options(store_trace = true, show_trace = false,
                             extended_trace = true)
     results = Optim.optimize(f_1, g!_1, h!_1, [0.0], Newton(), options)
-
-    @test_throws ErrorException Optim.x_trace(results)
+    #@test_throws ErrorException Optim.x_trace(results)
     @test Optim.g_converged(results)
     @test norm(Optim.minimizer(results) - [5.0]) < 0.01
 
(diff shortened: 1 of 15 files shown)


guimarqu:
I have modified a little bit the first call to `optimize` of `Newton` and `NewtonTrustRegion`. I added options to extend the trace and increase the covering.
I had to comment the `#@test_throws ErrorException Optim.x_trace(results)` because of this modification.

test/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -150,7 +150,9 @@ end
 
     d = TwiceDifferentiable(f, g!, h!, [0.0])
 
-    results = Optim.optimize(d, [0.0], NewtonTrustRegion())
+    options = Optim.Options(store_trace = false, show_trace = false,
+                            extended_trace = true)
+    results = Optim.optimize(d, [0.0], NewtonTrustRegion(), options)
```
pkofod: You're basically testing that it doesn't store the trace even if we tell it to do the extended trace, right? That makes sense. The old test should also still be there though, as the default options should show the same behavior.

test/multivariate/solvers/second_order/newton.jl
```diff
@@ -15,9 +15,10 @@
     Optim.optimize(NonDifferentiable(f_1, initial_x), [0.0], Newton())
     Optim.optimize(OnceDifferentiable(f_1, g!_1, initial_x), [0.0], Newton())
 
-    results = Optim.optimize(f_1, g!_1, h!_1, [0.0], Newton())
-
-    @test_throws ErrorException Optim.x_trace(results)
+    options = Optim.Options(store_trace = true, show_trace = false,
+                            extended_trace = true)
+    results = Optim.optimize(f_1, g!_1, h!_1, [0.0], Newton(), options)
+    #@test_throws ErrorException Optim.x_trace(results)
```
pkofod: How come you commented out this test?

guimarqu: Because I give to `optimize` the options `store_trace = true`and `extended_trace = true`. I just realized that I have only to set `store_trace = false` to have the ErrorException back.

test/multivariate/solvers/second_order/newton_trust_region.jl
guimarqu: So, I add the test to check if `x_trace` throws an `ErrorException`?

test/multivariate/solvers/second_order/newton.jl
pkofod: Oh, sure, but the test was there to check that `Optim.x_trace` actually fails if you didn't *ask* it for the trace :)

guimarqu: Yes and if I don't store the trace, it's like if I didn't ask for it, isn't it ?

pkofod:
Hi @guimarqu Just a quick remark! You don't have to name it `common_1order_trace!`, just call it `common_trace!`. The type annotation you have will take care of the `1order` part. Does that make sense? Just call it `common_trace!` and let dispatch handle the rest.

pkofod:
Thanks!

PR #455, opened 2017-08-22
Remove #uncomment for v0.8.0 stuff

anriseth:
These commented out codes seem superfluous?


pkofod:
Guess there should have been a note in `deprecate.jl`... thanks!

PR #462, opened 2017-09-03
[RFC] Add docstrings to ConjugateGradient, BFGS and the rest

pkofod:
Since new (and existing) users might not always be able to remember how to specify everything, we should have informative doc strings. I've added one for `ConjugateGradient` and one for `BFGS`. The output looks something like

conjugate

for the former and

bfgs

for the latter in a 80x43 char sized terminal. Feel free to make PRs against this branch to correct mistakes, suggest additional information and add docstrings for more methods.


johnmyleswhite:
LGTM

anriseth:
Sweet. That's very useful

pkofod:
I'll just merge this as is. Feel free to chip in with docs strings for missing algorithms ;)

PR #463, opened 2017-09-15
Fix most depwarns and errors on 0.7

yuyichao:


src/multivariate/solvers/zeroth_order/nelder_mead.jl
```diff
@@ -227,7 +227,7 @@ function update_state!(f::F, state::NelderMeadState{T}, method::NelderMead) wher
         else # f_reflect > f_highest
             # Inside constraction
             @simd for j in 1:n
-                @inbounds state.x_cache[j] = state.x_centroid[j] - γ *(state.x_reflect[j] - state.x_centroid[j])
+                @inbounds state.x_cache[j] = state.x_centroid[j] - state.γ *(state.x_reflect[j] - state.x_centroid[j])
```
yuyichao: This is a behavioral change but I think it's correct from the context...

pkofod: Good catch... I really wish these unicode irrationals had to be explicitly loaded.

pkofod:
I'll merge this when metadata is sorted out for nlsolversbase...

yuyichao:
Ah, that explains why this package haven't been updated for it.... I just pull from the latest master/tag and didn't pay attention to metadata....

pkofod:
> Ah, that explains why this package haven't been updated for it.... I just pull from the latest master/tag and didn't pay attention to metadata....

It's fine... We're not tagging before that is merged anyway, so.. But thanks for fixing these!

pkofod:
I'm not going to wait for mac tests on travis as they might not fire this year.... Thanks for doing this @yuyichao

PR #464, opened 2017-09-15
BUG: Fix printing for GoldenSearch.

sglyon:
the `metadata` doesn't have a `"best bound"` entry like the metadata for Brent does


pkofod:
Thanks

PR #473, opened 2017-10-16
Short explanation of the Rosenbrock function.

Wikunia:
There might be some people out there who are not familiar with the Rosenbrock function but are interested in this package ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ result = optimize(rosenbrock, zeros(2), BFGS())
 
 This minimizes the [Rosenbrock function](https://en.wikipedia.org/wiki/Rosenbrock_function) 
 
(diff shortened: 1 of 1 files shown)


pkofod:
Thanks for noticing our package, and thanks for this improvement of our readme! Hope you enjoyed using Optim so far. 

Yes it's a good idea to reference what we're doing, and show in math what the objective is. Sometimes I and others are so used to looking at the code and readme, that we don't really see that things like this can be helpful.

May I suggest you change the img-link to point to the image below?

formula

This way, the image should be accessible as long as github is. I'm not saying codecogs isn't going to stay accessible, but you never know!

Wikunia:
I just started to learn Julia and just creep around some packages and Optim seems quite interesting. 
Don't have a project for it atm but maybe try some stuff using it. Anyway thanks for the package. It looks nice :)

pkofod:
If it's hard to find out how to do something, the docs are probably lacking (and they are). If so, we do appreciate PRs :wink:

pkofod:
Thanks!

PR #474, opened 2017-10-18
Don't disallow sparse Hessians.

pkofod:
With https://github.com/JuliaNLSolvers/NLSolversBase.jl/commit/cee18b6f975a83c6974118a86b98b037660f2107 this allows for sparse hessians in `Newton()`. `NewtonTrustRegion()` I'm not so sure about currently. There's some type annotation things going on, but I also have to see if the code does something non-sparse-y things.

The solve step should obviously be different. I think we should go ahead and do the call overloading of a LinearSolver type for the `-H\g` calculation as in DiffEq.jl - though there were some issues that they had to work out (is it worked out? cc @ChrisRackauckas).

Just putting this up here. May merge it as it is and not advertise it, or may go all the way in this PR. At the very least we shouldn't be restricting things if there's no reason to.

I wonder if PositiveFactorizations could play nicely with abstract matrices @timholy ? Currently, it fails in `potrs!`

### DIFF
--- a/src/multivariate/solvers/second_order/newton.jl
+++ b/src/multivariate/solvers/second_order/newton.jl
@@ -28,12 +28,11 @@ end
 
 Base.summary(::Newton) = "Newton's Method"
 
-mutable struct NewtonState{T, N, F<:Base.LinAlg.Cholesky, Thd}
+mutable struct NewtonState{T, N, F<:Base.LinAlg.Cholesky}
     x::Array{T,N}
     x_previous::Array{T, N}
     f_x_previous::T
     F::F
-    Hd::Thd
     s::Array{T, N}
     @add_linesearch_fields()
 end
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I'll work out that issue tomorrow. It's very basic and just is about making sure the interface allows storing the factorization and other pieces at a level that is likely to not be reused in different threads. I got a big day today but tomorrow I'll work out exactly how the interface needs to change.

pkofod:
> I'll work out that issue tomorrow. It's very basic and just is about making sure the interface allows storing the factorization and other pieces at a level that is likely to not be reused in different threads. I got a big day today but tomorrow I'll work out exactly how the interface needs to change.

Yeah, no rush from our side, I was just interested in following the progress, so allowed myself to ping you :)

PR #475, opened 2017-10-18
Use dot notation

anriseth:
Newton is being fixed in #474


pkofod:
:+1: let's merge when travis finishes

anriseth:
I updated this because @pkofod did the same in #474 for Newton. I'm not sure how the axpy functionality works, but could it be that dot-notation can support a wider range of types?

pkofod:
> Out of curiosity: why was this changed?

Well, I think dots read a bit cleaner than level-1 blas calls. Do you think it matters for something? (performance, genericity, ...)

fredrikekre:
`state.alpha * state.s` will allocate a new vector. Should have been `state.alpha .* state.s` (?)

pkofod:
Yes, that should have been dotted as well.

PR #479, opened 2017-10-24
Add more dots

anriseth:
Thanks to https://github.com/JuliaNLSolvers/Optim.jl/pull/475#issuecomment-338956044

Will cause a merge conflict for #474 (I hope?)

### DIFF
--- a/src/multivariate/solvers/first_order/gradient_descent.jl
+++ b/src/multivariate/solvers/first_order/gradient_descent.jl
@@ -73,7 +73,7 @@ function update_state!(d, state::GradientDescentState{T}, method::GradientDescen
     # lssuccess = perform_linesearch!(state, method, d)
 
     # Update current position # x = x + alpha * s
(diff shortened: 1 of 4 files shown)


src/multivariate/solvers/first_order/gradient_descent.jl
```diff
@@ -73,7 +73,7 @@ function update_state!(d, state::GradientDescentState{T}, method::GradientDescen
     # lssuccess = perform_linesearch!(state, method, d)
 
     # Update current position # x = x + alpha * s
-    state.x .= state.x .+ state.alpha * state.s
+    state.x .= state.x .+ state.alpha .* state.s
```
ChrisRackauckas: I think it's less error-prone to do things like this one with `@.`

pkofod: It would have avoided my original mistake, so .. yeah :)

PR #480, opened 2017-10-25
Enable KrylovTrustRegion tests, and fix for new API

anriseth:
The tests for `KrylovTrustRegion` were not enabled in #416. I have enabled them, and fixed all the resulting errors.


anriseth:
That was a cheap way to increase code coverage :laughing:

PR #481, opened 2017-10-30
Pick increased f if allow_f_increases is true

anriseth:
I want an option for the optimizer to return the final values of `x` and `f(x)`, even if the last iteration results in `f_increased`.  

Let me know whether you think this is a good idea or not. I can also introduce an extra option `pick_increased_f`, instead of piggy-backing on `allow_f_increases`.

### DIFF
--- a/docs/src/user/config.md
+++ b/docs/src/user/config.md
@@ -44,7 +44,7 @@ In addition to the solver, you can alter the behavior of the Optim package by us
 * `f_calls_limit`: A soft upper limit on the number of objective calls. Defaults to `0` (unlimited).
 * `g_calls_limit`: A soft upper limit on the number of gradient calls. Defaults to `0` (unlimited).
 * `h_calls_limit`: A soft upper limit on the number of Hessian calls. Defaults to `0` (unlimited).
-* `allow_f_increases`: Allow steps that increase the objective value. Defaults to `false`. Take care when `allow_f_increases = true`. When the optimization stops a previous iterate may provide a lower objective value than the reported minimizer.
+* `allow_f_increases`: Allow steps that increase the objective value. Defaults to `false`. Note that, when setting this to `true`, the last iterate will be returned as the minimizer even if the objective increased.
 * `iterations`: How many iterations will run before the algorithm gives up? Defaults to `1_000`.
 * `store_trace`: Should a trace of the optimization algorithm's state be stored? Defaults to `false`.
 * `show_trace`: Should a trace of the optimization algorithm's state be shown on `STDOUT`? Defaults to `false`.
--- a/docs/src/user/config.md
+++ b/docs/src/user/config.md
@@ -44,7 +44,7 @@ In addition to the solver, you can alter the behavior of the Optim package by us
 * `f_calls_limit`: A soft upper limit on the number of objective calls. Defaults to `0` (unlimited).
 * `g_calls_limit`: A soft upper limit on the number of gradient calls. Defaults to `0` (unlimited).
 * `h_calls_limit`: A soft upper limit on the number of Hessian calls. Defaults to `0` (unlimited).
-* `allow_f_increases`: Allow steps that increase the objective value. Defaults to `false`.
+* `allow_f_increases`: Allow steps that increase the objective value. Defaults to `false`. Take care when `allow_f_increases = true`. When the optimization stops a previous iterate may provide a lower objective value than the reported minimizer.
 * `iterations`: How many iterations will run before the algorithm gives up? Defaults to `1_000`.
 * `store_trace`: Should a trace of the optimization algorithm's state be stored? Defaults to `false`.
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
What's the use case?

anriseth:
Preconditioning (I call optimize with iterations=1).

Steepest descent with fixed step length steps "too far", but I still want
to use this step and accelerate from it, or else the accelerator (N-GMRES)
stagnates.

anriseth:
> Is this what you always want though? I mean, say I want to allow f to increase because it might assist convergence, but I am actually looking for the minimizer. Is this then predictable behavior?

The only time `f_increased` is true at the final step is if you hit the maximum number of iterations. So the previous iterate is unlikely to be what you're looking for either. 
I'd say you "know what you're doing" when you `allow_f_increases`, to the extent that you also consider whether you've converged or just hit the maximum iterations. 

> I'm wondering if you could simply save the trace and take the last trace element instead?

This sounds a bit hack-y and reliant on a preconditioners trace functionality.

anriseth:
I can, as mentioned, just create an extra option for this functionality as well. In my opinion, however, it creates more clutter without preventing many, if any, actual problems as compared with the approach in this PR.

I will also make sure to note it in the documentation, wherever `allow_f_increases` is referenced.

pkofod:
Alright, I accept your argumentation. You basically say, that if you allow f to increase, you know that this might also happen on the very last iteration, correct?

I should be quite easy to generate a test case here. Just find an `f` such that `g!` will overstep the minimum and reach a higher objective value next iteration with a static step size, say
```
f(x) = 2*x[1]^2
g!(G, x) = copy!(G, 4*x[1])

using Optim
optimize(f, g!, [0.5], GradientDescent(linesearch=LineSearches.Static()), Optim.Options(allow_f_increases=true))
``` 
Then just test the behavior with various combinations of the options in the test you included.

src/types.jl
```diff
@@ -134,7 +134,7 @@ function Base.show(io::IO, r::MultivariateOptimizationResults)
         @printf io "     |f(x) - f(x')| / |f(x)| = %.2e \n" f_residual(r)
         @printf io "   * |g(x)| < %.1e: %s \n" g_tol(r) g_converged(r)
         @printf io "     |g(x)| = %.2e \n"  g_residual(r)
-        @printf io "   * stopped by an increasing objective: %s\n" f_increased(r)
+        @printf io "   * Stopped by an increasing objective: %s\n" (f_increased(r) && !iteration_limit_reached(r))
```
anriseth: I added this as the reason the optimization stopped is due to reaching the iteration limit, not necessarily because `f` increased.

anriseth:
> > I'm wondering if you could simply save the trace and take the last trace element instead?

> This sounds a bit hack-y and reliant on a preconditioner's trace functionality.

I am now convinced that there are use cases where one would like to use the preconditioner's trace. For example, if you only want to accelerate every `n`th step, you can do that by asking the preconditioner to run `n` iterations and then use the trace history. Such an approach is used by  [Siceur et al.](https://arxiv.org/abs/1606.04133), and Geletta and Cartis (not yet public). 

Using the trace functionality for this is a bit more wasteful in terms of memory allocation and complicates implementation, however, so I'm not sure if I want to go down that route within Optim, but rather leave that to an external package.

pkofod:
Just to be clear I'm fine with the idea here. I have to see your recent changes in a computer and then I'll approve (or not :o)

test/multivariate/f_increase.jl
```diff
@@ -0,0 +1,18 @@
+@testset "f increase behaviour" begin
+    f(x) = 2*x[1]^2
+    g!(G, x) = copy!(G, 4*x[1])
+
+
+    minimizers = [0.3, -1.5, 0.3, 0.5]
```
pkofod: I suspect these are  testing various combinations of the logical test? Could you comment which are which, and maybe test for that stopping criterion below?

docs/src/user/config.md
```diff
@@ -44,7 +44,7 @@ In addition to the solver, you can alter the behavior of the Optim package by us
 * `f_calls_limit`: A soft upper limit on the number of objective calls. Defaults to `0` (unlimited).
 * `g_calls_limit`: A soft upper limit on the number of gradient calls. Defaults to `0` (unlimited).
 * `h_calls_limit`: A soft upper limit on the number of Hessian calls. Defaults to `0` (unlimited).
-* `allow_f_increases`: Allow steps that increase the objective value. Defaults to `false`.
+* `allow_f_increases`: Allow steps that increase the objective value. Defaults to `false`. Take care when `allow_f_increases = true`. When the optimization stops a previous iterate may provide a lower objective value than the reported minimizer.
```
pkofod: This sentence is a bit awkward in my opinion. Could it simply be `Note, that when setting this to `true`, the last iterate will be returned as the minimizer even if the objective increased.` Or something like that.

PR #483, opened 2017-11-01
Fix preconditioning example

anriseth:
This example probably relied on an earlier version of ForwardDiff.


pkofod:
we should maybe doc test these

anriseth:
> we should maybe doc test these

What's a doc test? :p

PR #484, opened 2017-11-01
L-BFGS: Use scaled initial guess for inverse Hessian

anriseth:
If a preconditioner is not used, I propose that we default to the scaled initial guess for the inverse Hessian as given in Nocedal and Wright. (See 6.21 or 7.20 in the second edition).

I have also included a resetalpha setting for L-BFGS. It should probably be default for all line searches that `alphaguess!` sets `state.alpha=1`. None of the algorithms I know of uses the previous step length as an initial guess anyway. (Maybe for case of Momentum and AcceleratedGradientDescent; but they need a revamp of `alphaguess!` to work properly anyway)
And as the state is now, BackTracking will perform terribly, as it can only decrease step lengths.

### DIFF
--- a/src/multivariate/solvers/first_order/l_bfgs.jl
+++ b/src/multivariate/solvers/first_order/l_bfgs.jl
@@ -14,7 +14,7 @@ function twoloop!(s::Vector,
                   pseudo_iteration::Integer,
                   alpha::Vector,
                   q::Vector,
-                  scaleinvH::Bool,
+                  scaleinvH0::Bool,
                   precon,
                   devec_fun #all data is passed to this function is flat vectors, but precon might expect something different. this function undoes the flattening
                   )
@@ -42,7 +42,7 @@ function twoloop!(s::Vector,
     # Copy q into s for forward pass
     # apply preconditioner if precon != nothing
     # (Note: preconditioner update was done outside of this function)
-    if scaleinvH == true && typeof(precon) <: Void && upper > 0
+    if scaleinvH0 == true && upper > 0
         # Use the initial scaling guess if no preconditioner is used
         # See Nocedal & Wright (2nd ed), Equation (7.20)
         idx = mod1(upper, m)
@@ -77,7 +77,7 @@ struct LBFGS{T, IL, L, Tprep<:Union{Function, Void}} <: Optimizer
     P::T
     precondprep!::Tprep
     manifold::Manifold
-    scaleinvH::Bool
+    scaleinvH0::Bool
 end
 """
 # LBFGS
@@ -89,7 +89,7 @@ linesearch = LineSearches.HagerZhang(),
 P=nothing,
 precondprep = (P, x) -> nothing,
 manifold = Flat(),
-scaleinvH::Bool = true)
+scaleinvH0::Bool = true && (T <: Void)
 ```
 `LBFGS` has two special keywords; the memory length `m`,
 and the `scaleinvH0` flag.
@@ -118,8 +118,8 @@ function LBFGS(; m::Integer = 10,
                  P=nothing,
                  precondprep = (P, x) -> nothing,
                  manifold::Manifold=Flat(),
-                 scaleinvH::Bool = true)
-    LBFGS(Int(m), alphaguess, linesearch, P, precondprep, manifold, scaleinvH)
+                 scaleinvH0::Bool = true && (typeof(P) <: Void) )
+    LBFGS(Int(m), alphaguess, linesearch, P, precondprep, manifold, scaleinvH0)
 end
 
 Base.summary(::LBFGS) = "L-BFGS"
@@ -179,7 +179,7 @@ function update_state!(d, state::LBFGSState{T}, method::LBFGS) where T
     devec_fun(x) = real_to_complex(d,reshape(x, size(state.s)))
     twoloop!(vec(state.s), vec(gradient(d)), vec(state.rho), state.dx_history, state.dg_history,
              method.m, state.pseudo_iteration,
-             state.twoloop_alpha, vec(state.twoloop_q), method.scaleinvH, method.P, devec_fun)
+             state.twoloop_alpha, vec(state.twoloop_q), method.scaleinvH0, method.P, devec_fun)
     project_tangent!(method.manifold, real_to_complex(d,state.s), real_to_complex(d,state.x))
 
     # Save g value to prepare for update_g! call
--- a/src/multivariate/solvers/first_order/l_bfgs.jl
+++ b/src/multivariate/solvers/first_order/l_bfgs.jl
@@ -42,9 +42,16 @@ function twoloop!(s::Vector,
     # Copy q into s for forward pass
     # apply preconditioner if precon != nothing
     # (Note: preconditioner update was done outside of this function)
-    if scaleinvH0 == true && upper > 0
+    if scaleinvH0 == true && pseudo_iteration > 1
         # Use the initial scaling guess if no preconditioner is used
         # See Nocedal & Wright (2nd ed), Equation (7.20)
+
+        #=
+        pseudo_iteration > 1 prevents this scaling from happening
+        at the first iteration, but also at the first step after
+        a reset due to invH being non-positive definite (pseudo_iteration = 1).
+        TODO: Maybe we can still use the scaling as long as iteration > 1?
+        =#
         idx = mod1(upper, m)
         sidx = view(dx_history, :, idx) # TODO: should these use devec_fun?
         yidx = view(dg_history, :, idx) # TODO: should these use devec_fun?
--- a/src/multivariate/solvers/first_order/l_bfgs.jl
+++ b/src/multivariate/solvers/first_order/l_bfgs.jl
@@ -33,11 +33,11 @@ function twoloop!(s::Vector,
         if index < 1
             continue
         end
-        i = mod1(index, m)
-        # TODO: Will this vecdot be correct if x and q are complex?
-        @inbounds alpha[i] = rho[i] * vecdot(view(dx_history, :, i), q)
+        i   = mod1(index, m)
         dgi = view(dg_history, :, i)
-        q .-= alpha[i] .* dgi
+        dxi = view(dx_history, :, i)
+        @inbounds alpha[i] = rho[i] * vecdot(dxi, q)
+        @inbounds q .-= alpha[i] .* dgi
     end
 
     # Copy q into s for forward pass
@@ -53,10 +53,10 @@ function twoloop!(s::Vector,
         a reset due to invH being non-positive definite (pseudo_iteration = 1).
         TODO: Maybe we can still use the scaling as long as iteration > 1?
         =#
-        idx = mod1(upper, m)
-        sidx = view(dx_history, :, idx)
-        yidx = view(dg_history, :, idx)
-        scaling = dot(sidx, yidx) / sum(abs2, yidx)
+        i = mod1(upper, m)
+        dxi = view(dx_history, :, i)
+        dgi = view(dg_history, :, i)
+        scaling = dot(dxi, dgi) / sum(abs2, dgi)
         @. s = scaling*q
     else
         A_ldiv_B!(devec_fun(s), precon, devec_fun(q))
@@ -67,8 +67,9 @@ function twoloop!(s::Vector,
             continue
         end
         i = mod1(index, m)
-        @inbounds beta = rho[i] * vecdot(view(dg_history, :, i), s)
+        dgi = view(dg_history, :, i)
         dxi = view(dx_history, :, i)
+        @inbounds beta = rho[i] * vecdot(dgi, s)
         @inbounds s .+= dxi .* (alpha[i] - beta)
     end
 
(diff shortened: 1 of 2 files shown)


anriseth:
We should probably benchmark the scaling vs. no scaling. 
It probably has the biggest impact on BackTracking, whilst HagerZhang and MoreThuente seem to be living their own lives quite often.

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -34,26 +35,33 @@ function twoloop!(s::Vector,
...
-    A_ldiv_B!(devec_fun(s), precon, devec_fun(q))
-
+    if scaleinvH == true && typeof(precon) <: Void && upper > 0
+        # Use the initial scaling guess if no preconditioner is used
+        # See Nocedal & Wright (2nd ed), Equation (7.20)
+        idx = mod1(upper, m)
+        sidx = view(dx_history, :, idx) # TODO: should these use devec_fun?
+        yidx = view(dg_history, :, idx) # TODO: should these use devec_fun?
+        scaling = dot(sidx, yidx) / sum(abs2, yidx) # TODO: Should this be vecdot?
+        s .= scaling*q
```
pkofod: `s .= scaling.*q` to avoid allocating a temporary for `scaling*q`

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -79,15 +89,18 @@ LBFGS(; m::Integer = 10,
...
 snap2one = (0.75, Inf))
 ```
 `LBFGS` has three special keywords: memory length `m`, `extrapolate` flag,
 and `snap2one` tuple. The memory length determines how many previous Hessian
-approximations to store. The `extrapolate` can be set to `true` to enable an
+approximations to store. The `extrapolate` flag can be set to `true` to enable an
 extrapolation step used to determine the initial step-size for the line search
 step. The `snap2one` tuple provides boundaries to an interval in which the
 initial step-size should be mapped to the unit step-size.
+Note that `extrapolate=true` has higher priority than `resetalpha=true` (where resetalpha
```
pkofod: is "precedence"  a more precise word than "priority"?

pkofod:
Cool stuff . Curious  to see if it makes a difference!

src/multivariate/solvers/first_order/l_bfgs.jl
anriseth: Good catch :)

src/multivariate/solvers/first_order/l_bfgs.jl
anriseth: Yep

anriseth:
The documentation on L-BFGS will have to be changed when we implement https://github.com/JuliaNLSolvers/LineSearches.jl/pull/70, as the options related to alphaguess will be removed.

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -34,26 +35,33 @@ function twoloop!(s::Vector,
...
+        dgi = view(dg_history, :, i)
+        q .-= alpha[i] .* dgi
     end
 
     # Copy q into s for forward pass
     # apply preconditioner if precon != nothing
     # (Note: preconditioner update was done outside of this function)
-    A_ldiv_B!(devec_fun(s), precon, devec_fun(q))
-
+    if scaleinvH == true && typeof(precon) <: Void && upper > 0
```
cortner: I never studied this re-scaling. Do I understand correctly that this  possibly chooses a different scaling factor at each iteration?

What is the reason to not apply the scaling when you have a preconditioner?

anriseth: > Do I understand correctly that this possibly chooses a different scaling factor at each iteration?

Yes, it may change at each iteration. There's some discussion about it in N+W around eqns 6.21 and 7.20 (2nd ed). A thorough discussion about this, and other(?), scalings can be found in  [Gilbert and Lemaréchal - Some numerical experiments with variable-storage quasi-Newton algorithms](https://link.springer.com/article/10.1007%2FBF01589113?LI=true)

> What is the reason to not apply the scaling when you have a preconditioner?

If someone implements a preconditioner which is the same as this scaling, then you'd be applying it twice. I originally looked at implementing the scaling in the form of a preconditioner that uses `I`, but it was quite troublesome with the current API for updating preconditioners.

cortner: but suppose I have a preconditioned that gets the metric right, but not the scaling? Specifically, \kappa( P^{-1/2} H P^{-1/2} ) = O(1) but \sigma( P^{-1} H) \approx s with s far from 1. Then I can play some tricks to rescale it, but if LBFGS does it for me, then I'll be very happy! (As a matter of fact I have just such a situation.) 

So my point is, would it be possible to have scaling turned off by default if a preconditioner is used, but allow the user to turn it on? E.g., in the LBFGS type construction it could just be a kwarg that check whether a preconditioner is set or not.

anriseth:
Ah, yes good point. I'll make the change.

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -34,26 +35,40 @@ function twoloop!(s::Vector,
         end
         i = mod1(index, m)
         @inbounds alpha[i] = rho[i] * vecdot(view(dx_history, :, i), q)
-        @simd for j in 1:n
-            @inbounds q[j] -= alpha[i] * dg_history[j, i]
-        end
+        dgi = view(dg_history, :, i)
```
anriseth: Is a `view` useful at all?, or can I just do `q .-= alpha[i] .* dg_history[:,i]` ?

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -34,26 +35,40 @@ function twoloop!(s::Vector,
...
+        pseudo_iteration > 1 prevents this scaling from happening
+        at the first iteration, but also at the first step after
+        a reset due to invH being non-positive definite (pseudo_iteration = 1).
+        TODO: Maybe we can still use the scaling as long as iteration > 1?
+        =#
+        idx = mod1(upper, m)
+        sidx = view(dx_history, :, idx) # TODO: should these use devec_fun?
+        yidx = view(dg_history, :, idx) # TODO: should these use devec_fun?
+        scaling = dot(sidx, yidx) / sum(abs2, yidx) # TODO: Should this be vecdot?
+        @. s = scaling*q
```
anriseth: @antoine-levitt Do I need to do `devec_fun` on `s` and `q` here?

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -34,26 +35,40 @@ function twoloop!(s::Vector,
...
     for index in lower:1:upper
         if index < 1
             continue
         end
         i = mod1(index, m)
         @inbounds beta = rho[i] * vecdot(view(dg_history, :, i), s)
-        @simd for j in 1:n
-            @inbounds s[j] += dx_history[j, i] * (alpha[i] - beta)
-        end
+        dxi = view(dx_history, :, i)
```
anriseth: Same question on `view` as above.

src/multivariate/solvers/first_order/l_bfgs.jl
pkofod: since we're accessing memory in the right order here, I don't think this should be a problem

anriseth: So should I keep it as is, or go back to `q .-= alpha[i] .* dg_history[:,i]`

anriseth:
I've compared LBFGS with and without `scaleinvH0` for the different line searches, and overall it seems like `scaleinvH0=true` is the best default choice. It also seems like we can get some benefit from defaulting to BT3.

The below figure shows a scatter plot with different performance measures for LBFGS with combinations of (`scaleinvH0`,linesearch), with objective to reach a relative tolerance `f(x) <= f_min + 1e-10(f_0-f_min)`. There are 16 test problems, taken from my new branch on OptimTestProblems. The dots at the very top of each plot corresponds to failures.

figure_1

cortner:
This is very useful - thank you!

src/multivariate/solvers/first_order/l_bfgs.jl
pkofod: Its a bit difficult to see in my phone but I'll have a look later. Doubt this is going to be a bottleneck

anriseth:
I made some small updates. This is ready from my side.

antoine-levitt:
Just tested this on my problem because the benchmarks you have converge pretty quickly (also they have non-integer f/g calls?). I require pretty tight tolerances in my application, 1e-6 on the gradient
```
|--------+------------+--------+--------|
| method | scaleinvH0 | fcalls | gcalls |
|--------+------------+--------+--------|
| HZ     | true       |   1027 |   1027 |
| HZ     | false      |    866 |    866 |
| BT     | false      |   2103 |    628 |
| BT     | true       |    372 |    364 |
|--------+------------+--------+--------|
```
So slight degradation for HagerZhang, large improvement for backtracking. I'm happy to do more tests if you like.

src/multivariate/solvers/first_order/l_bfgs.jl
antoine-levitt: Yes, the vecdot is correct. update_state! and twoloop! only ever see incoming real vectors (the real_to_complex are there for communication with the rest of the world, e.g. the preconditioner that expects a complex array). You can merge this line with the next one? The view is correct in the sense that it will not allocate any new array (as opposed to without view). Speed wise, I'm not sure, but it's supposed to improve in the future so I'd say keep it.

src/multivariate/solvers/first_order/l_bfgs.jl
```diff
@@ -79,31 +96,38 @@ alphaguess = LineSearches.InitialStatic(),
 linesearch = LineSearches.HagerZhang(),
 P=nothing,
 precondprep = (P, x) -> nothing,
-manifold = Flat())
+manifold = Flat(),
+scaleinvH0::Bool = true && (T <: Void)
```
antoine-levitt: What is this T<:Void?

anriseth: Ah, good catch. It should be `P::T` above, i.e., we default to `true` if `P=nothing`.

anriseth:
> Just tested this on my problem because the benchmarks you have converge pretty quickly (also they have non-integer f/g calls?).

Thank you for running more tests :) It is definitely valuable. 
Non-integer values: My plots report performance ratios, which are functions of the number of f-calls, iterations, etc, to normalize across a collection of test problems. Here's a paper that explains the rationale of this approach; https://arxiv.org/pdf/cs/0102001.pdf

src/multivariate/solvers/first_order/l_bfgs.jl
antoine-levitt: That's not very intuitive, I'd expect T to be the ambiant scalar type. Why not `true && typeof(P) <: Void` as in the other place?

anriseth: Yes, that's better.

src/multivariate/solvers/first_order/l_bfgs.jl
anriseth: Great, thank you for the input. 

> You can merge this line with the next one?

True, however, I feel that it is more readable this way (I've changed them all to be consistent locally).

anriseth:
This is ready to merge from my side.

cortner:
I look forward to trying this out!

pkofod:
> Just tested this on my problem

what is "your problem" (ELI5) just for reference?

I'll try it out in a package of mine (the optimization part consists of some maximum likelihood problems)

pkofod:
Would love to hear back from all of you @anriseth @cortner @antoine-levitt if you do experiment with this. I'll try it out myself for sure.

antoine-levitt:
Essentially, minimizing a Dirichlet energy of a (complex, matrix-valued) function with some orthogonality constraints (so you see what those PRs were about!), the target application being the computation of Wannier functions in solids. The paper is in preparation so the code is not public yet but it should be pretty soon (and I can send it privately to anyone interested). The problem is not very well-conditioned (I'm not preconditioning, mainly because I'm lazy) and the problem is really the local convergence (I'm already starting in a good place, and just optimizing to really get to that local minimum to high precision). I use LBFGS with a lot of history (100). In particular for linesearches I would expect that the functional is very close to quadratic for most of the optimization.

anriseth:
> Would love to hear back from all of you @anriseth @cortner @antoine-levitt if you do experiment with this. I'll try it out myself for sure.

I've attached the performance profiles of the test problems described in [this paper](https://arxiv.org/abs/1710.05200). Using `scaleinvH0=true` helps for all the line searches.
figure_1

Tolerance is set to `\|g(x)\| <= 1e-10 \|g(x_0)\|`, and maxiter to 20,000.
The test set contains nonlinear least squares problems of dimensions 100 - 100,000. Each (problem,dimension) pair is run 100 times with randomized initial conditions. (1800 problem instances in total)

PR #485, opened 2017-11-02
Better univarate docs to avoid confusion

pkofod:
Some users seem a bit confused that you cannot use gradient descent or newton's method for univariate optimization, and of course you can. This hopefully helps a bit.


src/multivariate/optimize/optimize.jl
```diff
@@ -19,7 +19,7 @@ function optimize(d::D, initial_x::AbstractArray, method::M,
     initial_x = complex_to_real(d, initial_x)
 
     if length(initial_x) == 1 && typeof(method) <: NelderMead
-        error("Use optimize(f, scalar, scalar) for 1D problems")
+        error("You cannot use Nelder-Mead for univariate problems.")
```
anriseth: Maybe we could propose alternatives here?

PR #486, opened 2017-11-03
Use alphaguess functionality from LineSearches

### LINKED ISSUE #294: StepLength Extrapolation

anriseth:
Lets the user decide the initial step length. 
- Fixes #294 
- Needs https://github.com/JuliaNLSolvers/LineSearches.jl/pull/70

TODO: 
- [x] Documentation

This also makes some changes to default behaviour. I think we should do a quick benchmark to check whether they make things worse or not. **UPDATE: I'll leave the defaults untouched. We need a better benchmarking functionality to decide on default changes**
- [x] ConjugateGradient initial step length guess in the first iteration uses HZ#I0, instead of HZ#I12 (these are references to the HagerZhang paper)
- [x] Newton now defaults to BackTracking()
- [x] L-BFGS now defaults to BackTracking()
- [x] BFGS now defaults to BackTracking()


pkofod:
Just because I'm lazy, `BackTracking()` defaults to the cubic interpolation right?

anriseth:
It defaults to cubic, yes. We should probably do a benchmark in the future
to choose which order (or other linesearch) that's best as default.

pkofod:
> It defaults to cubic, yes. We should probably do a benchmark in the future to choose which order (or other linesearch) that's best as default.

I'm pretty sure I benchmarked backtracking at some point, and cubic was a bit better if I'm not mistaken (on some subset of cutest). Didn't save the results though...

pkofod:
Did you benchmark it since you reverted?

anriseth:
> Did you benchmark it since you reverted?

I ran some benchmarks, and there weren't any clear improvements in terms of the metric used in `OptimTests`. 
It is likely cheaper in terms of gradient evaluations (and potentially f-calls) to use BackTracking, however, we should have some clear rules on evaluating performance before changing defaults I'd say.

anriseth:
This is good to go now, we just need to tag LineSearches.  
It should not be making any breaking changes, or changes to defaults compared to current master.

anriseth:
This is good to go from my side. I propose this PR is merged before the others are, to make rebasing easier.

pkofod:
Is the line searches bump necessary? It's just that we can't just tag a patch if we change require and it should still work with an older version even if you will get warnings right?

anriseth:
The bump is necessary to allow for the alphaguess property in the solvers.
(E.g. alphaguess=InitialStatic()) Or have I misunderstood what you mean?

pkofod:
Alright then it's just my weird local version of optim, thought I was on the latest tag

anriseth:
The current master works with both the newest LineSearches and older versions, but this PR only works with the newest LineSearches :p

pkofod:
Thanks as always!

PR #488, opened 2017-11-09
Change convergence tests a bit

anriseth:
A few changes to testing and reporting convergence. It's not crucial to implement these, however, I think they make more sense than the current setup.

- I want to test `abs(f_x-f_x_previous) < tol*abs(f_x)`, rather than the problematic `abs(f_x-f_x_previous) / (abs(f_x)+eps)`.
This changes the f_residual and x_residual functions to the (imo) more properly named `f_abschange` and `x_abschange`.
- I also believe it's better to check for <= inequalities than <.
- There is much common functionality between the zeroth order utilities that I have put in its own file.


anriseth:
I realized there's a mistake to use f_abschange when printing the results.
I'll fix that soon.

pkofod:
I'll review this tomorrow, thanks

pkofod:
I'll try to find time tomorrow.

pkofod:
If you resolve the conflicts, we can merge it.

pkofod:
AV fails on the output tests.

PR #489, opened 2017-11-13
Allow callable types for callback

anriseth:
Currently, we can't use callable types as callback functions. This fixes that issue.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -1,5 +1,5 @@
 abstract type Optimizer end
-struct Options{T, TCallback <: Union{Void, Function}}
+struct Options{T, TCallback}
     x_tol::T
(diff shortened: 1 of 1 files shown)


pkofod:
I guess we don't dispatch on `Void`, but check for `nothing`?

anriseth:
> I guess we don't dispatch on Void, but check for nothing?

We check for `nothing` here https://github.com/JuliaNLSolvers/Optim.jl/blob/ea0bed508e4a5076c8c8b57865d831c2d802e3a1/src/utilities/update.jl#L19 Is that what you mean? :p

pkofod:
> Is that what you mean? :p

exactly

PR #492, opened 2017-11-28
Fix overwriting of initial guess

### LINKED ISSUE #491: Initial guess overwritten
```julia
using Optim
rosenbrock(x) =  (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
od = OnceDifferentiable(rosenbrock, zeros(2))
result = optimize(od, BFGS())

 * Algorithm: BFGS
 * Starting Point: [0.9999999926033423,0.9999999852005353]
 * Minimizer: [0.9999999926033423,0.9999999852005353]
```

pkofod:
Fixes #491 will add test

### DIFF
--- a/src/multivariate/optimize/interface.jl
+++ b/src/multivariate/optimize/interface.jl
@@ -40,15 +40,15 @@ promote_objtype(method::SecondOrderSolver, initial_x, obj_args...) = TwiceDiffer
 promote_objtype(method::SecondOrderSolver, initial_x, td::TwiceDifferentiable) = td
 
 # use objective.last_f_x, since this is guaranteed to be in Non-, Once-, and TwiceDifferentiable
-function optimize(objective::AbstractObjective, initial_x::AbstractArray = objective.last_x_f; kwargs...)
+function optimize(objective::AbstractObjective, initial_x::AbstractArray = copy(objective.last_x_f); kwargs...)
     method = fallback_method(objective)
     checked_kwargs, method = check_kwargs(kwargs, method)
     options = Options(; checked_kwargs...)
     optimize(objective, initial_x, method, options)
 end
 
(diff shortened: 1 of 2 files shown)


pkofod:
there seems to be some sort of nelder mead bug as well - oh well, will fix tomorrow

pkofod:
I know I added the `initial_x`-less methods, but maybe it's simply safer to always have people input a state. cc the bugfinder @antoine-levitt

antoine-levitt:
Ideally, the NDifferentiable constructors would not require initial_x, and this would go away. I think the fact that it does right now stems from the fact that the NDifferentiable structure conflates two things: 1) logic to accomodate finite differences / autodiff 2) a caching mechanism. A cleaner design would be to separate the two, so that NDifferentiable would not require initial_x, and the caching structure could be instanced by Optim, once the initial_x is known. I'm guessing that would be a non-trivial change, though.

pkofod:
> Ideally, the NDifferentiable constructors would not require initial_x, and this would go away. I think the fact that it does right now stems from the fact that the NDifferentiable structure conflates two things: 1) logic to accomodate finite differences / autodiff 2) a caching mechanism. A cleaner design would be to separate the two, so that NDifferentiable would not require initial_x, and the caching structure could be instanced by Optim, once the initial_x is known. I'm guessing that would be a non-trivial change, though.

Non-trivial yes, but it's one I've been experimenting with locally. I do like the idea, but it does cause some irritating things if you then want to provide the type for a sparse hessian (Jacobian in NLsolve). I have branches of Optim, NLsolve, NLSolversBase and Linesearches locally that has two types:

- NDifferentiable
- NDifferentiableCache

where the autodiff thing is just lazy, that is you provide it at the end of the NDifferentiables.

Right now I'm actually not really sure why I didn't finish it, but maybe I should. I do agree that it is the cause of this, and that it is annoying/ugly to supply `initial_x` twice.

PR #496, opened 2017-12-11
Use new NLSolversBase.

pkofod:
https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/27

This has some small implications for Optim (a new positional argument in a constructor) but it really isn't that big of a deal as it can be deprecated. This advantage here is less code in Optim that could just as easily be used in NLsolve. It will however also have an effect on the usage of `optimize`, as I intend to reintroduce the `autodiff` keyword and suggest that people use
```
optimize(f, g!, initial_x, method, options; autodiff = :central`
```
instead of
```
df = OnceDifferentiable(f, g!, initial_x; autodiff = :central)
optimize(df, initial_x, method, options)
```
Going forward `OnceDifferentiable` should only be used if you want to reuse the caches. There should be no "feature" advantage, and if there is it should only be for advanced users (say using a weird type for the gradient to leverage multiple dispatch to do weird things during optimization).

**Internal changes**
However, I don't think it's totally fair to say there are no changes to Optim other than what described above. I really want to remove the `update_g!` (that actually updates f as well) and `update_h!` methods and have them in the `update_x` (call it `update`) methods. I think the current thing where f and g are updated in the main loop makes it more difficult to read the algorithm, so let's not do that anymore. It was introduced before `value_gradient!` existed.


src/multivariate/optimize/optimize.jl
```diff
@@ -13,31 +16,34 @@ after_while!(d, state, method, options) = nothing
...
     stopped, stopped_by_callback, stopped_by_time_limit = false, false, false
     f_limit_reached, g_limit_reached, h_limit_reached = false, false, false
     x_converged, f_converged, f_increased = false, false, false
+
     g_converged = if typeof(method) <: NelderMead
         nmobjective(state.f_simplex, state.m, n) < options.g_tol
     elseif  typeof(method) <: ParticleSwarm || typeof(method) <: SimulatedAnnealing || typeof(method) <: KrylovTrustRegion
         # TODO: remove KrylovTrustRegion when TwiceDifferentiableHV is in NLSolversBase
-        g_converged = false
+        false
```
anriseth: I think `typeof(method) <: KrylovTrustRegion` can be removed now.

pkofod: ah, thought I had found all your todos :) thanks

PR #501, opened 2017-12-25
Use abstract type instead of typealias for solver distinction.

pkofod:
@anriseth @jonathanBieler

Are there other things you found annoying while trying to base your stuff off of Optim? I'll be happy to generalize where needed.

### DIFF
--- a/src/deprecate.jl
+++ b/src/deprecate.jl
@@ -63,6 +63,6 @@ end
 function optimize(df::OnceDifferentiable,
     l::Array{T},
     u::Array{T},
-    F::Fminbox{O}; kwargs...) where {T<:AbstractFloat,O<:Optimizer}
+    F::Fminbox{O}; kwargs...) where {T<:AbstractFloat,O<:AbstractOptimizer}
     throw(ErrorException("Optimizing an objective `obj` without providing an initial `x` has been deprecated without backwards compatability. Please explicitly provide an `x`: `optimize(obj, x, l, u, method, options)``"))
 end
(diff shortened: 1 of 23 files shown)


anriseth:
Thanks :)

One thing that can be convenient to access from outside of Optim is multiple dispatch on functions such as `default_convergence_assessment`: https://github.com/JuliaNLSolvers/Optim.jl/blob/d55a8a443abed8cbca2a0b8aa38fe2bc148bf8ac/src/utilities/assess_convergence.jl#L47 
and (this one is my fault :p ) https://github.com/JuliaNLSolvers/Optim.jl/blob/d55a8a443abed8cbca2a0b8aa38fe2bc148bf8ac/src/multivariate/solvers/zeroth_order/zeroth_utils.jl#L1

pkofod:
I'll fix the 0th order states and the default convergence. Maybe I should try to implement something outside of Optim again (or update constrainedoptim.jl) to get a feel myself.

anriseth:
There might be something connected with default tracing functions as well,
however, new algorithms probably have special info that they should be
providing via trace.

pkofod:
> There might be something connected with default tracing functions as well,
> however, new algorithms probably have special info that they should be
> providing via trace.

Yes I think they'll probably by "special" in some way, with some specific things to add to the extended trace. I do think that our trace/convergence/options stuff is far too hardcoded but we can fix that at a later stage.

pkofod:
@anriseth @jonathanBieler I hope this removes some of the annoyances. I will tag a new version soon.

jonathanBieler:
Thanks, I'll try this in a few days.

PR #505, opened 2018-01-13
Acceleration / nonlinear preconditioning with N-GMRES and O-ACCEL

### LINKED ISSUE #477: Nonlinear preconditioning / acceleration algorithms
I'm interested in nonlinear (right) preconditioning for nonlinear systems and optimization, also referred to as acceleration. A discussion on how to support this would be interesting. PETSc supports this for nonlinear systems, which allows you to mix and match solvers in order to address particular features of a problem in a flexible way. ([Brune et al. Composing Scalable Nonlinear Algebraic Solvers](https://arxiv.org/abs/1607.04254)). For existing literature on optimization acceleration, see https://arxiv.org/abs/1606.04133 and https://arxiv.org/abs/1106.4426.

In particular, I'm interested in implementing Nonlinear GMRES and my own variant, as they can just be added on top of any of the existing solvers we already have -- often with great benefit to high-dimensional problems. See a manuscript I recently wrote: [Objective acceleration for unconstrained optimization](https://arxiv.org/abs/1710.05200) (@cortner you mentioned that you'd be interested to have a look)

So the design and implementation options for nonlinear preconditioning are:
- Be solver specific: i.e. create preconditioning only for a `NGMRES`. For my own testing purposes, this would suffice.
- Enable preconditioning for all solvers: i.e. in the `optimize` function. This is more complicated, however, gives more flexibility for people to experiment with new and interesting optimization approaches.

I can also add that @antoine-levitt implemented the Nonlinear GMRES idea for `NLsolve`, in form of Anderson acceleration, as requested by @ChrisRackauckas. So taking the approach of enabling preconditioning for all solvers is equally relevant for `NLsolve`.

anriseth:
This implements solvers that do two things:
- Show an example of nonlinear preconditioning.
- Introduce two nonlinear preconditioning algorithms that significantly improves the performance of Gradient Descent.

This implements two acceleration algorithms, [Nonlinear GMRES](http://www.hansdesterck.net/Publications-by-topic/nonlinear-preconditioning-for-nonlinear-optimization) and ["Objective Acceleration"](https://arxiv.org/abs/1710.05200).
I have tested them as accelerators on top of Gradient Descent and L-BFGS. For GD it is very efficient, for L-BFGS I am not sure if it has value ( the situations where the accelerators improve L-BFGS, it can be better to combine them with GD instead ).

Give me feedback on 
- whether you think these belong in Optim.jl, or should be a separate package (@pkofod ?);
- whether this approach to nonlinear preconditioning is useful or not (@ChrisRackauckas ?).

TODO:
- [x] Make the algorithms work with complex numbers (help @antoine-levitt ?)
- [x] See if it is reasonable to support Manifolds (help @antoine-levitt ? )
- [x] Add documentation.
- [x] Add tests.
- [x] Add specialised tests
- [x] Get https://github.com/JuliaLang/METADATA.jl/pull/12920 merged 
- [x] Add issue for improving the tracing
- [x] Add issue: use QR updating for system matrix 

If it works well and people are happy, this fixes #477 .

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.6.0
 PositiveFactorizations
 Compat 0.18.0
-LineSearches 3.2.0
+LineSearches 3.2.4
 Calculus
 NLSolversBase 4.0.0
 ForwardDiff 0.5.0
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -42,6 +42,7 @@ pages:
                 - 'Conjugate Gradient': 'algo/cg.md'
                 - 'Gradient Descent': 'algo/gradientdescent.md'
                 - '(L-)BFGS': 'algo/lbfgs.md'
+                - 'Acceleration': 'algo/ngmres.md'
             - Hessian Required:
                 - Newton: 'algo/newton.md'
                 - Newton with Trust Region: 'algo/newton_trust_region.md'
--- a/docs/src/algo/ngmres.md
+++ b/docs/src/algo/ngmres.md
@@ -42,6 +42,8 @@ GMRES for linear problems.
 Application of the algorithm to optimization is covered, for example, in [2].
 A description of O-ACCEL and its connection to N-GMRES can be found in [3].
 
+*We recommend trying [LBFGS](lbfgs.md) on your problem before N-GMRES or O-ACCEL. All three algorithms have similar computational cost and memory requirements, however, L-BFGS is more efficient for many problems.*
+
 ## Example
 
 This example shows how to accelerate `GradientDescent` on the Extended Rosenbrock problem.
--- a/docs/src/algo/ngmres.md
+++ b/docs/src/algo/ngmres.md
@@ -1,0 +1,162 @@
+# Acceleration methods: N-GMRES and O-ACCEL
+## Constructors
+```julia
+NGMRES(;
+        alphaguess = LineSearches.InitialStatic(),
+        linesearch = LineSearches.HagerZhang(),
+        manifold = Flat(),
+        wmax::Int = 10,
+        ϵ0 = 1e-12,
+        nlprecon = GradientDescent(
+            alphaguess = LineSearches.InitialPrevious(),
+            linesearch = LineSearches.Static(alpha=1e-4,scaled=true),
+            manifold = manifold),
+        nlpreconopts = Options(iterations = 1, allow_f_increases = true),
+      )
+```
+
+```julia
+OACCEL(;manifold::Manifold = Flat(),
+       alphaguess = LineSearches.InitialStatic(),
+       linesearch = LineSearches.HagerZhang(),
+       nlprecon = GradientDescent(
+           alphaguess = LineSearches.InitialPrevious(),
+           linesearch = LineSearches.Static(alpha=1e-4,scaled=true),
+           manifold = manifold),
+       nlpreconopts = Options(iterations = 1, allow_f_increases = true),
+       ϵ0 = 1e-12,
+       wmax::Int = 10)
+```
+
+## Description
+These algorithms take a step given by the nonlinear preconditioner `nlprecon`
+and proposes an accelerated step on a subspace spanned by the previous
+`wmax` iterates.
+
+- N-GMRES accelerates based on a minimization of an approximation to the $\ell_2$ norm of the
+gradient.
+- O-ACCEL accelerates based on a minimization of a n approximation to the objective.
+
+N-GMRES was originally developed for solving nonlinear systems [1], and reduces to
+GMRES for linear problems.
+Application of the algorithm to optimization is covered, for example, in [2].
+A description of O-ACCEL and its connection to N-GMRES can be found in [3].
+
+## Example
+
+This example shows how to accelerate `GradientDescent` on the Extended Rosenbrock problem.
+First, we try to optimize using `GradientDescent`.
+
+```julia
+using Optim, OptimTestProblems
+UP = OptimTestProblems.UnconstrainedProblems
+prob = UP.examples["Extended Rosenbrock"]
+optimize(UP.objective(prob), UP.gradient(prob), prob.initial_x, GradientDescent())
+```
+The algorithm does not converge within 1000 iterations.
+
+```
+Results of Optimization Algorithm
+ * Algorithm: Gradient Descent
+ * Starting Point: [-1.2,1.0, ...]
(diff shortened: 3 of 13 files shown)


anriseth:
I ran some tests on a set of nonlinear least squares problems with 100 randomized initial guesses (described in [my paper](https://arxiv.org/abs/1710.05200) ), and on all available unconstrained CUTEst problems of size 20 to 100,000 from the standard initial guess.

The tests are available at https://github.com/anriseth/AccelerationBenchmark.jl/tree/master/benchmark/preconditioning 

The plots below are all performance profiles. The stopping criterion was a relative decrease in the gradient norm of `1e-8`. An optimization run is considered a failure if it takes more than 60 seconds (soft limit), or takes more than 10,000 iterations. These criterions seem to have been quite strict for some of the CUTEst problems, as even L-BFGS only solves about 70%. (We should probably look properly into this in the future). 
HagerZhang is used as linesearch.
(The `-10` behind N-GMRES and O-ACCEL means that we use a history of 10 previous iterates, the same as the default for L-BFGS).

**Nonlinear Least Squares problems, performance profiles for a selection of solvers.** 
perfprof_nls_all
perfprof_nls_oaccel_ngmres
perfprof_nls_oaccel_gd

** CUTEst problems ** 
perfprof_cutest_all
perfprof_cutest_oaccel_ngmres
perfprof_cutest_oaccel_gd

antoine-levitt:
Complex numbers: should be more or less transparent to the solver. The only places you need to worry about is when interacting with the outside world: preconditionners (in the linear sense) and manifolds. See how it's done in other solvers. Here probably nothing needs to be done.

Manifolds: call project_gradient after computing gradients, and call retract when moving around the manifold (ie proposing a new x). manifolds.jl in tests has an example using both manifolds and complex numbers.

About the PR itself: it looks great, thanks for doing it!

I don't like the terminology "preconditioning" for these since it clashes uncomfortably with the established use of linear preconditioning and might be confusing to users.

I don't see theoretically how accelerating anything else than a fixed-point method (ie gradient descent with fixed step) could be useful, but these things are weird. I'd be very interested in a direct comparison of LBFGS with these methods.

antoine-levitt:
Ah sorry, you're too fast ;-) Interesting results! It seems to be more robust than LBFGS for hard problems, do you know why that is?

The algorithm uses a slightly suboptimal method for solving the least squares problem (forming the Gram/covariance/overlap matrix). Haven't looked into the details but probably it can be converted to a QR factorization (like it's done in anderson acceleration in NLSolve). Optimally it should use QR updating (which is waiting on https://github.com/JuliaLang/julia/issues/20791 in NLSolve). It probably does not make a huge difference, but worth pointing out in a TODO item in case someone gets bored on a long winter night ;-)

src/utilities/trace.jl
```diff
@@ -1,6 +1,6 @@
 # First order methods trace, used by AcceleratedGradientDescent,
 # ConjugateGradient, GradientDescent, LBFGS and MomentumGradientDescent
-function common_trace!(tr, d, state, iteration, method::Union{LBFGS, AcceleratedGradientDescent, GradientDescent, MomentumGradientDescent, ConjugateGradient}, options)
+function common_trace!(tr, d, state, iteration, method::Union{LBFGS, AcceleratedGradientDescent, GradientDescent, MomentumGradientDescent, ConjugateGradient, AbstractNGMRES}, options)
```
pkofod: Should this not just be `method::FirstOrderOptimizer` according to the new setup? I think I missed this doing the other PR.

anriseth:
> Interesting results! It seems to be more robust than LBFGS for hard problems, do you know why that is?

The problem is in the `time_limit`. L-BFGS is struggling for one particular problem type, so it hits the time limit before converging. So it is still very robust, but shows as a failure in my plots due to this. The same is true for Gradient Descent in the NLS problems, and also for all the solvers in the CUTEst test set (some of those problems are really expensive). 

> The algorithm uses a slightly suboptimal method for solving the least squares problem (forming the Gram/covariance/overlap matrix).

Yes, that's a good idea. It is particularly obvious when comparing the timings of `O-ACCEL-20` to `O-ACCEL-10`, and similarly for N-GMRES, that the larger window size is the most time-consuming element for cheap-to-evaluate objectives. 
I agree that should be added as a future issue / TODO :)

> Complex numbers / manifolds

Thanks, I'll look into that

> I don't like the terminology "preconditioning" 

I think I should update all comments and variables to something along the lines of "nonlinear preconditioner". Or do you have a better terminology? I use nonlinear preconditioner because that is an established term in the communities I work in.

antoine-levitt:
I like "extrapolation",  "acceleration" or "mixing". Nonlinear preconditioning seems to refer to something else in domain decomposition methods.

anriseth:
@antoine-levitt are you talking about what to call the optimizer that is called before N-GMRES (default Gradient Descent)? Or more generally what to refer to the approach of combining two optimizers in the fashion I have done here?

> I like "extrapolation", "acceleration" or "mixing". 

There are two sides here. N-GMRES is doing the "acceleration" or "extrapolation", whilst (e.g.) GradientDescent is the nonlinear preconditioner. If you say "mixing", it's not clear whether you refer to GD or N-GMRES, and it also becomes quite similar to "composition", which is another way to combine nonlinear solvers (maybe not optimizers?). I have chosen to refer to N-GMRES as an accelerator, and whichever optimizer you put inside it as a nonlinear (right) preconditioner. 

> Nonlinear preconditioning seems to refer to something else in domain decomposition methods.

The DD nonlinear preconditioning I could find online all seem to use the following approach: Take a DD iteration (either full cycle, or on one level), and then apply (e.g.) Newton - repeat. In which case, the DD is a nonlinear (right) preconditioner for Newton. 

For reference, y thinking about these topics come from http://www.mcs.anl.gov/papers/P2010-0112.pdf and http://www.hansdesterck.net/Publications-by-topic/nonlinear-preconditioning-for-nonlinear-optimization

antoine-levitt:
I tend to think of these methods as having a "base" iteration, and an acceleration method on top (my thinking comes from domains like electronic structure where there is a natural fixed-point iteration). Your terminology makes sense as well - these methods have been rediscovered 100 times under different names and with different interpretations, so this is as good as any other.

anriseth:
This is ready to merge now. Any other details that people want to address before merging?

antoine-levitt:
Looks good! I would write somewhere in the docs that it has about the same memory/computing requirements as LBFGS and is often less efficient, so users go to LBFGS before trying this.

pkofod:
Good stuff @anriseth ! Thanks

PR #507, opened 2018-01-14
Use OptimTestProblems

anriseth:
This uses OptimTestProblems for testing.

Some notes:
- I haven't thoroughly checked why solvers don't converge on problems, so we should go through that on all the `skip` and `*_exceptions` problems in `run_optim_problems`
- None of the solvers deals with "Trigonometric", as they all converge to a local minimum
- In `run_optim_problems`, I now allow the TwiceDifferentiable solvers to run the problems that only provide gradients via finite differencing

Fixes https://github.com/JuliaNLSolvers/OptimTestProblems.jl/issues/3


anriseth:
We need https://github.com/JuliaLang/METADATA.jl/pull/12865 for tests to work.

pkofod:
Wonderful, will be good to have these out of Optim!

ChrisRackauckas:
Re-run

pkofod:
momentum is acting up

anriseth:
(L-)BFGS failed due to `f_increase`. Windows is behaving differently, so @pkofod how would you like to deal with that? Should we just skip all problems that fail on Windows (but not Linux) for now?

pkofod:
Yes that is fine

anriseth:
Ready to merge. The Travis MacOS queue has a many-hour long backlog, so we should probably wait and see that things pass there as well.

ChrisRackauckas:
Travis Mac is just down. I wouldn't wait.

anriseth:
Mac didn't fail us :champagne:

pkofod:
Great stuff

PR #508, opened 2018-01-17
Add gradient!! et al. to Manifold objectives

anriseth:
The newest version of NLSolversBase includes functions such as `value_gradient!!` etc. It seems like we should overload these for Manifolds as well.

### DIFF
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -8,8 +8,8 @@
 # Alan Edelman, Tomás A. Arias, and Steven T. Smith
 # SIAM. J. Matrix Anal. & Appl., 20(2), 303–353. (51 pages)
 
-# Optimization Algorithms on Matrix Manifolds 
-# P.-A. Absil, R. Mahony, R. Sepulchre 
+# Optimization Algorithms on Matrix Manifolds
+# P.-A. Absil, R. Mahony, R. Sepulchre
 # Princeton University Press, 2008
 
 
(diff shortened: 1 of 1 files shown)


anriseth:
This commit is also in #505. So we could just close this one, or would you like to keep it separate for commit history reasons @pkofod ?

pkofod:
I'm fine with merging this separately. Would be lovely with a few tests that make sure that it's working as intended.

anriseth:
After looking a bit more through the Optim code and tests, it seems to me like the type
`ManifoldObjective` and the corresponding `value`, `gradient` etc. methods are only used in the context of line searches, and not for the actual algorithms. All the first-order solvers passes `ManifoldObjective(df)` to `perform_linesearch!` (See e.g. https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/multivariate/solvers/first_order/gradient_descent.jl#L74 )

@antoine-levitt is this correct?

antoine-levitt:
LGTM. The point of this is to trick the linesearches into projecting down on the manifold, ie evaluating f(p(x+hd))  instead of f(x+hd). The objective function should only ever see points on the manifold, which x+hd is not. In the optimization algorithm, this is done by manually calling retract. This is a bit of a hack, but it works ;) the proper solution is to redefine the linesearch API to work on functions of a single variable, but I was too lazy to do that.

anriseth:
Okay, so for the current usage, we don't actually need to dispatch the `!!` functions for Manifolds. They are only called by Optim in the `initial_state` methods, whilst ManifoldObjective is only used from within `perform_linesearch!`.

We don't need this PR, however, I think we should update the code for manifolds in two ways:
1. Consider calling `ManifoldObjective(df)` when doing `state.linesearch` within `perform_linesearch!`. 
2. Think of tests that can catch that manifold objectives are called correctly from a solver ( especially relevant for new solvers )

I'll create an issue.

pkofod:
Good point, it's just a force evaluations function, so it shouldn't be called during a procedure in most cases

PR #509, opened 2018-01-18
Be less strict in the Options constructor

pkofod:
This is inspired by the annoying way you have to specify options for bigfloats, see https://discourse.julialang.org/t/optimization-routines-for-bigfloat-doubledouble-higher-precision-objective-and-gradient-functions/8448/3?u=pkofod

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -21,7 +21,7 @@ struct Options{T, TCallback}
     extended_trace::Bool
     show_every::Int
     callback::TCallback
-    time_limit::T
+    time_limit::Float64
 end
 
 function Options(;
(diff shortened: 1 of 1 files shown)


src/types.jl
```diff
@@ -21,7 +21,7 @@ struct Options{T, TCallback}
     extended_trace::Bool
     show_every::Int
     callback::TCallback
-    time_limit::T
+    time_limit::Float64
```
anriseth: Does this work for all systems? Can you do typeof(time()) instead?

pkofod: What would cause it to not work? I can make it a separate type parameter, sure.

anriseth:
I was just wondering if all systems support Float64 ?

ChrisRackauckas:
Yes all systems do.

anriseth:
> Yes all systems do.

Okay, so as long as `time()` returns a `Float64` all the time there's no issue.

pkofod:
It seems to be hardcoded as a double here, so https://github.com/JuliaLang/julia/blob/a627e51e17de8f7cb16394604147a84e99d24473/src/support/timefuncs.h#L16

PR #511, opened 2018-01-18
const on OptimizationTrace typealias

pkofod:


pkofod:
Oh, that should be an easy fix, as the part that breaks that is not really important for what I wanted. It's a problem that it was not tested though

PR #518, opened 2018-01-23
DiffeqDiffTools derivatives

### LINKED ISSUE #476: Move from Calculus.jl to DiffEqDiffTools.jl
DiffEqDiffTools supports much faster finite differencing routines than Calculus. 
Ref: discussion on gitter.

anriseth:
This tries to use the DiffeqDiffTools finite difference derivatives rather than Calculus.

I made a comparison between the running times using Calculus and this PR, available here:
https://gist.github.com/anriseth/ccc51401525b923fbdc8147b2c8ed36f 
**Summary**: 
- DiffEqDiffTools is much faster at approximating the Hessian if it has access to the gradient
- DiffEqDiffTools is much worse if we approximate the Hessian using the Jacobian of a finite-difference gradient (People should not do this anyway?)
- DiffEqDiffTools is slightly slower at approximating gradients than Calculus. *Could be because the problems tested are too cheap, so the cache allocation is more costly?*

**Preliminary conclusion:**
If this enables FD gradients for complex number optimization, then the little extra cost in gradient evaluation is worth it. We need to think about how to deal with Hessian when only `f` is supplied, however.

**TODOs**:
- [x] Test gradients for complex number optimization
- [x] Update docs on complex number optimization and say we support finite difference gradients.
- [x] Update REQUIRE version DiffEqDiffTools
- [x] Create an issue reminding us to update to DiffEqDiffTools Hessians when https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/issues/13 is closed.
- [x] ~Add a `warn_once` for Newton that it may be faster to use LBFGS instead of Finite Differences? (Alternatively, mention it in the docs)~
- [x] ~Remove all Calculus remnants~
- [x] ~Add `GradientCache` and `JacobianCache` as parameter inputs for OnceDifferentiable?~

Closes #476 (almost)

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -5,7 +5,8 @@ module Optim
     using Compat
     using LineSearches
     using NLSolversBase
-    using Calculus
+using Calculus
+    using DiffEqDiffTools
 #    using ReverseDiff
     using ForwardDiff
 
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -5,7 +5,7 @@ module Optim
     using Compat
     using LineSearches
     using NLSolversBase
-using Calculus
+    using Calculus
     using DiffEqDiffTools
 #    using ReverseDiff
     using ForwardDiff
--- a/src/objective_types.jl
+++ b/src/objective_types.jl
@@ -3,9 +3,9 @@ function OnceDifferentiable(f, x_seed::AbstractArray{T}, F::Real = zero(T),
                             autodiff = :finite) where T
     if autodiff == :finite
         # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
-        gcache = DiffEqDiffTools.GradientCache(DF, x_seed, nothing, nothing, nothing, Val{:central})
+        gcache = DiffEqDiffTools.GradientCache(x_seed, x_seed, nothing, nothing, nothing, Val{:central})
         function g!(storage, x)
-            DiffEqDiffTools.finite_difference_gradient!(storage, f, x, Val{:central})
+            DiffEqDiffTools.finite_difference_gradient!(storage, f, x, gcache)
             return
         end
         function fg!(storage, x)
@@ -35,11 +35,11 @@ function TwiceDifferentiable(f, g!, x_seed::AbstractVector{T}, F::Real = zero(T)
     end
     if autodiff == :finite
         # TODO: Create / request Hessian functionality in DiffEqDiffTools?
+        #       (Or is it better to use the finite difference Jacobian of a gradient?)
         # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
         jcache = DiffEqDiffTools.JacobianCache(x_seed, nothing, nothing, nothing, Val{:central})
         function h!(storage, x)
             DiffEqDiffTools.finite_difference_jacobian!(storage, g!, x, jcache)
-            #Calculus.finite_difference_hessian!(f, x, storage)
             return
         end
     elseif autodiff == :forward
@@ -58,11 +58,11 @@ function TwiceDifferentiable(d::OnceDifferentiable, x_seed::AbstractVector{T} = 
                              F::Real = zero(T); autodiff = :finite) where T<:Real
     if autodiff == :finite
         # TODO: Create / request Hessian functionality in DiffEqDiffTools?
+        #       (Or is it better to use the finite difference Jacobian of a gradient?)
         # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
         jcache = DiffEqDiffTools.JacobianCache(x_seed, nothing, nothing, nothing, Val{:central})
         function h!(storage, x)
             DiffEqDiffTools.finite_difference_jacobian!(storage, d.df, x, jcache)
-            #Calculus.finite_difference_hessian!(d.f, x, storage)
             return
         end
     elseif autodiff == :forward
@@ -76,25 +76,21 @@ end
 
 function TwiceDifferentiable(f, x::AbstractVector{T}, F::Real = zero(T);
                              autodiff = :finite) where T
-    # TODO: Assigning DF in input parameters is ambiguous with NLSolversBase
-    DF= NLSolversBase.alloc_DF(x,F)
     if autodiff == :finite
         # TODO: Allow user to specify Val{:central}, Val{:forward}, Val{:complex}
-        gcache = DiffEqDiffTools.GradientCache(DF, x, nothing, nothing, nothing, Val{:central})
+        gcache = DiffEqDiffTools.GradientCache(x, x, nothing, nothing, nothing, Val{:central})
         function g!(storage, x)
-            DiffEqDiffTools.finite_difference_gradient!(storage, f, x, Val{:central})
+            DiffEqDiffTools.finite_difference_gradient!(storage, f, x, gcache)
             return
         end
         function fg!(storage::Vector, x::Vector)
             g!(storage, x)
             return f(x)
         end
-        # TODO: Create / request Hessian functionality in DiffEqDiffTools?
         # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
-        jcache = DiffEqDiffTools.JacobianCache(x, nothing, nothing, nothing, Val{:central})
         function h!(storage::Matrix, x::Vector)
-            DiffEqDiffTools.finite_difference_jacobian!(storage, g!, x, jcache)
-            #Calculus.finite_difference_hessian!(f, x, storage)
+            # TODO: Wait to use DiffEqDiffTools until they introduce the Hessian feature
+            Calculus.finite_difference_hessian!(f, x, storage)
             return
         end
     elseif autodiff == :forward
@@ -112,5 +108,5 @@ function TwiceDifferentiable(f, x::AbstractVector{T}, F::Real = zero(T);
     else
         error("The autodiff value $(autodiff) is not supported. Use :finite or :forward.")
     end
-    TwiceDifferentiable(f, g!, fg!, h!, x, F, DF)
+    TwiceDifferentiable(f, g!, fg!, h!, x, F)
 end
--- a/src/objective_types.jl
+++ b/src/objective_types.jl
@@ -2,8 +2,8 @@ function OnceDifferentiable(f, x_seed::AbstractArray{T}, F::Real = real(zero(T))
                             DF::AbstractArray = NLSolversBase.alloc_DF(x_seed, F);
                             autodiff = :finite) where T
     if autodiff == :finite
-        # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
-        gcache = DiffEqDiffTools.GradientCache(x_seed, x_seed, nothing, nothing, nothing, Val{:central})
+        # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex} (requires care when using :forward I think)
+        gcache = DiffEqDiffTools.GradientCache(x_seed, x_seed, Val{:central})
         function g!(storage, x)
             DiffEqDiffTools.finite_difference_gradient!(storage, f, x, gcache)
             return
@@ -37,7 +37,7 @@ function TwiceDifferentiable(f, g!, x_seed::AbstractVector{T}, F::Real = real(ze
         # TODO: Create / request Hessian functionality in DiffEqDiffTools?
         #       (Or is it better to use the finite difference Jacobian of a gradient?)
         # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
-        jcache = DiffEqDiffTools.JacobianCache(x_seed, nothing, nothing, nothing, Val{:central})
+        jcache = DiffEqDiffTools.JacobianCache(x_seed, Val{:central})
         function h!(storage, x)
             DiffEqDiffTools.finite_difference_jacobian!(storage, g!, x, jcache)
             return
@@ -60,7 +60,7 @@ function TwiceDifferentiable(d::OnceDifferentiable, x_seed::AbstractVector{T} = 
         # TODO: Create / request Hessian functionality in DiffEqDiffTools?
         #       (Or is it better to use the finite difference Jacobian of a gradient?)
         # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
-        jcache = DiffEqDiffTools.JacobianCache(x_seed, nothing, nothing, nothing, Val{:central})
+        jcache = DiffEqDiffTools.JacobianCache(x_seed, Val{:central})
         function h!(storage, x)
             DiffEqDiffTools.finite_difference_jacobian!(storage, d.df, x, jcache)
             return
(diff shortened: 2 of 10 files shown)


pkofod:
cc @dextorious for comments

dextorious:
The cases where Calculus.jl looks faster than DiffEqDiffTools.jl are essentially benchmarking allocating the cache. With a pre-allocated GradientCache DiffEqDiffTools.jl is a bit over 2x faster than Calculus.jl regardless of problem size. Allocating the cache, however, takes between a few microseconds (for tiny problems) and a hundred or so microseconds (I got 50us for a R^N->R problem with N=1000), so small problems disproportionately benefit from having a preallocated cache. For reasonably large problems (or expensive functions) DiffEqDiffTools.jl will pull ahead, but if small problems with non-reusable caches are a frequent usecase for Optim.jl, we can look into creating a fast path tailored to that usecase.

As for Hessians, implementing proper handling of Hessians is the next big item on the todo list for DiffEqDiffTools.jl. Right now, if you have a known gradient, you can just compute the Jacobian and that'll be fast and reliable. Otherwise, computing the Jacobian of the gradient can work, but is obviously a very sub-optimal solution. A proper solution, including support for complex-valued functions, will take me a bit of time, but it's going to happen in the near future.

Finally, a note on complex support. We don't do any checking for complex differentiability (Cauchy-Riemann would be expensive), so there's an infinite variety of ways to contrive sufficiently ill-behaved functions that will yield nonsense outputs. There might also be bugs there that haven't been found yet, it's hard to cover every edge case in tests and the library is still very new.

anriseth:
Thanks @dextorious 

> but if small problems with non-reusable caches are a frequent usecase for Optim.jl, we can look into creating a fast path tailored to that usecase.

I think that many such cases can be handled in one of two ways within Optim already. 
1. Optimizations of the same function with different parameters. *Can be dealt with by creating a callable struct instance `f` and then reuse the same `OnceDifferentiable(f)` multiple times*
2. Multiple problems of same size and type. *Can be dealt with by adding a `GradientCache` input for the `OnceDifferentiable`s that can be reused multiple times.* (This would also solve 1.)

> A proper solution, including support for complex-valued functions, will take me a bit of time, but it's going to happen in the near future.

At the moment we have complex-support for neither gradients nor Hessians, so the current DiffEqDiffTools features are already a great improvement :)

ChrisRackauckas:
>Multiple problems of same size and type. Can be dealt with by adding a GradientCache input for the OnceDifferentiables that can be reused multiple times. (This would also solve 1.)

Yes, DiffEqDiffTools.jl should be treated like ForwardDiff and its cache should be made at the time of the OnceDifferentiable (like the config). Additionally, is this testing @dextorious 's newest branch? That should be much faster.

anriseth:
> Yes, DiffEqDiffTools.jl should be treated like ForwardDiff and its cache should be made at the time of the OnceDifferentiable (like the config).

[That is what I do in this PR](https://github.com/JuliaNLSolvers/Optim.jl/pull/518/files#diff-ab516b956239be1e33f110faaa3a8b4cR6). The "issue" is that the test problems are so cheap to solve that the creation of the `GradientCache` becomes a significant factor. For more costly problems I assume the DiffEqDiffTools approach becomes faster / scales better.

> Additionally, is this testing @dextorious 's newest branch? That should be much faster.

I am using the [gradientcache branch](https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/tree/gradientcache), is that correct?

src/objective_types.jl
```diff
@@ -60,19 +73,28 @@ function TwiceDifferentiable(d::OnceDifferentiable, x_seed::AbstractVector{T} =
...
+function TwiceDifferentiable(f, x::AbstractVector{T}, F::Real = zero(T);
+                             autodiff = :finite) where T
+    # TODO: Assigning DF in input parameters is ambiguous with NLSolversBase
+    DF= NLSolversBase.alloc_DF(x,F)
     if autodiff == :finite
-        function g!(storage::Vector, x::Vector)
-            Calculus.finite_difference!(f, x, storage, :central)
+        # TODO: Allow user to specify Val{:central}, Val{:forward}, Val{:complex}
+        gcache = DiffEqDiffTools.GradientCache(DF, x, nothing, nothing, nothing, Val{:central})
+        function g!(storage, x)
```
ChrisRackauckas: Do you need to `let` block this?

src/objective_types.jl
```diff
@@ -1,7 +1,11 @@
-function OnceDifferentiable(f, x_seed::AbstractArray{T}, F = zero(T); autodiff = :finite) where T
+function OnceDifferentiable(f, x_seed::AbstractArray{T}, F::Real = zero(T),
+                            DF::AbstractArray = NLSolversBase.alloc_DF(x_seed, F);
+                            autodiff = :finite) where T
     if autodiff == :finite
+        # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
+        gcache = DiffEqDiffTools.GradientCache(DF, x_seed, nothing, nothing, nothing, Val{:central})
         function g!(storage, x)
-            Calculus.finite_difference!(f, x, storage, :central)
+            DiffEqDiffTools.finite_difference_gradient!(storage, f, x, Val{:central})
```
ChrisRackauckas: this isn't using `gcache`?

src/objective_types.jl
```diff
@@ -60,19 +73,28 @@ function TwiceDifferentiable(d::OnceDifferentiable, x_seed::AbstractVector{T} =
...
+                             autodiff = :finite) where T
+    # TODO: Assigning DF in input parameters is ambiguous with NLSolversBase
+    DF= NLSolversBase.alloc_DF(x,F)
     if autodiff == :finite
-        function g!(storage::Vector, x::Vector)
-            Calculus.finite_difference!(f, x, storage, :central)
+        # TODO: Allow user to specify Val{:central}, Val{:forward}, Val{:complex}
+        gcache = DiffEqDiffTools.GradientCache(DF, x, nothing, nothing, nothing, Val{:central})
+        function g!(storage, x)
+            DiffEqDiffTools.finite_difference_gradient!(storage, f, x, Val{:central})
```
ChrisRackauckas: This isn't using `gcache`?

src/objective_types.jl
anriseth: Ah, that's embarrassing, I forgot to add it back. Thank you!

src/objective_types.jl
anriseth: Can you expand on this comment, please?
Are you saying I should just do `g!(storage, x) = DiffEqDiffTools.finite_difference_gradient!(...)`? 
If so, we can probably update that - I don't know why it was of this form previously (@pkofod ?)

ChrisRackauckas: A better fix is just to fix the inferrability of the `gcache`, which right now isn't fully specializing due to `::Datatype` calls instead of `::Type{...}`. That's on our end, but for now let block localizing would give you type-stability in the gradient call and get rid of a single dynamic dispatch per gradient calculation (which might not actually matter for the benchmarks though)

dextorious:
So, I may have spoked a bit hastily on the gradient performance front. @ChrisRackauckas pointed out an inference bug in the cache construction and I further specialized the gradient calculation for subtypes of StridedArrays, which doesn't require any storage in cache (but still uses GradientCache for type information). With these changes, constructing an instance of GradientCache without allocations takes no more than a few nanoseconds and performance for small problems should no longer be an issue. As a result, there should no longer be any situations where DiffEqDiffTools is slower than Calculus (if you find an exception, that's a performance bug, please let me know).

anriseth:
After the latest update of DiffEqDiffTools, gradient evaluation shows approximately the same number of using DiffEqDiffTools and Calculus. 

Interestingly, there are 2-5 extra allocations during an optimization run when using GradientCache rather than just calling `finite_difference_gradient!(..., Val{:central})`. Is that expected behavior, or am I doing something wrong @dextorious ? 


**Regarding complex numbers**, I think there is a case you haven't considered.
If I have a function `f: C^n \to R`, that is, complex input and real output, then the current code seems to miss out the derivatives in the imaginary directions?

Try for example with `f(x) = sum(abs2, x)`, which I think should have gradient `g(x) = 2conj(x)`. (Correct me if I'm wrong). 
```jl
julia> x = ones(2)*(1+im)
2-element Array{Complex{Float64},1}:
 1.0+1.0im
 1.0+1.0im

julia> f(x)
4.0

julia> g(x)
2-element Array{Complex{Float64},1}:
 2.0-2.0im
 2.0-2.0im

julia> DiffEqDiffTools.finite_difference_gradient(f, x)
2-element Array{Complex{Float64},1}:
 2.0+0.0im
 2.0+0.0im
```

dextorious:
> Interestingly, there are 2-5 extra allocations during an optimization run when using GradientCache rather than just calling finite_difference_gradient!(..., Val{:central}). Is that expected behavior, or am I doing something wrong @dextorious ?

That sounds like a bug, there should be at most one extra allocation in that case, assuming you're using subtypes of StridedArrays. I'll investigate.

> Regarding complex numbers, I think there is a case you haven't considered.
If I have a function f: C^n \to R, that is, complex input and real output, then the current code seems to miss out the derivatives in the imaginary directions?

Right, that's the sort of silly bug I expected to find. I'll fix it today.

anriseth:
Complex derivatives now work :D 

This is ready to merge whenever DiffEqDiffTools is tagged.

src/objective_types.jl
```diff
@@ -1,7 +1,11 @@
-function OnceDifferentiable(f, x_seed::AbstractArray{T}, F = zero(T); autodiff = :finite) where T
+function OnceDifferentiable(f, x_seed::AbstractArray{T}, F::Real = real(zero(T)),
+                            DF::AbstractArray = NLSolversBase.alloc_DF(x_seed, F);
+                            autodiff = :finite) where T
     if autodiff == :finite
+        # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex}
+        gcache = DiffEqDiffTools.GradientCache(x_seed, x_seed, nothing, nothing, nothing, Val{:central})
```
anriseth: @dextorious passing `x_seed` to the `df` input of GradienCache isn't needed, but is it problematic?

If I understand correctly, it would be fine to allow `GradientCache(df::Union{Void,AbstractArray{<:Number}},...)`, as `df` is not used for the `R^n\to R` or `C^n\to R` cases?.

dextorious: For `f: vector->scalar` maps `df` is only used to check for the `R^n -> C` case, so passing `x_seed` there shouldn't be problematic. Letting users pass `nothing` there could lead to issues where we can't deduce the correct code path (in general, not for your usecase), but I'll see if we can do it sanely with some warnings.

anriseth: Okay, no worries, then. I'm happy for you to leave it as-is.

antoine-levitt:
Re complex finite differencing, make sure you use the correct definition of a gradient. The one that makes optimization algorithms work (ie, that is compatible with the real-to-complex isomorphisms between R^2n and C^n) is the one outlined in http://julianlsolvers.github.io/Optim.jl/latest/algo/complex/. In particular, @anriseth, **the gradient of f(x) = sum(abs2, x) should be g(x) = 2x, not g(x) = 2conj(x)**. If you're not convinced by the isomorphism argument above, try for yourself what the gradient algorithm xn+1 -= alpha*g(xn) looks like. The definition of a gradient C^n -> R (the one of interest to optim) is fundamentally different from the notion of a complex-differential function.

anriseth:
> In particular, @anriseth, the gradient of f(x) = sum(abs2, x) should be g(x) = 2x, not g(x) = 2conj(x).

Yes, we convinced ourselves about the same result on gitter last week :) I believe we got the correct implementation in DiffEqDiffTools :+1:

pkofod:
Is diffeqdifftools tagged and would you mind resolving conflicts here @anriseth ?

src/Optim.jl
```diff
@@ -5,7 +5,8 @@ module Optim
     using Compat
     using LineSearches
     using NLSolversBase
-    using Calculus
+using Calculus
```
ChrisRackauckas: spacing

pkofod: yeah, but this will be removed though, as gradients and Hessians are in now, right?

ChrisRackauckas: finishing gradients right now. Hessians I don't think so, @dextorious ?

pkofod: ah, alright! Sorry, I'm a bit out of the loop these days. I didn't mean to impose any extra work on you guys, just thought that was done :)

ChrisRackauckas:
The DiffEqDiffTools release with gradients was just tagged. Also, there's an API documented on the README now.

src/Optim.jl
dextorious: Yeah, Hessians are the last major item on the immediate TODO list. Hopefully we'll get those done soon as well.

anriseth:
I've fixed the conflicts, and the tests should pass when https://github.com/JuliaLang/METADATA.jl/pull/13372 is merged. (I had a failure in `arrays.jl` on my laptop, which I think is unrelated?)

Many thanks to the DiffEq guys :D

pkofod:
> (I had a failure in arrays.jl on my laptop, which I think is unrelated?)

non-deterministic?

anriseth:
> non-deterministic?

Yes, ParticleSwarm fails in testset "Vector" the first time I call "runtests.jl", but not anytime after that ...

anriseth:
This is good to go :+1: @pkofod

pkofod:
> Yes, ParticleSwarm fails in testset "Vector" the first time I call "runtests.jl", but not anytime after that ..

another one of those...

src/objective_types.jl
```diff
@@ -1,7 +1,11 @@
-function OnceDifferentiable(f, x_seed::AbstractArray{T}, F = zero(T); autodiff = :finite) where T
+function OnceDifferentiable(f, x_seed::AbstractArray{T}, F::Real = real(zero(T)),
+                            DF::AbstractArray = NLSolversBase.alloc_DF(x_seed, F);
+                            autodiff = :finite) where T
     if autodiff == :finite
+        # TODO: Allow user to specify Val{:central}, Val{:forward}, :Val{:complex} (requires care when using :forward I think)
```
pkofod: @dextorious might be able to correct me, but I think the vast majority of users are better off with central differences, so in my mind it is OK that we only expose that. Complex should simply depend on `T` right?

anriseth: The TODO was there just in case someone wants to fix it in the future :p I would not bother using :central or :complex here myself.

`Val{:complex}` here means calculating finite differences of a function `f:R -> R` by taking a step in a complex direction. See here https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/

pkofod: ah, yes. I remember looking that up at some point when looking through DiffEqDiffTools. Yeah just leave it there.

antoine-levitt: > Val{:complex} here means calculating finite differences of a function f:R -> R by taking a step in a complex direction. See here https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/

That's a very cute trick! Probably better to rename it complex_step or something so that it is clear it refers to a specific algorithm and is not intended for complex functions. (also this kind of functionality is dangerous because it silently fails when the function is not given in complex analytic form, even if it is real analytic and perfectly good for differencing, e.g. x -> abs(x)^2 - which sounds ridiculous as written but is perhaps less when written norm(x)^2 or something)

dextorious: The naming isn't ideal, though in the literature it's almost uniformly referred to as "complex-step differentiation". As the API becomes more stabilized and we move on to finally writing some documentation for all this, this and other rough edges will probably be improved along the way. It's true that it has limited applicability and I wouldn't suggest using it as a default recommendation for anyone, but it does give extremely good accuracy quite cheaply, which is sometimes useful.

As for complex vs forward, I'd say pretty much the only reason to ever use forward differences is that they're substantially faster (on the order of 2x usually). If accuracy isn't an issue and for some reason the differentiation is very performance sensitive, I'd recommend `Val{:forward}`. Central differences are the sensible default.

anriseth:
Errors should be fixed now :)

PR #520, opened 2018-01-26
Allow for AbstractArrays such as ArrayPartition from RecursiveArrayTo…

pkofod:
…ols.

Loooots of unnecessary annotations here. There are more places that could be cleaned up as well.

Anyway, you can now do what was asked in https://discourse.julialang.org/t/optim-arguments-higher-dimensional-arrays-and-gradient-specification/8500/10 (see some output there)

- [x] needs tests
- [x]  needs zeroth order support


antoine-levitt:
AFAICT this changes the storage of LBFGS from a matrix to an array of (dynamic) arrays. You gain in generality and simplicity but you may lose in performance because you lose data locality. In any case if you decide to do this you should remove the `devec_fun` part (which was there to avoid the side-effects of `vec()`ing the vectors for `twoloop!`

Looking at this I discovered that twoloop! seems to do a matrix vector product "by hand" with dot products, which is a bad idea (BLAS2 is better than BLAS1). Am I missing something?

ChrisRackauckas:
>You gain in generality and simplicity but you may lose in performance because you lose data locality.

Not necessarily though. Breaking up a larger array into smaller pieces can actually improve caching performance and this is demonstrated in some broadcasting operations on ArrayPartitions (at least on v0.6). In addition, using so many views can hurt. It's less painful on v0.7 if you hit a spot with the GC optimization, but that doesn't always come into effect. I would be very weary of any performance statement either way without a benchmark, and wouldn't be surprised to see the Array of Arrays approach scale better (though perform a little worse on small systems). 

>Looking at this I discovered that twoloop! seems to do a matrix vector product "by hand" with dot products, which is a bad idea (BLAS2 is better than BLAS1). Am I missing something?

Yeah, looks to me like that may be the case too. If `twoloop!` could be BLAS2 then that would definitely be faster.

pkofod:
Ah, I hadn't merged the gmres changes I see...

Thanks for the comments @antoine-levitt . In my personal experience, I havn't found the array of arrays vs matrix to be a real show-stopper, especially for large rproblems as Chris mentions. you may be right, and I'll be happy to benchmark `twoloops`. Regarding the `devec`: I'll have a look.

I'm not sure about the matrix vector multiplication thing. `q` and `r` are updated each iteration, so it's not the same thing that's being dotted with `dxi` and `dgi` each time if that's what you meant?

antoine-levitt:
I'm not so familiar with the implementation of LBFGS, but I was under the impression that it built a low-rank approximation to the inverse Hessian. If that is the case, the inverse Hessian should look like P D P^T, where P is N x m, and so it should be a matvec?

Interesting point about views and cache. In any case an array of arrays forbids a BLAS2 implementation.

pkofod:
It's not building it though. Two-loops is solving directly for the search direction

antoine-levitt:
That's the point of the low rank approximation : it is storing the inverse hessian as P D P^T, where P is N by m and D m by m. Doing so applying the inverse hessian is O(N m), compared with BFGS's O(N^2).

Looking at Wikipedia it seems like this "two loop" algorithm is the standard way of doing things. I guess it's so that you store the full history and not some linear combinations or previous vectors to build an orthogonal basis for the range of the inverse hessian, a la Arnoldi? I would think that would be faster due to BLAS 2, but there might be a good reason

pkofod:
> that or everybody just uses the code in Nocedal & Wright.

I think it's from the original paper, Nocedal (1980) [https://doi.org/10.1090/S0025-5718-1980-0572855-7]

original

And then he describes the two-loop recursion. So I think people maybe just take his word for it that it's the best way? :) He says efficiently, but from what I can see, he doesn't qualify what he means by "efficiently".

anriseth:
Maybe the scaling of the inverse Hessian requires a twoloop approach? (Ref #484 )

antoine-levitt:
http://users.iems.northwestern.edu/~nocedal/lbfgsb.html, which is used by scipy for instance, uses a matrix representation (see for instance http://users.iems.northwestern.edu/~nocedal/PDFfiles/limited.pdf). In 1980 I would imagine the BLAS1/BLAS2 tradeoffs and software availability were pretty different from what they are today...

pkofod:
> http://users.iems.northwestern.edu/~nocedal/lbfgsb.html, which is used by scipy for instance, uses a matrix representation (see for instance http://users.iems.northwestern.edu/~nocedal/PDFfiles/limited.pdf). In 1980 I would imagine the BLAS1/BLAS2 tradeoffs and software availability were pretty different from what they are today...

I'll have a look at the paper.

I totally agree that these things evolve over time. When I worked on porting the libm-stuff there were a long of things in there that were from a time where floating point comparisons were much more expensive, so I don't doubt that this could be the case here as well. If it is the case, and if Nocedal and Wright make a third edition of their book, then they should mention this :)

I won't have time to look at this until after next week, but it seems worth experimenting with.

pkofod:
Just to explain why I did it, it's because there is no fallback for `a .= b` where `a` is an array and `b` is an ArrayPartition from RecursiveArrayTools.jl, so I really did this to avoid writing out a loop to copy in the contents in each column.

@ChrisRackauckas what's the time frame on a fix for that infinite recursion? Then I'd be happy to keep the matrix version - we could even allow people to choose which update they want.

antoine-levitt:
It does seem worth it to look carefully at the LBFGS implementation, as LBFGS really is the solver of choice for large scale problems (at least in the cases I tried).

Anyway, sorry for sidetracking the discussion, the proposed PR looks great, the memory locality probably does not matter that much, and the implementation should eventually be rewritten anyway.

pkofod:
> Anyway, sorry for sidetracking the discussion,

I appreciate the input. I had never considered not using the recursion. Let's open an issue instead.

ChrisRackauckas:
>@ChrisRackauckas what's the time frame on a fix for that infinite recursion?

Let me look at that today.

ChrisRackauckas:
Ugh, it's calling `broadcast_c` for the in-place thing for some reason. I think it's hitting some broadcast bug... but there's no reason to really attempt to do too much because of 

https://github.com/JuliaLang/julia/pull/25377

That said, I can easily overload `copy!`. Good enough?

pkofod:
Sure , as long as it works I'm glad 😁

ChrisRackauckas:
Alright, `copy!` works. Since that normally falls back to `a.=b`, that's a good replacement for now.

pkofod:
I just need to include a test for the various solvers and a function that takes in ArrayPartitions and then I'll merge

pkofod:
This is waiting for an NLSolversBase patch to be accepted in metadata

PR #517, opened 2018-01-21
Work with OptimTestProblems 1.2

anriseth:
Update tests to work with https://github.com/JuliaNLSolvers/OptimTestProblems.jl/pull/9
(Whenever that's merged and OptimTestProblems ~2.0~ 1.2 is registered)

PR #533, opened 2018-02-17
Introduce counter for f_tol (IPNewton support)

anriseth:
This PR lets the user set the option `successive_f_tol::Int = n` so that an optimization routine only converges if `f_converged = true` happens `n+1` times in a row. The default is `Options(successive_f_tol=0)`.

This functionality is taken from IPNewton. I like this option, however, as the `f_tol` convergence criteria is a bit artificial. The same goes for `x_tol`, and I can add that as well if you guys wish. Alternatively we can wait with `x_tol` until (if) we rewrite the convergence assesment code.

I also changed the order of the `assess_convergence` functions. As most people care about the unconstrained one, it feels natural to have it first in the file.

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -65,7 +65,7 @@ function optimize(d::D, initial_x::Tx, method::M,
         end
 
         # Check time_limit; if none is provided it is NaN and the comparison
-        # will always return false.xas
+        # will always return false.
         stopped_by_time_limit = time()-t0 > options.time_limit ? true : false
         f_limit_reached = options.f_calls_limit > 0 && f_calls(d) >= options.f_calls_limit ? true : false
         g_limit_reached = options.g_calls_limit > 0 && g_calls(d) >= options.g_calls_limit ? true : false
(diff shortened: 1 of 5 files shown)


src/multivariate/optimize/optimize.jl
```diff
@@ -61,6 +64,8 @@ function optimize(d::D, initial_x::AbstractArray{Tx, N}, method::M,
             stopped_by_callback = trace!(tr, d, state, iteration, method, options)
         end
 
+        # Check time_limit; if none is provided it is NaN and the comparison
+        # will always return false.xas
```
pkofod: `false.xas`? :)

pkofod:
Great, thanks. Was `xas` a typo?

anriseth:
> Great, thanks. Was xas a typo? That is, your covfefe

Yes, my own baby covfefe :laughing:  https://www.babycenter.com/baby-names-xas-968375.htm

Fixed it now.

anriseth:
This needs #517 to pass the newly added tests from OptimTestProblems 1.2

PR #535, opened 2018-02-20
Suppress trace output.

pkofod:


test/multivariate/solvers/zeroth_order/particle_swarm.jl
```diff
@@ -27,11 +27,12 @@
     res = Optim.optimize(rosenbrock_s, initial_x, ParticleSwarm(lower, upper, n_particles),
                              options)
     @test norm(Optim.minimizer(res) - [1.0, 1.0]) < 0.1
-
+    @suppress_out begin
     # TODO: Is there a way to suppress the trace results?
```
anriseth: You can remove this line now :D

test/REQUIRE
```diff
@@ -1,2 +1,3 @@
 OptimTestProblems 1.2
 RecursiveArrayTools
+Suppressor
```
anriseth: Seems unhappy with your newline?

pkofod: what's unhappy?

pkofod: ah, it didn't show out here in the comments

Issue #173, opened 2016-02-25
Experience with SimulatedAnnealing?


pkofod:
I've been running some benchmarks on all our methods, and find that `SImmulatedAnnealing` is doing **very** poorly. Does anyone have experience with using the method as implemented here? I'm kind of suspecting that a bug has been introduced at some point. I will post code when I have something presentable, but many well-known test problems have been tested, and it often runs for the maximum number of iterations. 

Edit: I should try to increase the number of iterations... Maybe 1000 is just not enough for the method to converge?

johnmyleswhite:
Simulated annealing almost never works well in practice, but it would be good to know if there's a bug in our implementation. Perhaps contrast a run of our method with R's implementation in the `optim` function?

Top culprits are: (a) the neighbor function and (b) the annealing schedule function.

multidis:
> Simulated annealing almost never works well in practice

As with any global heuristic, it depends on how one runs it. Generalized (*C. Tsallis, Physica A 233 (1996) 395-406*) simulated annealing based on a generalization of Gibbs statistics can work well. Some examples can be found in the `GenSA` R-package, *Xiang et al. R Journal 5/1 (2013) 13-28* where it outperforms differential evolution. And there is an exact random number generator (RNG) for that distribution (*T. Schanze, Comp. Phys. Comm. 175 (2006) 708–712*).

I started the implementation based the `Optim.jl` "simple" simulated annealing function a year ago but never had time to complete. Thought it would be good to have at least one global method in `Optim.jl`, conceptually different from DE implementation in other Julia packages. Would be glad to share existing code including the RNG with testing if someone has interest.

pkofod:
I would love to see it, and compare with our current implementation. I will also return with a bit more specific examples when I get the time to write it up.

multidis:
Sure, I'll collect what I have in the next few days.

pkofod:
Okay, I didn't have much time this evening, so I cannot make a complete report, but I might have been a bit too harsh on `SimulatedAnnealing` (correct spelling this time...). Generally, it's in the middle of the field, so to speak. Accelerated gradient descent is almost always closest to the minimizer (and consequently has the smallest "minimum error"), and is quite fast as well. Nelder Mead is quite often in the second place.

[...]

PR #553, opened 2018-03-17
Add simulated annealing with bounds.

pkofod:
If I make this PR, I might actually finish the Optim.jlization of SAMIN that I should have done years ago :) 

https://github.com/mcreel/Econometrics.jl/blob/master/src/Optimization/samin.jl

ref #173


pkofod:
@anriseth you always have a good eye for stupid mistakes

@mcreel looks like I finally got my act together ;)

docs/mkdocs.yml
```diff
@@ -46,6 +46,8 @@ pages:
             - Hessian Required:
                 - Newton: 'algo/newton.md'
                 - Newton with Trust Region: 'algo/newton_trust_region.md'
+        - Constrained:
```
anriseth: Can we add some note about Fminbox and ConstrainedOptim here? Just to ensure that users don't make the assumption that `samin` is all we have to offer.

src/multivariate/solvers/constrained/samin.jl
```diff
@@ -0,0 +1,323 @@
...
+
+# TODO
+# Handle traces
+# * details: a px3 matrix. p is the number of times improvements were found.
+#            The columns record information at the time an improvement was found
+#            * first: cumulative number of function evaluations
+#            * second: temperature
+#            * third: function value
+#
+# Add doc entry
```
anriseth: Doc entry has been added ;)

src/multivariate/solvers/constrained/samin.jl
```diff
@@ -0,0 +1,323 @@
...
+                                            false, #f_increased,
+                                            tr,
+                                            f_calls(d),
+                                            g_calls(d),
+                                            h_calls(d))
+
+end
+
+# TODO
+# Handle traces
```
anriseth: Are you planning to handle traces?

test/multivariate/solvers/constrained/samin.jl
```diff
@@ -0,0 +1,12 @@
...
+using OptimTestProblems
+probs = OptimTestProblems.MultivariateProblems.UnconstrainedProblems.examples
+prob = probs["Himmelblau"]
+
+xtrue = prob.solutions
+f = OptimTestProblems.MultivariateProblems.objective(prob)
+x0 = prob.initial_x
+optimize(f, x0, x0.-100., x0.+100.0, Optim.SAMIN())
+
+optimize(f, x0, SimulatedAnnealing(), Optim.Options(iterations = 3000))
```
anriseth: Maybe add some `@test` lines here to ensure consistency with future edits of the algorithm?

test/multivariate/solvers/constrained/samin.jl
```diff
@@ -0,0 +1,12 @@
+using Revise
```
anriseth: - Remove the `using` keywords here
- Add `@testset`
- Use `MVP` instead of `OptimTestProblems.MultivariateProblems`.
- Include the file in `runtests.jl`

docs/mkdocs.yml
anriseth: I guess we can say call it "With derivatives" or something like that?

mcreel:
Great! I believe that people think of SA as a global, derivative-free method, more than as a constrained method. It is true that there are box constraints, but they are typically set wide enough so that they should not be binding at the solution, and the method gives a warning if the final results are close to the bounds. The box is a convenient approximation to the idea of an abstract compact parameter space. Just some thoughts to help with thinking about how to classify the method.

mcreel:
I tried it out from master, looks good! For the Rosenbrock example, with the wide bounds, setting rt=0.5 is too low a temperature, and the algorithm "quenches", which means it cools too fast. If you set SAMIN(rt=0.9) you get a much more accurate solution.

PR #556, opened 2018-03-22
Update samin.md

mcreel:
The example was "quenching" because the cooling rate was too fast. This small change obtains a much more accurate solution.

### DIFF
--- a/docs/src/algo/samin.md
+++ b/docs/src/algo/samin.md
@@ -25,23 +25,23 @@ julia> using Optim, OptimTestProblems
 
 julia> prob = OptimTestProblems.UnconstrainedProblems.examples["Rosenbrock"];
 
-julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(), Optim.Options(iterations=20000))
-Results of Optimization Algorithm
+julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(rt=0.9), Optim.Options(iterations=20000))
+Results of Optimization AlgorithmResults of Optimization Algorithm
  * Algorithm: SAMIN
  * Starting Point: [-1.2,1.0]
- * Minimizer: [1.0170012560033472,1.0343750729202243]
- * Minimum: 2.897402e-04
- * Iterations: 5901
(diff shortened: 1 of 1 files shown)


docs/src/algo/samin.md
```diff
@@ -25,23 +25,23 @@ julia> using Optim, OptimTestProblems
 
 julia> prob = OptimTestProblems.UnconstrainedProblems.examples["Rosenbrock"];
 
-julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(), Optim.Options(iterations=20000))
-Results of Optimization Algorithm
+julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(rt=0.9), Optim.Options(iterations=20000))
+Results of Optimization AlgorithmResults of Optimization Algorithm
```
ChrisRackauckas: spacing

pkofod: this is not spacing, this is because @mcreel missed a line (that should've been deleted) when he pasted the new output :)

pkofod:
Would it maybe be valuable to show the output of the default also, to show how `rt` can help here by not reducing the temperature too quickly? May be a valuable lesson to users.

mcreel:
Actually, the better solution might be to make the default rt=0.9, and tighten up the f_tol and x_tol. Better that people get the right answer and spend more time than get the wrong answer. This is a method that requires some experience to make work well, and reading the reference papers helps a lot.  I see that the algorithm is stopping before the iteration limit is hit, but that the convergence message is false. We should probably make the SAMIN convergence rules follow the general rules of Optim, so that it's possible to see a positive convergence result. I will look into that in the next few days.

pkofod:
> Actually, the better solution might be to make the default rt=0.9, and tighten up the f_tol and x_tol. Better that people get the right answer and spend more time than get the wrong answer. This is a method that requires some experience to make work well, and reading the reference papers helps a lot. I see that the algorithm is stopping before the iteration limit is hit, but that the convergence message is false. We should probably make the SAMIN convergence rules follow the general rules of Optim, so that it's possible to see a positive convergence result. I will look into that in the next few days.

I'm fine with a less aggressive temperature scheme. 

I'm :+1: with the convergence change as well.

mcreel:
I will prepare another PR with a few more suggested changes

PR #558, opened 2018-03-23
samin changes: defaults, verbosity, and documentation

mcreel:
This increases the default temperature reduction rate from 0.5 to 0.9, which is more conservative, and which will be more likely to give the correct solution, at the cost of more iterations. The default tolerances are made more strict. The final results are displayed by default. The documentation has been extended with a bit of advice, and with examples of both successful and unsuccessful runs.

### DIFF
--- a/docs/src/algo/samin.md
+++ b/docs/src/algo/samin.md
@@ -25,23 +25,23 @@ julia> using Optim, OptimTestProblems
 
 julia> prob = OptimTestProblems.UnconstrainedProblems.examples["Rosenbrock"];
 
-julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(), Optim.Options(iterations=20000))
-Results of Optimization Algorithm
+julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(rt=0.9), Optim.Options(iterations=20000))
+Results of Optimization AlgorithmResults of Optimization Algorithm
  * Algorithm: SAMIN
  * Starting Point: [-1.2,1.0]
- * Minimizer: [1.0170012560033472,1.0343750729202243]
- * Minimum: 2.897402e-04
- * Iterations: 5901
+ * Minimizer: [1.0000002252116529,1.0000004516135839]
+ * Minimum: 5.086195e-14
+ * Iterations: 19851
  * Convergence: false
-   * |x - x'| ≤ 0.0e+00: false
-     |x - x'| = NaN
+   * |x - x'| ≤ 0.0e+00: false 
+     |x - x'| = NaN 
    * |f(x) - f(x')| ≤ 0.0e+00 |f(x)|: false
      |f(x) - f(x')| = NaN |f(x)|
-   * |g(x)| ≤ 0.0e+00: false
-     |g(x)| = NaN
+   * |g(x)| ≤ 0.0e+00: false 
+     |g(x)| = NaN 
    * Stopped by an increasing objective: false
    * Reached Maximum Number of Iterations: false
- * Objective Calls: 5901
+ * Objective Calls: 19851
  * Gradient Calls: 0
 ```
 
--- a/docs/src/algo/samin.md
+++ b/docs/src/algo/samin.md
@@ -21,13 +21,13 @@ multiple times to verify that the same solution is found each time. If this is n
 increase rt. When in doubt, start with a conservative rt, for example, rt=0.95, and allow for
 a generous iteration limit. The algorithm requires lower and upper bounds on the parameters,
 although these bounds are often set rather wide, and are not necessarily meant to reflect
-constraints in the model,but rather bounds that enclose the parameter space. If the final
+constraints in the model, but rather bounds that enclose the parameter space. If the final
 `x`s are very close to the boundary (which can be checked by setting verbosity=1), it is a
 good idea to restart the optimizer with wider bounds, unless the bounds actually reflect
 hard constraints on `x`.
 
 ## Example
-This is example shows a successful minimization:
+This example shows a successful minimization:
 ```julia
 julia> using Optim, OptimTestProblems
 
@@ -65,7 +65,7 @@ Results of Optimization Algorithm
  * Gradient Calls: 0
 ```
 ## Example
-This is example shows an unsuccessful minimization, because the cooling rate,
+This example shows an unsuccessful minimization, because the cooling rate,
 rt=0.5, is too rapid:
 ```julia
 julia> using Optim, OptimTestProblems
--- a/docs/src/algo/samin.md
+++ b/docs/src/algo/samin.md
@@ -3,35 +3,55 @@
 ```julia
 SAMIN(; nt::Int = 5     # reduce temperature every nt*ns*dim(x_init) evaluations
         ns::Int = 5     # adjust bounds every ns*dim(x_init) evaluations
-        rt::T = 0.5     # geometric temperature reduction factor: when temp changes, new temp is t=rt*t
+        rt::T = 0.9     # geometric temperature reduction factor: when temp changes, new temp is t=rt*t
         neps::Int = 5   # number of previous best values the final result is compared to
-        f_tol::T = 1e-8 # the required tolerance level for function value comparisons
-        x_tol::T = 1e-5 # the required tolerance level for x
-        coverage_ok::Bool = false, # increase temperature until parameter space is covered
+        f_tol::T = 1e-12 # the required tolerance level for function value comparisons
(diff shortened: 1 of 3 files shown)


docs/src/algo/samin.md
```diff
@@ -3,46 +3,105 @@
...
+increase rt. When in doubt, start with a conservative rt, for example, rt=0.95, and allow for
+a generous iteration limit. The algorithm requires lower and upper bounds on the parameters,
+although these bounds are often set rather wide, and are not necessarily meant to reflect
+constraints in the model,but rather bounds that enclose the parameter space. If the final
+`x`s are very close to the boundary (which can be checked by setting verbosity=1), it is a
+good idea to restart the optimizer with wider bounds, unless the bounds actually reflect
+hard constraints on `x`.
 
 ## Example
+This is example shows a successful minimization:
```
pkofod: either "is an example that" or "this example"

src/multivariate/solvers/constrained/samin.jl
```diff
@@ -22,12 +22,12 @@
...
+        rt::T = 0.9     # geometric temperature reduction factor: when temp changes, new temp is t=rt*t
         neps::Int = 5   # number of previous best values the final result is compared to
-        f_tol::T = 1e-8 # the required tolerance level for function value comparisons
-        x_tol::T = 1e-5 # the required tolerance level for x
-        coverage_ok::Bool = false, # increase temperature until parameter space is covered
-        verbosity::Int = 0) # scalar: 0, 1, 2 or 3 (default = 0).
+        f_tol::T = 1e-12 # the required tolerance level for function value comparisons
+        x_tol::T = 1e-6 # the required tolerance level for x
+        coverage_ok::Bool = false, # if false, increase temperature until initial parameter space is covered
+        verbosity::Int = 1) # scalar: 0, 1, 2 or 3 (default = 1).
```
pkofod: Is there a specific reason why you thing we should print this separately from the MultivariateOptimizationResults printing?

docs/src/algo/samin.md
```diff
@@ -3,46 +3,105 @@
...
+bounds constraints as described in Goffe et. al. (1994) and Goffe (1996). A key control
+parameter is rt, the geometric temperature reduction rate, which should be between zero
+and one. Setting rt lower will cause the algorithm to contract the search space more quickly,
+reducing the run time. Setting rt too low will cause the algorithm to narrow the search
+too quickly, and the true minimizer may be skipped over. If possible, run the algorithm
+multiple times to verify that the same solution is found each time. If this is not the case,
+increase rt. When in doubt, start with a conservative rt, for example, rt=0.95, and allow for
+a generous iteration limit. The algorithm requires lower and upper bounds on the parameters,
+although these bounds are often set rather wide, and are not necessarily meant to reflect
+constraints in the model,but rather bounds that enclose the parameter space. If the final
```
pkofod: there's a missing space after the comma "model,but"

src/multivariate/solvers/constrained/samin.jl
mcreel: Well, the way it is in Optim.jl master, right now, it prints what is below. The algorithm has stopped (at a poor solution, because rt was too small), but none of the convergence criteria are that are printed out have been met. It seems to me that users will not know why it stopped. If we print out the information I added, we see that the f_tol and x_tol criteria of samin have been satisfied. In the future, it would be good to make the default MultivariateOptimizationResults code print out just why the samin algorithm stopped, by hooking into the specified f_tol and x_tol. I don't know how to do that, at the moment.

julia> res = Optim.optimize(prob.f, prob.initial_x, fill(-100.0, 2), fill(100.0, 2), SAMIN(), Optim.Options(iterations=20000))
 * Algorithm: SAMIN
 * Starting Point: [-1.2,1.0]
 * Minimizer: [1.0170012560033472,1.0343750729202243]
 * Minimum: 2.897402e-04
 * Iterations: 5901
 * Convergence: false
   * |x - x'| ≤ 0.0e+00: false
     |x - x'| = NaN
   * |f(x) - f(x')| ≤ 0.0e+00 |f(x)|: false
     |f(x) - f(x')| = NaN |f(x)|
   * |g(x)| ≤ 0.0e+00: false
     |g(x)| = NaN
 * Objective Calls: 5901
 * Gradient Calls: 0

mcreel:
If I make fixes in my tree, will this same pull request pick up my fixed files, or the versions from when the PR was made. That is, will I need to make a new PR when the changes made here are in their final form?

pkofod:
Just push to the same branch that you made the PRs from, it will update here.

pkofod:
I'm not asking you to do this, just wanted to post this. You probably know the Schwefel function (https://www.sfu.ca/~ssurjano/schwef.html) - looks real nasty... 

```
julia> optimize(f, [0.0,0.0], [-500.0,-500.0], [500.0,500.0], SAMIN(rt = 0.9), Optim.Options(iterations=20000))
 * Algorithm: SAMIN
 * Starting Point: [0.0,0.0]
 * Minimizer: [420.96874651821395,420.96874672236663]
 * Minimum: 2.545513e-05
 * Iterations: 16101
 * Convergence: false
   * |x - x'| ≤ 0.0e+00: false 
     |x - x'| = NaN 
   * |f(x) - f(x')| ≤ 0.0e+00 |f(x)|: false
     |f(x) - f(x')| = NaN |f(x)|
   * |g(x)| ≤ 0.0e+00: false 
     |g(x)| = NaN 
 * Objective Calls: 16101
 * Gradient Calls: 0
```
unfortunately I didn't ask the PR author to implement the bounds like I did for SAMIN and Fminbox, so in ParticleSwarm the bounds are given to the method constructor (I will fix that asap), but it's interesting that the swarm quickly finds the global minimum, but very imprecisely. The sequentially lowered temperature in SAMIN really ensures convergence eventually. 

```
julia> optimize(f, [0.0,0.0], ParticleSwarm(lower=[-500.0,-500.0], upper=[500.00,500.0], n_particles=8), Optim.Options(iterations=200000))
 * Algorithm: Particle Swarm
 * Starting Point: [0.0,0.0]
 * Minimizer: [420.9687465711766,420.9687464513483]
 * Minimum: 2.545513e-05
 * Iterations: 200000
 * Convergence: false
   * |x - x'| ≤ 1.0e-32: false 
     |x - x'| = NaN 
   * |f(x) - f(x')| ≤ 1.0e-32 |f(x)|: false
     |f(x) - f(x')| = NaN |f(x)|
   * |g(x)| ≤ 1.0e-08: false 
     |g(x)| = NaN 
   * Reached Maximum Number of Iterations: true
 * Objective Calls: 1800000
 * Gradient Calls: 0
```

mcreel:
About my question, thanks, I think that this PR would work, then. About the Scwefel function, it seems that both methods are finding the solution, but unless I'm missing something, SA is faster than particle swarm, having used 16101 function evaluations versus 1800000. Why do you say that particle swarm is imprecise: it looks to me that the two methods find the same solution?

pkofod:
> Why do you say that particle swarm is imprecise: it looks to me that the two methods find the same solution?

Yeah scratch that, I think I had some unfortunate runs (it's stochastic after all) and misread the output when I increased the number of iterations before I posted it...

pkofod:
@mcreel I will merge when the tests pass .

mcreel:
Is there something that needs to be done, or is this just some problem with the testing infrastructure?

pkofod:
> Is there something that needs to be done, or is this just some problem with the testing infrastructure?

After the new tolerances, the old test seems to fail, so either we need to up the number of iterations, change the tolerance, or something else. Feel free to increase the number of iterations

mcreel:
Ah, I see that the test fails because the search width is still wide when the default number of iterations has been reached. The conservative solution would be to increase the number of iterations. As of now, samin uses the Optim options.iterations as the limit. I myself don't know how to manipulate that for the tests, as I haven't had a chance to study how the tests are done. If there's an easy way to change it, please just set it to some generous limit, like 1e6.

test/multivariate/solvers/constrained/samin.jl
```diff
@@ -4,6 +4,6 @@
     xtrue = prob.solutions
     f = OptimTestProblems.MultivariateProblems.objective(prob)
     x0 = prob.initial_x
-    res = optimize(f, x0, x0.-100., x0.+100.0, Optim.SAMIN(), Optim.Options(iterations=4000))
+    optimize(f, x0, x0.-100., x0.+100.0, Optim.SAMIN(), Optim.Options(iterations=1000000))
```
anriseth: The tests fail, as this line should be `res = optimize(...`

pkofod: That was my bad, I think I introduced it in the merge conflict resolution, I'll fix it.

PR #561, opened 2018-03-25
Arbitrary precision support

mohamed82008:
This requires modifying `LineSearches` first according to https://github.com/JuliaNLSolvers/LineSearches.jl/pull/89, but I am submitting the PR now anyways for review. More changes are also likely necessary given the flux that Optim and LineSearches master branches are in right now.

### DIFF
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -90,13 +90,14 @@ end
 # Default preconditioner for box-constrained optimization
 # This creates the inverse Hessian of the barrier penalty
 function precondprepbox!(P, x, l, u, mu)
-    @. P.diag = 1/(mu*(1/(x-l)^2 + 1/(u-x)^2) + 1)
+    T = eltype(x)
+    @. P.diag = 1/(mu*(1/max(x-l, sqrt(realmin(T)))^2 + 1/max(u-x, sqrt(realmin(T)))^2) + 1)
 end
 
 struct Fminbox{T<:AbstractOptimizer} <: AbstractOptimizer end
 Fminbox() = Fminbox{ConjugateGradient}() # default optimizer
 
-Base.summary(::Fminbox{O}) where {O} = "Fminbox with $(summary(O()))"
+Base.summary(::Fminbox{O}) where {O} = "Fminbox with $(summary(O(T)))"
 
 function optimize(obj,
                   initial_x::AbstractArray{T},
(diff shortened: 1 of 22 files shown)


src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -276,11 +277,18 @@ function optimize(
         results.x_converged, results.f_converged, results.g_converged, converged, f_increased = assess_convergence(x, xold, minimum(results), fval0, g, x_tol, f_tol, g_tol)
         f_increased && !allow_f_increases && break
     end
-    return MultivariateOptimizationResults(Fminbox{O}(), false, initial_x, minimizer(results), df.f(minimizer(results)),
+
+    Tx = typeof(x)
+    _x_abschange = maxdiff(x,xold)
```
anriseth: There is a method in the API called `x_abschange(x, x_previous)` that you can use here.  (It seems like we returned the vector 2-norm before. I assume that was a mistake (?)

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -90,7 +90,8 @@ end
 # Default preconditioner for box-constrained optimization
 # This creates the inverse Hessian of the barrier penalty
 function precondprepbox!(P, x, l, u, mu)
-    @. P.diag = 1/(mu*(1/(x-l)^2 + 1/(u-x)^2) + 1)
+    T = eltype(x)
+    @. P.diag = 1/(mu*(1/max(x-l, sqrt(realmin(T)))^2 + 1/max(u-x, sqrt(realmin(T)))^2) + 1)
```
anriseth: Did you hit an issue where `x \approx l` or something like that?

(This makes readability a bit worse, so I was wondering if this is necessary or if there are more readable ways to fix this. )

mohamed82008: I think any problem where the minimum is exactly at the boundary will be problematic. If minimizing x^2 where x >= 2 works at every precision, I would be surprised. But let me double check.

src/multivariate/solvers/constrained/fminbox.jl
mohamed82008: I thought that was a mistake yes since `default_convergence_assessment` checks if the `maxdiff` or `x_abschange` is less than the `x_tol`. I can use `x_abschange`.

mohamed82008:
I will try to add tests the coming weekend. I already tested parts in my code but I agree that more extensive tests are necessary to ensure everything is working fine.

antoine-levitt:
Those T(1)/10 are not very readable... Is T(0.1) not sufficient?

pkofod:
> Is T(0.1) not sufficient?

That'd be fine, yeah.

pkofod:
@mohamed82008 tell me if you need help here

mohamed82008:
I would appreciate the help yes. If not, I can still get around to do it, but definitely not before this coming weekend, and hopefully not after the next weekend!

pkofod:
I have to go through this, but I think I've covered most of this elsewhere... I looks like it would be a **nightmare** to fix conflicts here :

PR #564, opened 2018-03-30
Specialize on functions, reduce overhead

chriselrod:
I am tracking this PR:
https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/52

As said there, I plan to
- [X] Allow for specialization on functions. 
- [ ]  Make sure there are no closure-problems with the autodiff functions.
- [ ]  Create an optimize_light function, that simply drops a lot of the checking optimize does, and returns extremely simplified output (eg, min-val, minimizer tuple) instead of optimization results. The intention would be for that to be an API for other package developers / to be used within other functions, rather than the end user who will always prefer optimize.

Btw, my most recent benchmarks:
```julia
julia> @benchmark optimize(rosenbrock, $initial_x, $hage)
BenchmarkTools.Trial: 
  memory estimate:  13.06 KiB
  allocs estimate:  237
  --------------
  minimum time:     15.690 μs (0.00% GC)
  median time:      17.262 μs (0.00% GC)
  mean time:        26.261 μs (28.44% GC)
  maximum time:     53.599 ms (99.92% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark optimize(rosenbrock, $initial_x, $back)
BenchmarkTools.Trial: 
  memory estimate:  4.08 KiB
  allocs estimate:  78
  --------------
  minimum time:     11.923 μs (0.00% GC)
  median time:      12.794 μs (0.00% GC)
  mean time:        14.638 μs (8.33% GC)
  maximum time:     12.225 ms (99.70% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark optimize(rosenbrock, $initial_x, $ntr)
BenchmarkTools.Trial: 
  memory estimate:  54.67 KiB
  allocs estimate:  573
  --------------
  minimum time:     40.106 μs (0.00% GC)
  median time:      42.640 μs (0.00% GC)
  mean time:        57.500 μs (22.61% GC)
  maximum time:     56.389 ms (99.85% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

versus a 0.7 run with deprecation warnings turned off, when I started:
```julia
julia> @benchmark optimize(rosenbrock, $init, $hage)
BenchmarkTools.Trial: 
  memory estimate:  25.34 KiB
  allocs estimate:  862
  --------------
  minimum time:     45.316 μs (0.00% GC)
  median time:      48.131 μs (0.00% GC)
  mean time:        58.067 μs (13.32% GC)
  maximum time:     46.686 ms (99.79% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark optimize(rosenbrock, $init, $back)
BenchmarkTools.Trial: 
  memory estimate:  8.38 KiB
  allocs estimate:  238
  --------------
  minimum time:     17.914 μs (0.00% GC)
  median time:      19.176 μs (0.00% GC)
  mean time:        25.607 μs (23.08% GC)
  maximum time:     46.197 ms (99.87% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark optimize(rosenbrock, $init, $ntr)
BenchmarkTools.Trial: 
  memory estimate:  102.52 KiB
  allocs estimate:  1540
  --------------
  minimum time:     95.029 μs (0.00% GC)
  median time:      108.193 μs (0.00% GC)
  mean time:        124.683 μs (13.10% GC)
  maximum time:     48.850 ms (99.61% GC)
  --------------
  samples:          10000
  evals/sample:     1
```


antoine-levitt:
If I understand correctly this uses copies for complex_to_real? I don't think that will do the right thing, because we rely on the fact we can modify the real view of the array. The tests should cover that, hopefully. I have no good solution except to let the complex code be slow until `reinterpret` is fixed.

chriselrod:
You are right. The tests do cover that.
I was getting a test error, because the type was wrong on 0.7 -- a reinterpreted array instead of a regular array.
That was my hack, without looking into how they worked.

Thankfully, tests still failed because the answers were wrong.

pkofod:
Out of curiosity ,what kind of overhead are you seeing in your application?

chriselrod:
Do you mean these tests?
They're simply:
```julia
using Optim
rosenbrock(x) =  (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
result = optimize(rosenbrock, zeros(2), BFGS())
```
from the readme.

For my own application, evaluating `fg` takes about 1 microsecond, so `Optim` makes a significant chunk of the runtime.

pkofod:
Alright, so it's for the super fast problems. I'm 100% on board having both choices, I was just curious

PR #568, opened 2018-04-06
Native complex

antoine-levitt:
(LineSearches PR incoming)

This adds a bunch of `real` and `'` to accept complex arrays into the first-order optimizers. To test, simply set complex_to_real and real_to_complex to no-ops in NLSolversBase. I tested the gradient descents, CG, LBFGS and BFGS, all of which behave nearly identically to the previous version. @anriseth the code for ngmres looks daunting, and I'm afraid I'm going to screw up, could you help in porting the functionality? All that's needed should be a couple type annotations, real()s and `'` on outer products.

### DIFF
--- a/src/multivariate/solvers/first_order/bfgs.jl
+++ b/src/multivariate/solvers/first_order/bfgs.jl
@@ -69,13 +69,13 @@ function initial_state(method::BFGS, options, d, initial_x::AbstractArray{T}) wh
     BFGSState(initial_x, # Maintain current state in state.x
               similar(initial_x), # Maintain previous state in state.x_previous
               copy(gradient(d)), # Store previous gradient in state.g_previous
-              T(NaN), # Store previous f in state.f_x_previous
+              real(T)(NaN), # Store previous f in state.f_x_previous
               similar(initial_x), # Store changes in position in state.dx
               similar(initial_x), # Store changes in gradient in state.dg
               similar(initial_x), # Buffer stored in state.u
               method.initial_invH(initial_x), # Store current invH in state.invH
               similar(initial_x), # Store current search direction in state.s
-              @initial_linesearch()...) # Maintain a cache for line search results in state.lsr
+              @initial_linesearch()...)
 end
 
 
--- a/src/multivariate/solvers/first_order/bfgs.jl
+++ b/src/multivariate/solvers/first_order/bfgs.jl
@@ -59,11 +59,11 @@ end
 function initial_state(method::BFGS, options, d, initial_x::AbstractArray{T}) where T
     n = length(initial_x)
     initial_x = copy(initial_x)
-    retract!(method.manifold, real_to_complex(d,initial_x))
+    retract!(method.manifold, initial_x)
 
     value_gradient!!(d, initial_x)
 
-    project_tangent!(method.manifold, real_to_complex(d,gradient(d)), real_to_complex(d,initial_x))
+    project_tangent!(method.manifold, gradient(d), initial_x)
     # Maintain a cache for line search results
     # Trace the history of states visited
     BFGSState(initial_x, # Maintain current state in state.x
@@ -86,7 +86,7 @@ function update_state!(d, state::BFGSState, method::BFGS)
     # Search direction is the negative gradient divided by the approximate Hessian
     A_mul_B!(vec(state.s), state.invH, vec(gradient(d)))
     scale!(state.s, -1)
-    project_tangent!(method.manifold, real_to_complex(d,state.s), real_to_complex(d,state.x))
+    project_tangent!(method.manifold, state.s, state.x)
 
     # Maintain a record of the previous gradient
     copy!(state.g_previous, gradient(d))
@@ -99,7 +99,7 @@ function update_state!(d, state::BFGSState, method::BFGS)
     # Update current position
     state.dx .= state.alpha.*state.s
     state.x .= state.x .+ state.dx
-    retract!(method.manifold, real_to_complex(d,state.x))
+    retract!(method.manifold, state.x)
 
     lssuccess == false # break on linesearch error
 end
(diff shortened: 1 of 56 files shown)


anriseth:
> @anriseth the code for ngmres looks daunting, and I'm afraid I'm going to screw up, could you help in porting the functionality? All that's needed should be a couple type annotations, real()s and ' on outer products.

I'll have a look :p 


> To test, simply set complex_to_real and real_to_complex to no-ops in NLSolversBase.

Are you planning to PR this, or is it better for @chriselrod to do it in https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/51 ?

anriseth:
@mohamed82008 when this merges you may want to ensure that the arbitrary floating point types you are using support `real(x) = x` (and `'`)

src/multivariate/solvers/first_order/bfgs.jl
```diff
@@ -110,20 +110,20 @@ function update_h!(d, state, method::BFGS)
     state.dg .= gradient(d) .- state.g_previous
 
     # Update the inverse Hessian approximation using Sherman-Morrison
-    dx_dg = vecdot(state.dx, state.dg)
+    dx_dg = real(vecdot(state.dx, state.dg))
```
anriseth: How come it is sufficient to get the real part from `vecdot`?

antoine-levitt: It doesn't actually matter: dx is related to dg by a hermitian matrix, so that this dot product is real. I just checked, it's actually 1e-8 in the first iterations, for some reason I don't quite understand

mohamed82008:
> when this merges you may want to ensure that the arbitrary floating point types you are using support real(x) = x (and ')

I will double check, thanks. Worst case I will locally define the functions I need in my branch ;)

antoine-levitt:
Should not be necessary: Base defines real(x::Real) = x and conj(x::Real) = x

src/multivariate/solvers/first_order/bfgs.jl
pkofod: > I just checked, it's actually 1e-8 in the first iterations, for some reason I don't quite understand

what do you mean by this?

antoine-levitt: Simply that : from what I understand of the algorithm it should always be zero, but when I println it during the iterations it starts at 1e-8 and goes down from there. Maybe that's specific to the problem I'm looking at though (holomorphic gradient)... Will investigate more.

pkofod:
@antoine-levitt sorry if I'm daft, but I'm not really sure how this is supposed to go down. What's the plan exactly? Merge this, merge the LineSearch pr and then what?

antoine-levitt:
OK I played around with it, including objective functions with non-holomorphic gradients, and everything seems to work out well: the complex version of the algorithms are sensibly identical to their real counterpart, modulo numerical noise. The real(vecdot()) vs vecdot() has no effect. It's kind of hard to know exactly if they are truly equivalent, but I'm satisfied that at least morally it should not matter, so... good enough for now I guess.

Re merge, the plan was to merge those, which are only non-functional changes since the complex_to_real wrappers are still in place, and then I do a followup PR removing those wrappers. But I can do everything at once if it's more convenient for you.

pkofod:
> Re merge, the plan was to merge those, which are only non-functional changes since the complex_to_real wrappers are still in place, and then I do a followup PR removing those wrappers.

This would be perfect. Let me know when you're done working onthe PRs and I'll look at them and merge if everything is :+1:  . Regarding real of vecdot or not, is this simply because you get a complex number out (I mean the type) but it is really a real number, or?

antoine-levitt:
well this is awkward <_< these real() that I thought had no effect on BFGS? Actually without them BFGS does not converge in the manifolds tests. There's something I don't understand well yet in there, but with this code everything works fine, so...

antoine-levitt:
(CI fails because it needs a synchronized NLSolversBase)

pkofod:
I'll get the rest of NLSolversBase taggable and then we can try to get this one merges asap

pkofod:
> OK I played around with it, including objective functions with non-holomorphic gradients, a

are these functions already tests here? if not, would you please add them here so we can include them in tests and benchmarks?

antoine-levitt:
Done. I also revamped the test a bit, and also added some comparison code vs the real-ified solvers in comments. Some solvers are identical (GD,CG), some have small differences (LBFGS, OACCEL, NGMRES), BFGS has sizeable differences (but it's not easy to tell which of the real and complex is more efficient). This is interesting theoretically, but I don't have time to dig further.

test/multivariate/manifolds.jl
```diff
@@ -17,7 +17,7 @@
     for method in (Optim.GradientDescent, Optim.ConjugateGradient, Optim.LBFGS, Optim.BFGS,
                    Optim.NGMRES, Optim.OACCEL)
         debug_printing && print_with_color(:green, "Solver: $(summary(method()))\n")
-        res = Optim.optimize(fmanif, gmanif!, x0, method(manifold=manif))
+        res = Optim.optimize(fmanif, gmanif!, x0, method(manifold=manif), Optim.Options(allow_f_increases=false))
```
anriseth: I'm surprised NGMRES and OACCEL converge with `allow_f_increases=false` :o

antoine-levitt: Sorry, that shouldn't have been there, I reverted it.

pkofod:
This is line search related, I'll fix it with a PR to your branch

pkofod:
@antoine-levitt 

Can I get you to 
```
git fetch origin
```
(or whatever your remote is called that references JuliaNLSolvers)
and then
```
git cherrypick c6a7e50119d3156e0337f87fccd5c262fc628a47
```
to pull in a fix to the current failure? thanks

pkofod:
Good to go right?

antoine-levitt:
Yup!

PR #569, opened 2018-04-06
Change default tolerances, and document relative/absolute inf norm

antoine-levitt:


docs/src/algo/cg.md
```diff
@@ -35,9 +35,9 @@ Results of Optimization Algorithm
  * Minimum: 5.144946e-18
  * Iterations: 21
  * Convergence: true
-   * |x - x'| < 1.0e-32: false
+   * |x - x'| < 0.0e+00: false
```
anriseth: The bounds have been turned into less-than-or-equal now, and it seems that we forgot to update the documentation.

pkofod:
Besides the comments by @anriseth this looks fine to me!

pkofod:
You don't have to do this now, but I've even been wondering if we should show these by default at all. Since they default to 0 now, we could change the printing to only be done if you set it to something different.

anriseth:
> You don't have to do this now, but I've even been wondering if we should show these by default at all. Since they default to 0 now, we could change the printing to only be done if you set it to something different.

There's a corner case to take into account if we do this: If the x or f differences are exactly zero, then Optim still considers this convergence.

pkofod:
> There's a corner case to take into account if we do this: If the x or f differences are exactly zero, then Optim still considers this convergence.

That is indeed a good point.

PR #582, opened 2018-04-10
Update docs

anriseth:


pkofod:
Maybe we should remove the "latest" badge at the top (in the readme)? This seems to be a great deal of confusion to users. They come to the read me, click "latest docs" because why would you want old docs? And then they're not sure why it doesn't work on their METADATA version of Optim. It'll still be mentioned at the bottom of the readme as it is now.

anriseth:
merge-ready?

anriseth:
Thanks!

Travis does not seem to have updated http://julianlsolvers.github.io/Optim.jl/latest/ ? It hasn't updated with the changed LICENSE that you added several days ago either ...

PR #583, opened 2018-04-10
Fix ParticleSwarm for Fminbox

anriseth:
Fminbox adds linesearch parameters to all solvers that are not explicitly listed. This means that ParticleSwarm and the trust region crash with an error.

@pkofod why is Newton not supported in Fminbox? 
We should also decide how to handle NewtonTrustRegion and KrylovTrustRegion. 
Things to consider:
- Error and say they are not supported 
- Include them in the list together with ParticleSwarm etc. 
- Add a try-catch block around the line that tries to call `optimizer(linesearch=...)` and error with a message saying something along the lines of "$optimizer is not supported with Fminbox. If you think it should be, report it on Github"

### DIFF
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -244,7 +244,7 @@ function optimize(
                 alphaguess = LineSearches.InitialPrevious()
             end
             _optimizer = O(alphaguess = alphaguess, linesearch = linesearch, P = P, precondprep = pcp)
-        elseif O in (NelderMead, SimulatedAnnealing)
+        elseif O in (NelderMead, SimulatedAnnealing, ParticleSwarm)
             _optimizer = O()
         else
             if linesearch == nothing


anriseth:
@iwelch can you verify whether this fixes your problems with ParticleSwarm? 
In future, can you please provide error messages (and preferably minimal working examples) when you report problems?

pkofod:
A tiny test perhaps? Other than that, it looks good to me.

pkofod:
> @pkofod why is Newton not supported in Fminbox?

Because no one sat down and wrote the barrier hessian I guess? Don't know, I rarely use this part of the package, which is probably also where there are a few rough edges.

anriseth:
> A tiny test perhaps? Other than that, it looks good to me.

Yes, something along the lines of looping through all the solvers and checking that they work/error in a sensible way?

pkofod:
> Yes, something along the lines of looping through all the solvers and checking that they work/error in a sensible way?

Yes, something like that.

Issue #542, opened 2018-03-01
Fminbox with Optim.Options gives error


simonbyrne:
```
using Optim
f(x) = dot(x,x)
Optim.optimize(OnceDifferentiable(f, [1.0,2.0], autodiff=:forward),
               [1.0,2.0], [0.0,-Inf], [Inf,Inf],
               Fminbox{BFGS}(), Optim.Options(x_tol=1e-4))
```

```
julia> Optim.optimize(OnceDifferentiable(f, [1.0,2.0], autodiff=:forward),
                      [1.0,2.0], [0.0,-Inf], [Inf,Inf],
                      Fminbox{BFGS}(), Optim.Options(x_tol=1e-4))
ERROR: No default objective type for Optim.Fminbox{Optim.BFGS}() and (NLSolversBase.OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1},Val{false}}(f, Optim.#6, Optim.#7, 0.0, [NaN, NaN], [NaN, NaN], [NaN, NaN], [0], [0]), [1.0, 2.0], [0.0, -Inf]).
Stacktrace:
 [1] optimize at /Users/simon/.julia/v0.6/Optim/src/multivariate/optimize/interface.jl:80 [inlined]
 [2] optimize(::NLSolversBase.OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1},Val{false}}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Optim.Fminbox{Optim.BFGS}, ::Optim.Options{Float64,Void}) at /Users/simon/.julia/v0.6/Optim/src/multivariate/optimize/interface.jl:77
```

mohamed82008:
Seems like a dispatch problem. This works:
```julia
julia> Optim.optimize(OnceDifferentiable(f, [1.0,2.0], autodiff=:forward),
                      [1.0,2.0], [0.0,-Inf], [Inf,Inf],
                      Fminbox{BFGS}(), optimizer_o=Optim.Options(x_tol=1e-4))
Results of Optimization Algorithm
 * Algorithm: Fminbox with BFGS
 * Starting Point: [1.0,2.0]
 * Minimizer: [1.732490485369608e-14,-9.553697010690142e-24, ...]
 * Minimum: 1.931213e-26
 * Iterations: 7
 * Convergence: true
   * |x - x'| ≤ 1.0e-04: false
     |x - x'| = 2.90e-06
[...]

PR #584, opened 2018-04-10
Update Fminbox to better API

pkofod:
I guess ref #583 (I can take the tests here if you just close that PR, or you can PR your tests against this branch)

also ref #542 

The idea here is to have `Fminbox` accept the method  instance

```
Fminbox(ConjugateGradient())
```
or 
```
Fminbox(BFGS(); mu0 = ..., mufactor = ..., precondprep = ...)
```
(I know this current version has a keyword, but that'll go)

I also moved the weird way of specifying the options into `Options`, so the API is now like the rest of the package.

I also plan to have
```
optimize(f, [g, h], x, lb, ub, method, ...)
```
where method might be `BFGS()` or similar dispatch to
```
optimize(f, [g, h], x, lb, ub, Fminbox(method), ...)
```
that is if bounds are given, it will default to Fminbox (the only possibility we have right now, but other methods could be implemented).

edit: don't mind the weird first commit...

### DIFF
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -1,7 +1,7 @@
 # Attempt to compute a reasonable default mu: at the starting
 # position, the gradient of the input function should dominate the
 # gradient of the barrier.
-function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}; mu0::T = convert(T, NaN), mu0factor::T = 0.001) where T
+function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0factor::T = 0.001, mu0::T = convert(T, NaN)) where T
     if isnan(mu0)
         gbarriernorm = sum(abs, gbarrier)
         if gbarriernorm > 0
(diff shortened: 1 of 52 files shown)


anriseth:
> I guess ref #583 (I can take the tests here if you just close that PR, or you can PR your tests against this branch)

I'll close the PR. It is probably easier to keep it all within one PR, as there are several WIP PRs to deal with already.

src/types.jl
```diff
@@ -29,12 +34,17 @@ function Options(;
...
+        outer_f_tol::Real = 0.0,
+        outer_g_tol::Real = 1e-8,
         f_calls_limit::Int = 0,
         g_calls_limit::Int = 0,
         h_calls_limit::Int = 0,
         allow_f_increases::Bool = false,
+        allow_outer_f_increases::Bool = false,
         successive_f_tol::Int = 0,
         iterations::Integer = 1_000,
+        outer_iterations = 1000,
```
anriseth: `outer_iterations::Int` ? 
(And should we change `iterations::Integer` to `::Int` as well?)

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -179,10 +198,14 @@ function optimize(
         warn("Initial position cannot be on the boundary of the box. Moving elements to the interior.\nElement indices affected: $boundaryidx")
     end
     df.df(gfunc, x)
```
anriseth: Let's update this to `gradient!(df, x) ; gfunc .= gradient(df)`? (Or however is best to use the NLSolversBase interface here)

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -90,14 +90,44 @@ end
...
 function precondprepbox!(P, x, l, u, mu)
-    @. P.diag = 1/(mu*(1/(x-l)^2 + 1/(u-x)^2) + 1)
+    @. P.diag = 1/(mu[1]*(1/(x-l)^2 + 1/(u-x)^2) + 1)
 end
 
-struct Fminbox{T<:AbstractOptimizer} <: AbstractOptimizer end
-Fminbox() = Fminbox{ConjugateGradient}() # default optimizer
+struct Fminbox{T} <: AbstractOptimizer
+    inner_optimizer::T
+    mu0
```
mohamed82008: Type unstable?

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -196,66 +219,20 @@ function optimize(
...
-    while !converged && iteration < iterations
+
+    while !converged && iteration < outer_iterations
         # Increment the number of steps we've had to perform
         iteration += 1
 
         copy!(xold, x)
         # Optimize with current setting of mu
-        funcc = (g, x) -> barrier_combined(gfunc, gbarrier,  g, x, fb, mu)
+        funcc = (g, x) -> barrier_combined(gfunc, gbarrier,  g, x, fb, mu[1])
```
mohamed82008: May also trigger the closure bug.

src/types.jl
mohamed82008: If it is a single number, I am not sure it is worth supporting all sizes of integers, unless there is an application with more iterations than 9,223,372,036,854,775,807.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -179,10 +198,14 @@ function optimize(
         warn("Initial position cannot be on the boundary of the box. Moving elements to the interior.\nElement indices affected: $boundaryidx")
     end
     df.df(gfunc, x)
-    mu = isnan(mu0) ? initial_mu(gfunc, gbarrier; mu0factor=mufactor) : mu0
+    mu = isnan(mu0) ? [initial_mu(gfunc, gbarrier; mu0factor=mufactor)] : [mu0]
```
mohamed82008: Why is `mu` an array?

src/types.jl
pkofod: It really doesn't matter... "worth" is without meaning when it's free, but I can change it to keep consistence.

src/multivariate/solvers/constrained/fminbox.jl
pkofod: Good point

src/multivariate/solvers/constrained/fminbox.jl
pkofod: yes, this is WIP :)

src/multivariate/solvers/constrained/fminbox.jl
pkofod: If you look at the changes you'll realize it is to avoid redefining `pcp` and thus `_optimizer` on every iteration. I think it's significantly simpler to do it like this compared to defining a type and call overloading it. However, if the closure bug (which seriously is the biggest gripe in Julia to me) has bite here, then it might be worth it to do the complicated thing.

mohamed82008: Hmm, I see. A `Base.RefValue` would probably work better then which is created using `Ref(mu0)` for example. I am not 100% sure the closure bug will bite, but I have become somewhat closure-phobic over the last few months.

pkofod: A red could work, and I think I’ve used that elsewhere, so maybe for consistency I should do that. I really think it’s annoying that the closure bug is still around. We got the first perf regression issue here at optim just after v0.5.0 hit , and I really hoped it would be fixed in v0.5.1, but...

mohamed82008: Ya, it's annoying but since it's a bug fix, I think it may be done in the 1.0 cycle.

anriseth:
Is there a way to add the `inplace` and `autodiff` keywords to this new interface? (Just wondering whether there will be too many ambiguities for dispatch)

PR #596, opened 2018-04-20
Move updates

pkofod:
These functions where made to fix some of the problems now handled in *Differentiable, but weren't removed when we updated. I like it much better that `update_state!` clearly shows when g, etc is updated.

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -1,6 +1,3 @@
-# Update the Hessian
-update_h!(d, state, method) = nothing
-update_h!(d, state, method::SecondOrderOptimizer) = hessian!(d, state.x)
 
 after_while!(d, state, method, options) = nothing
 
@@ -44,7 +41,7 @@ function optimize(d::D, initial_x::Tx, method::M,
         # TODO: Do the same for x_tol?
         counter_f_tol = f_converged ? counter_f_tol+1 : 0
         converged = converged | (counter_f_tol > options.successive_f_tol)
-        
+
         if tracing
             # update trace; callbacks can stop routine early by returning true
             stopped_by_callback = trace!(tr, d, state, iteration, method, options)
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -1,12 +1,3 @@
-update_g!(d, state, method) = nothing
-function update_g!(d, state, method::M) where M<:Union{FirstOrderOptimizer, Newton}
-    # Update the function value and gradient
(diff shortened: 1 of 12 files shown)


pkofod:
No , that’s certainly not the intention. Why would that happen?

src/multivariate/optimize/optimize.jl
```diff
@@ -47,16 +35,13 @@ function optimize(d::D, initial_x::Tx, method::M,
...
         update_state!(d, state, method) && break # it returns true if it's forced by something in update! to stop (eg dx_dg == 0.0 in BFGS, or linesearch errors)
-        update_g!(d, state, method) # TODO: Should this be `update_fg!`?
         x_converged, f_converged,
         g_converged, converged, f_increased = assess_convergence(state, d, options)
         # For some problems it may be useful to require `f_converged` to be hit multiple times
         # TODO: Do the same for x_tol?
         counter_f_tol = f_converged ? counter_f_tol+1 : 0
         converged = converged | (counter_f_tol > options.successive_f_tol)
 
-        !converged && update_h!(d, state, method) # only relevant if not converged
```
anriseth: Previously, update_h was only called if `converged = false`.

src/multivariate/solvers/first_order/bfgs.jl
```diff
@@ -100,15 +100,18 @@ function update_state!(d, state::BFGSState, method::BFGS)
     state.dx .= state.alpha.*state.s
     state.x .= state.x .+ state.dx
     retract!(method.manifold, state.x)
+    value_gradient!(d, state.x)
+
+    bfgs_update!(d, state)
```
anriseth: Now, update_h is called before we do the conference check, so we do an unnecessary update at the final iteration (when converged = true)

pkofod: Ah, okay, now I get it, I thought you meant during all iterations. In the final iteration, yes. We could maybe update these at the beginning, although that would require us to have an ugly "if first iteration" type of branch.

Maybe it's okay to have it in the main loop after all, and I should just replace it with `value_gradient!` calls.

anriseth: I'm not sure what the best solution is. It shouldn't be that costly to do `{l,}bfgs_update`, so it may not be worth the hassle. Hessian evaluations in Newton (et al.) may be more costly, but maybe it's worth it if the new approach makes it easier to maintain the code and add new features.

pkofod:
I'm not sure this is wanted after all.. Thanks for taking part @anriseth

PR #605, opened 2018-05-26
Fminbox enhancements

alyst:
Miscellaneous code enhancements/cleanups, mostly around Fminbox.

### DIFF
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -48,12 +48,12 @@ function function_barrier(gfunc, gbarrier, x::AbstractArray,
 end
 
 function barrier_combined(gfunc, gbarrier, g, x::AbstractArray,
-                          fb::Function, mu::Real)
+                          fb::Function, mu::Ref{<:Real})
     valfunc, valbarrier = fb(gbarrier, x, gfunc)
     if g !== nothing
-        g .= gfunc .+ mu.*gbarrier
+        g .= gfunc .+ mu[].*gbarrier
     end
-    return convert(eltype(x), valfunc + mu*valbarrier) # FIXME make this unnecessary
+    return convert(eltype(x), valfunc + mu[]*valbarrier) # FIXME make this unnecessary
 end
 
 function limits_box(x::AbstractArray{T}, d::AbstractArray{T},
@@ -204,6 +204,12 @@ function optimize(
     # Count the total number of outer iterations
     iteration = 0
 
+    # define the function (dfbox) to optimize by the inner optimizer
+    funcc = (g, x) -> barrier_combined(gfunc, gbarrier, g, x, fb, mu)
+    dfbox = OnceDifferentiable(x -> funcc(nothing, x),
+                               (g, x) -> (funcc(g, x); g),
+                               funcc, initial_x, zero(T))
+
     xold = similar(x)
     converged = false
     local results
@@ -216,9 +222,7 @@ function optimize(
 
         copy!(xold, x)
         # Optimize with current setting of mu
-        funcc = (g, x) -> barrier_combined(gfunc, gbarrier,  g, x, fb, mu[])
         fval0 = funcc(nothing, x)
-        dfbox = OnceDifferentiable(x->funcc(nothing, x), (g, x)->(funcc(g, x); g), funcc, initial_x, zero(T))
         if show_trace > 0
             println("#### Fminbox #$iteration: Calling optimizer with mu = ", mu[], " ####")
         end
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -59,8 +59,8 @@ end
 function limits_box(x::AbstractArray{T}, d::AbstractArray{T},
                     l::AbstractArray{T}, u::AbstractArray{T}) where T
     alphamax = convert(T, Inf)
-    @simd for i in eachindex(x)
-        @inbounds if d[i] < 0
+    @inbounds for i in eachindex(x)
+        if d[i] < 0
             alphamax = min(alphamax, ((l[i]-x[i])+eps(l[i]))/d[i])
         elseif d[i] > 0
             alphamax = min(alphamax, ((u[i]-x[i])-eps(u[i]))/d[i])
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
 end
 
 function barrier_box(g, x::AbstractArray{T}, l::AbstractArray{T}, u::AbstractArray{T}) where T
-    n = length(x)
-    calc_g = !(g === nothing)
+    calc_g = g !== nothing
 
     v = zero(T)
-    for i = 1:n
+    @inbounds for i in eachindex(x)
         thisl = l[i]
         if isfinite(thisl)
             dx = x[i] - thisl
-            if dx <= 0
-                return convert(T, Inf)
-            end
+            (dx <= zero(T)) && return convert(T, Inf)
             v -= log(dx)
-            if calc_g
-                g[i] = -one(T)/dx
-            end
+            calc_g && (g[i] = -one(T)/dx)
         else
-            if calc_g
-                g[i] = zero(T)
-            end
+            calc_g && (g[i] = zero(T))
         end
         thisu = u[i]
         if isfinite(thisu)
             dx = thisu - x[i]
-            if dx <= 0
-                return convert(T, Inf)
-            end
+            (dx <= zero(T)) && return convert(T, Inf)
             v -= log(dx)
-            if calc_g
-                g[i] += one(T)/dx
-            end
+            calc_g && (g[i] += one(T)/dx)
         end
     end
     return v
 end
 
-function function_barrier(gfunc, gbarrier, x::AbstractArray{T}, f::F, fbarrier::FB) where {T, F<:Function, FB<:Function}
+function function_barrier(gfunc, gbarrier, x::AbstractArray,
+                          f::Function, fbarrier::Function)
     vbarrier = fbarrier(gbarrier, x)
-    if isfinite(vbarrier)
-        vfunc = f(gfunc, x)
-    else
-        vfunc = vbarrier
-    end
-    return vfunc, vbarrier
+    return (isfinite(vbarrier) ? f(gfunc, x) : vbarrier), vbarrier
 end
 
-function barrier_combined(gfunc, gbarrier, g, x::AbstractArray{T}, fb::FB, mu::T) where {T, FB<:Function}
-    calc_g = !(g === nothing)
+function barrier_combined(gfunc, gbarrier, g, x::AbstractArray,
+                          fb::Function, mu::Real)
     valfunc, valbarrier = fb(gbarrier, x, gfunc)
-    if calc_g
+    if g !== nothing
         g .= gfunc .+ mu.*gbarrier
     end
-    return convert(T, valfunc + mu*valbarrier) # FIXME make this unnecessary
+    return convert(eltype(x), valfunc + mu*valbarrier) # FIXME make this unnecessary
 end
 
-function limits_box(x::AbstractArray{T}, d::AbstractArray{T}, l::AbstractArray{T}, u::AbstractArray{T}) where T
+function limits_box(x::AbstractArray{T}, d::AbstractArray{T},
+                    l::AbstractArray{T}, u::AbstractArray{T}) where T
     alphamax = convert(T, Inf)
-    for i = 1:length(x)
-        if d[i] < 0
-            @inbounds alphamax = min(alphamax, ((l[i]-x[i])+eps(l[i]))/d[i])
+    @simd for i in eachindex(x)
+        @inbounds if d[i] < 0
+            alphamax = min(alphamax, ((l[i]-x[i])+eps(l[i]))/d[i])
         elseif d[i] > 0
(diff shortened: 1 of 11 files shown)


pkofod:
Thanks for showing Fminbox some attention! I'll return and review later.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
-            end
+            calc_g && (g[i] += one(T)/dx)
         end
     end
     return v
 end
 
-function function_barrier(gfunc, gbarrier, x::AbstractArray{T}, f::F, fbarrier::FB) where {T, F<:Function, FB<:Function}
+function function_barrier(gfunc, gbarrier, x::AbstractArray,
+                          f::Function, fbarrier::Function)
```
pkofod: This is *old* code. It shouldn't even annotate  `Function` here, as it is overly specific.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
             dx = x[i] - thisl
-            if dx <= 0
-                return convert(T, Inf)
-            end
+            (dx <= zero(T)) && return convert(T, Inf)
             v -= log(dx)
-            if calc_g
-                g[i] = -one(T)/dx
-            end
+            calc_g && (g[i] = -one(T)/dx)
```
pkofod: I'm not really sure I love this syntax. I know it's valid, but I'd prefer the old one. It's okay for the return above, but I think it's slightly harder to read when it's mutation.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
         else
-            if calc_g
-                g[i] = zero(T)
-            end
+            calc_g && (g[i] = zero(T))
```
pkofod: same as above

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
             dx = thisu - x[i]
-            if dx <= 0
-                return convert(T, Inf)
-            end
+            (dx <= zero(T)) && return convert(T, Inf)
             v -= log(dx)
-            if calc_g
-                g[i] += one(T)/dx
-            end
+            calc_g && (g[i] += one(T)/dx)
```
pkofod: same as above

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
-        vfunc = vbarrier
-    end
-    return vfunc, vbarrier
+    return (isfinite(vbarrier) ? f(gfunc, x) : vbarrier), vbarrier
 end
 
-function barrier_combined(gfunc, gbarrier, g, x::AbstractArray{T}, fb::FB, mu::T) where {T, FB<:Function}
-    calc_g = !(g === nothing)
+function barrier_combined(gfunc, gbarrier, g, x::AbstractArray,
+                          fb::Function, mu::Ref{<:Real})
```
pkofod: feel free to remove the `Function` annotation

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
 end
 
-function limits_box(x::AbstractArray{T}, d::AbstractArray{T}, l::AbstractArray{T}, u::AbstractArray{T}) where T
+function limits_box(x::AbstractArray{T}, d::AbstractArray{T},
+                    l::AbstractArray{T}, u::AbstractArray{T}) where T
     alphamax = convert(T, Inf)
-    for i = 1:length(x)
-        if d[i] < 0
-            @inbounds alphamax = min(alphamax, ((l[i]-x[i])+eps(l[i]))/d[i])
+    @simd for i in eachindex(x)
```
pkofod: Does the `@simd` annotation really do anything here? Julia (or llvm I guess) seems to be pretty good at doing this on its own. Note `@simd` annotations can sometimes slow code down, so unless you've benchmarked a specific improvement, I think it's safer to avoid it.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -17,67 +17,53 @@ function initial_mu(gfunc::AbstractArray{T}, gbarrier::AbstractArray{T}, mu0fact
...
+        @inbounds if d[i] < 0
```
pkofod: why not place this annotation where outside the `for` loop?

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -93,47 +79,53 @@ function precondprepbox!(P, x, l, u, mu)
     @. P.diag = 1/(mu[]*(1/(x-l)^2 + 1/(u-x)^2) + 1)
 end
 
-struct Fminbox{T, Tf, P} <: AbstractConstrainedOptimizer
-    method::T
-    mu0::Tf
-    mufactor::Tf
+struct Fminbox{O<:AbstractOptimizer, T, P<:Function} <: AbstractConstrainedOptimizer
```
pkofod: why `P<:Function`? That will rule out callable types etc. This should not be here. The `O<:`  is fine.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -93,47 +79,53 @@ function precondprepbox!(P, x, l, u, mu)
...
-        mu0 = NaN,
-        mufactor = 0.001,
-        precondprep = (P, x, l, u, mu) -> precondprepbox!(P, x, l, u, mu))
 
-Fminbox(method, promote(mu0, mufactor)..., precondprep) # default optimizer
+function Fminbox(method::AbstractOptimizer = LBFGS();
+                 mu0::Real = NaN, mufactor::Real = 0.001,
+                 precondprep = (P, x, l, u, mu) -> precondprepbox!(P, x, l, u, mu))
+    if method isa Newton
+        warn("Newton is not supported as the Fminbox optimizer. Defaulting to LBFGS.")
```
pkofod: I'm not sure this shouldn't just be an error instead. If you are in this branch you specifically(!) asked for `Newton`, so I think it's safer to make a full stop, and then have the user choose an appropriate method.

I know you didn't *introduce* this code, but simply moved it, so I'll just open an issue instead https://github.com/JuliaNLSolvers/Optim.jl/issues/606 .

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -174,8 +165,8 @@ function optimize(
     # initialization only makes use of the magnitude, we can fix this
     # by using the sum of the absolute values of the contributions
     # from each edge.
-    boundaryidx = Array{Int,1}()
-    for i = 1:length(gbarrier)
+    boundaryidx = Vector{Int}()
+    for i in eachindex(gbarrier)
```
pkofod: Is this necessarily better when you're indexing into `x`, `l` and `u` as well? I think this should be `eachindex(gbarrier, x, l, u)` if you want to do it like this.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -203,20 +194,27 @@ function optimize(
...
 
     g = similar(x)
-    fval_all = Array{Vector{T}}(0)
+    fval_all = Vector{Vector{T}}(0)
 
     # Count the total number of outer iterations
     iteration = 0
 
+    # define the function (dfbox) to optimize by the inner optimizer
+    funcc = (g, x) -> barrier_combined(gfunc, gbarrier, g, x, fb, mu)
```
pkofod: Can be moved outside because `mu` is a ref, right? I think it used to be constructed every time because `mu` used to be a number.

src/utilities/perform_linesearch.jl
```diff
@@ -1,34 +1,31 @@
...
-    # If invH is not positive definite, reset it
-    dphi0 = real(vecdot(gradient(d), state.s))
-    if dphi0 >= zero(dphi0)
-        # "reset" Hessian approximation
-        if M <: BFGS
-            copy!(state.invH, method.initial_invH(state.x))
-        elseif M <: LBFGS
-            state.pseudo_iteration = 1
-        end
+# Reset the search direction if it becomes corrupted
```
pkofod: @anriseth what do you think of these changes in `perform_linesearch.jl`?

src/multivariate/solvers/constrained/fminbox.jl
anriseth: I also like the old notation better.

src/utilities/perform_linesearch.jl
anriseth: I like the introduction of `reset_search_direction`. 

It may be nice to keep `checked_dphi0`, in case any future solvers would like to use it in a slightly different way than "calculate dphi0 and reset the search direction if something is wrong".
It adds a little bit of complexity / modularity, so it may not be worth keeping `checked_dphi0`

(By keep it I mean to default all methods to
```jl
function checked_dphi0(state, d, method)
   dphi_0 = real(vecdot(gradient(d), state.s))
   # reset the direction if it becomes corrupted
￼   if dphi_0 >= zero(dphi_0) && reset_search_direction!(state, d, method)
￼       dphi_0 = real(vecdot(gradient(d), state.s)) # update after direction reset
￼   end
   return dphi_0
end
```

src/multivariate/solvers/constrained/fminbox.jl
alyst: My idea is that type annotation helps autodocumenting the code and detecting argument mismatch bugs early.
Is there a way to express in v0.7 that `f` could be either function or callable object?

src/multivariate/solvers/constrained/fminbox.jl
alyst: Ok, I'll revert it. Personally, I prefer the terse variant as I better see the code structure.

src/multivariate/solvers/constrained/fminbox.jl
pkofod: No, I don't think there is unfortunately. There are hacks that will get you most of the way, but...

I'd rather have a good doc string than an unnecessary type annotation to be honest.

pkofod: @ChrisRackauckas I'm asking because I know you have faced the same question. There's no "generically callable" annotation, right?

src/multivariate/solvers/constrained/fminbox.jl
alyst: It should result in the same LLVM code, `eachindex` is just more "idiomatic". I'll update it with your suggestion.

src/multivariate/solvers/constrained/fminbox.jl
alyst: Exactly, I indicated it in the commit message.

src/utilities/perform_linesearch.jl
alyst: I would prefer not to introduce `checked_dphi0()` right now.
When it would be really necessary, it would be more clear how to properly modularize the code.

src/multivariate/solvers/constrained/fminbox.jl
fredrikekre: Base uses 
```julia
julia> Base.Callable
Union{Function, Type}
```
in a couple of places, but that rules out objects etc, so perhaps better to leave it untyped if possible.

pkofod: > so perhaps better to leave it untyped if possible.

:+1:

src/multivariate/solvers/constrained/fminbox.jl
pkofod: Right. I remember putting this on a mental "todo"-list, but forgot, so thanks! This could bring significant improvements in some problems.

src/multivariate/solvers/constrained/fminbox.jl
pkofod: Are you sure `eachindex(gbarrier, x, l, u)` and `eachindex(gbarrier)` compile down to the same thing? Are you assuming that they're all of the same type, or? I'm not saying it's super important, because yes they'll probably always be the same type of arrays with the same number of elements, just saying that you might as well write it like that once you go the `eachindex` route. But it's a minor thing.

pkofod:
Thanks for the quick changes. I'll go over it one more time after tests pass, and then this should be pretty good to go. If some behavior changes or bugs are fixed, I'm probably going to ask for tests.

src/multivariate/solvers/constrained/fminbox.jl
alyst: > Are you sure eachindex(gbarrier, x, l, u) and eachindex(gbarrier) compile down to the same thing?

Sure, `l` and `u` could be different to `x` w.r.t supporting fast linear indexing (which the `master` branch implies), so `eachindex(gbarrier)` could be even incorrect.
AFAIU, `eachindex(a, b, c, ...)` would check the types and lengths of all inputs. On this line, it doesn't matter, but `eachindex(x, l, u)` in `limits_box()` and `barrier_box()` may create some overhead, so for consistency I decided to keep 1-argument version everywhere. I think it's better to have an explicit check for array sizes (and indexing traits) once during the initialization.

pkofod:
Thanks a lot @alyst !

PR #608, opened 2018-05-29
Move IPNewton in from ConstrainedOptim.

pkofod:
@anriseth do you know what's up with Fminbox and GMRES here? Check the log after the tests have run.

@timholy your work finally produces fruit. Thanks for the effort!

I'm not sure how to do it, but especially @timholy, but certainly also @anriseth deserve credit here. I'll see if I can add authors to the commit.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -5,11 +5,11 @@ module Optim
     using Compat
     using LineSearches
     using NLSolversBase
-    using Calculus
+    #using Calculus
     using DiffEqDiffTools
 #    using ReverseDiff
     using ForwardDiff
-
+    import NaNMath
     import Parameters: @with_kw, @unpack
 
     import Compat.String
@@ -23,12 +23,17 @@ module Optim
 
     import NLSolversBase: NonDifferentiable,
                           OnceDifferentiable,
-                          TwiceDifferentiable
+                          TwiceDifferentiable,
+                          nconstraints,
+                          nconstraints_x
 
     export optimize,
            NonDifferentiable,
            OnceDifferentiable,
            TwiceDifferentiable,
+
+           TwiceDifferentiableConstraints,
+
            OptimizationState,
            OptimizationTrace,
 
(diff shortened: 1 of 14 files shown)


anriseth:
> @anriseth do you know what's up with Fminbox and GMRES here? Check the log after the tests have run.

The tests seem to be fine, what am I looking for? If you have an issues locally, it may be that you have to update LineSearches, as the Fminbox+NGMRES tests require https://github.com/JuliaNLSolvers/LineSearches.jl/pull/101 to pass

pkofod:
> The tests seem to be fine, what am I looking for? If you have an issue locally, it may be that you have to update LineSearches, as the Fminbox+NGMRES tests require JuliaNLSolvers/LineSearches.jl#101 to pass

Oh, I think you're right. Wow, is this really a PR for IPNewton where tests pass??? :champagne:

pkofod:
Seems to be multi-authored now

src/Optim.jl
```diff
@@ -5,11 +5,11 @@ module Optim
     using Compat
     using LineSearches
     using NLSolversBase
-    using Calculus
+    #using Calculus
```
anriseth: Is this because we've moved autodiff et al. to NLsolversBase?
If so, let's remove all the differentiation imports and requires?

pkofod: This was actually because it NLSolversBase and Calculus both export `hessian`, and I tried commenting out `Calculus`, and forgot about it because it didn't change anything (I had expected to come back and add it back in). I'll just leave it like this, and remove them all in a separate PR.

anriseth:
> edit: your name is =?UTF-8?q?Asbj=C3=B8rn=20Nilsen=20Riseth?= btw @anriseth

Ah, yes, my name has never been good friends the world's computer systems.

pkofod:
> Ah, yes, my name has never been good friends the world's computer systems.

Seems that the PR travis run failed, hmm

pkofod:
I wonder if those errors come from this PR or from the fminbox one?

PR #613, opened 2018-06-06
Update for new line search functionality.

pkofod:
@anriseth this should sync Optim with LineSearches.

### DIFF
--- a/src/multivariate/solvers/first_order/ngmres.jl
+++ b/src/multivariate/solvers/first_order/ngmres.jl
@@ -355,7 +355,11 @@ function update_state!(d, state::NGMRESState{X,T}, method::AbstractNGMRES) where
         # This may be used in perform_linesearch!/alphaguess! when moving from x^P to x^A
         # TODO: make this a function?
         state.f_x_previous = state.nlpreconstate.f_x_previous
-        state.dphi_0_previous = state.nlpreconstate.dphi_0_previous # assumes precon is a linesearch based method. TODO: Deal with trust region based methods
+        if typeof(method.alphaguess!) <: LineSearches.InitialConstantChange
+            if typeof(state.nlpreconstate.alphaguess!) <: LineSearches.InitialConstantChange
+                method.alphaguess!.dphi_0_previous = state.nlpreconstate.alphaguess!.dphi_0_previous # assumes precon is a linesearch based method. TODO: Deal with trust region based methods
+            end
+        end
         # state.x_previous and state.x are dealt with by reference
 
         lssuccess = perform_linesearch!(state, method, ManifoldObjective(method.manifold, d))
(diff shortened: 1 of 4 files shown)


src/utilities/perform_linesearch.jl
```diff
@@ -40,10 +40,8 @@ function perform_linesearch!(state, method, d)
         state.alpha, ϕalpha =
             method.linesearch!(d, state.x, state.s, state.alpha,
                                state.x_ls, phi_0, dphi_0)
-        state.dphi_0_previous = dphi_0
```
anriseth: Should you not update `dphi_0_previous` if `is` is of type `InitialConstantChange`?

src/utilities/perform_linesearch.jl
```diff
@@ -40,10 +40,8 @@ function perform_linesearch!(state, method, d)
...
         return true # lssuccess = true
     catch ex
-        state.dphi_0_previous = dphi_0
```
anriseth: Same as above

src/utilities/perform_linesearch.jl
pkofod: https://github.com/JuliaNLSolvers/LineSearches.jl/blob/c3d076343e2928408a4798961c802632ff692ed0/src/initialguess.jl#L113 ?

src/multivariate/solvers/first_order/ngmres.jl
```diff
@@ -355,7 +355,11 @@ function update_state!(d, state::NGMRESState{X,T}, method::AbstractNGMRES) where
         # This may be used in perform_linesearch!/alphaguess! when moving from x^P to x^A
         # TODO: make this a function?
         state.f_x_previous = state.nlpreconstate.f_x_previous
-        state.dphi_0_previous = state.nlpreconstate.dphi_0_previous # assumes precon is a linesearch based method. TODO: Deal with trust region based methods
+        if typeof(method.alphaguess!) <: LineSearches.InitialConstantChange
+            if typeof(state.nlpreconstate.alphaguess!) <: LineSearches.InitialConstantChange
+                method.alphaguess!.dphi_0_previous = state.nlpreconstate.alphaguess!.dphi_0_previous # assumes precon is a linesearch based method. TODO: Deal with trust region based methods
```
anriseth: Isn't this property now called `dϕ_0_previous`?

src/utilities/perform_linesearch.jl
anriseth: Yes, I don't see the value of `dϕ_0_previous` being assigned or updated anywhere in either Optim or LineSearches?

pkofod: That's an oversight. That's supposed to happen in LineSearches!

src/multivariate/solvers/first_order/ngmres.jl
anriseth: @pkofod this will crash because the property in alphaguess! Is no longer called `dphi_0_previous`; it uses the symbolic version of phi.

pkofod: Right. Since GMRES uses a specific type it should probably have tests with combinations of these line searches, right?

PR #616, opened 2018-06-07
Error if Newton is supplied as optimizer and allow for inplace and au…

### LINKED ISSUE #589: autodiff keyword for Fminbox

pkofod:
…todiff in Fminbox optimize calls.

fixes #589 and #606

### DIFF
--- a/src/multivariate/solvers/constrained/fminbox.jl
+++ b/src/multivariate/solvers/constrained/fminbox.jl
@@ -147,6 +147,7 @@ function optimize(f,
 
     g! = inplace ? g : (G, x) -> copy!(G, g(x))
     od = OnceDifferentiable(f, g!, initial_x, zero(T))
+
     optimize(od, l, u, initial_x, F)
 end
 
@@ -156,7 +157,7 @@ function optimize(f,
                   initial_x::AbstractArray{T},
                   F::Fminbox = Fminbox(); inplace = true, autodiff = :finite) where T<:AbstractFloat
 
-    od = OnceDifferentiable(f, initial_x, zero(T); autodiff = :finite)
+    od = OnceDifferentiable(f, initial_x, zero(T); autodiff = autodiff)
     optimize(od, l, u, initial_x, F)
 end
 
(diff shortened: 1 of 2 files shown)


src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -99,8 +99,7 @@ function Fminbox(method::AbstractOptimizer = LBFGS();
                  mu0::Real = NaN, mufactor::Real = 0.001,
                  precondprep = (P, x, l, u, mu) -> precondprepbox!(P, x, l, u, mu))
     if method isa Newton
-        warn("Newton is not supported as the Fminbox optimizer. Defaulting to LBFGS.")
-        method = LBFGS()
+        error("Newton is not supported as the Fminbox optimizer.")
```
anriseth: ArgumentError? :p

pkofod: How fancy... sure :)

pkofod:
I know the g is wrong

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -140,15 +139,27 @@ function optimize(obj,
...
     optimize(od, l, u, initial_x, F)
 end
 
+function optimize(f,
+                  l::AbstractArray{T},
+                  u::AbstractArray{T},
+                  initial_x::AbstractArray{T},
+                  F::Fminbox = Fminbox(); inplace = true, autodiff = :finite) where T<:AbstractFloat
+
+    od = OnceDifferentiable(f, initial_x, zero(T); autodiff = :finite)
```
anriseth: This should say `autodiff = autodiff` I assume?
And does `OnceDifferentiable` take `inplace` arguments?

pkofod: It should, but it doesn't really. The keyword has been introduced a few places, but it's not used. Oops. We can tag a feature tag at NLSolversBase, and tag Optim after that (to avoid changing REQUIRE and having to tag a major version again just after another).

PR #617, opened 2018-06-12
use inplace keyword in constructors from NLSolversBase.

pkofod:


pkofod:
I think an Optim major tag is good to go after this. @anriseth ?

anriseth:
Great :) And then we can leave that as the latest major for 0.6 and focus on 0.7 from now on?

PR #618, opened 2018-06-14
Add usage example to complex optimization

anriseth:
We should probably find a clever way to add Literate examples to the Examples section of existing docs. Alternatively, we could just list links to them in the `complex.md` file.

Also, @antoine-levitt, do you have any references for me that can be useful to add to the docs?


pkofod:
> We should probably find a clever way to add Literate examples to the Examples section

This is something for you to consider @iewaij

iewaij:
Yeah I can work on it!

antoine-levitt:
Really the definition should be enough. There's https://arxiv.org/pdf/0906.4835.pdf which is very verbose but covers the material.

anriseth:
Good to go :+1:

PR #622, opened 2018-06-18
Remove unnecessary update_h in Newton

anriseth:


pkofod:
I just noticed that yesterday as well...

PR #628, opened 2018-06-20
Fix #626

### LINKED ISSUE #626: Problem with some line searches and manifolds
This is the simplest reproducer I could find:

```
srand(0)
const n = 4
const m = 1
M = randn(n,n)
# M = M'M
const A = (M+M')/2
f(x) = -vecdot(x,A*x)
g(x) = -2*A*x
g!(stor,x) = copy!(stor,g(x))
x0 = randn(n)

manif = Optim.Sphere()
Optim.optimize(f, g!, x0, Optim.GradientDescent(manifold=manif, linesearch=Optim.Static(alpha=.01)), Optim.Options(show_trace=true, allow_f_increases=true,g_tol=1e-6))
```

Notice it does converge, but the reported gradient is wrong. BackTracking is also affected, MoreThuente and HagerZhang are OK. Any idea what's going on?

antoine-levitt:
And add tests.


pkofod:
Tests still seem to fail here.

antoine-levitt:
Yes but unrelated to this PR I think

anriseth:
Seems ready to merge?

PR #632, opened 2018-06-21
Add IPNewton docs and constrained optimization Literate example

anriseth:
I'm moving the IPNewton "tutorial" from the ConstrainedOptim README to the docs, and taking the liberty of introducing Literate in our documentation toolbox :) 

@djlacombe you can use this PR as a template if you wish to add tutorials to our docs :) [I can add your maximum likelihood example to this PR for now if you like, and make sure the commit is attributed to you]

PR #634, opened 2018-06-22
Add Literate generator script

anriseth:
I forgot to add a file to #632, so the docs generation failed.

This would have been caught if we run the docs generation script from `runtests.jl` in the same way it's done in LineSearches and JuAFEM. @pkofod if you prefer to keep `.travis.yml` as before I can revert the second commit in this PR, otherwise, this is good-to-go.


pkofod:
So will this mean that every time a PR is made, docs will be updated, instead of after it's merged?

anriseth:
> So will this mean that every time a PR is made, docs will be updated, instead of after it's merged?

I think it only happens when something is committed to `master` and when a new release is tagged.

PR #635, opened 2018-06-22
Clean up and add comments.

pkofod:


anriseth:
Travis 0.6 breaks because I forgot to add `Documenter` to `test/REQUIRE`. It's fixed with #636

PR #639, opened 2018-06-25
Fix example docs and add admonition to mkdocs

anriseth:
I've fixed the following:
- mkdocs got confused with indented code in my IPNewton example
- Correct link to the maximum likelihood example
- Add `admonition` functionality to MkDocs (to enable `tip` blocks etc.)

@pkofod I have noticed two issues with the `windmill` theme:
- The search page does not work https://github.com/gristlabs/mkdocs-windmill/pull/16
- They don't have any CSS code for the admonition blocks, so they don't stand out as much as in other themese / the HTML format from Documenter.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,9 +1,9 @@
 Optim.jl is licensed under the MIT License:
 
 > Copyright (c) 2012: John Myles White, Tim Holy, and other contributors.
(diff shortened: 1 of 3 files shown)


anriseth:
I used CSS once 13 years age :laughing:

pkofod:
> I used CSS once 13 years ago 😆

Well, if you're willing to work for 5 dollars a months, I'll hire you!

anriseth:
Ready to go :+1:

PR #641, opened 2018-06-25
Allow f increases

pkofod:
As argued on gitter, this is really LineSearches related if anything.

Agree? If not, let's discuss it.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -41,7 +41,7 @@ function Options(;
         f_calls_limit::Int = 0,
         g_calls_limit::Int = 0,
         h_calls_limit::Int = 0,
-        allow_f_increases::Bool = false,
+        allow_f_increases::Bool = true,
         allow_outer_f_increases::Bool = false,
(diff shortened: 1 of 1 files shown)


anriseth:
Agree with the linesearch based methods (Does this affect the zeroth order methods and the Trust Region methods?)

If the objective is C¹ then Optim guarantees to step in a descent direction https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/utilities/perform_linesearch.jl#L26.
Therefore, the only reason for ending up with an objective increase is wrong/non-continuous gradient, floating point issues, or line search failure/Static step.

anriseth:
If we are worried about the non-linesearch methods, we can set their default behaviour in the same fashion as https://github.com/JuliaNLSolvers/Optim.jl/blob/c800549a6be838fd23388c58a3461dce47342d2a/src/multivariate/solvers/first_order/ngmres.jl#L455 (but with the option set to `false`).

PR #642, opened 2018-06-28
pass options arg through to fminbox

noahstier:
I noticed my f_calls_limit option was not respected. Before this change, the below example calls f 1157 times, after the change it calls f 287 times.

```
import Optim

x0 = [2.0, 2.0]
lowerbounds = [1.0, 1.0]
upperbounds = [3.0, 3.0]

result = Optim.optimize(
  f,
  lowerbounds,
  upperbounds,
  x0,
  Optim.Fminbox(Optim.NelderMead()),
  Optim.Options(f_calls_limit=100)
)
println(result.f_calls)
```


pkofod:
Thanks! Will merge when at a computer

PR #643, opened 2018-06-28
fix a lot of v0.7 stuff!

pkofod:


pkofod:
@anriseth OptimTestProblems has a ton of problems :] I wonder how much of that femtocleaner can fix

anriseth:
Did you use femtocleaner here, or `sed`/search-replace?

I can try femtocleaner on OptimTestProblems in the next few days, and try to fix missing stuff manually

anriseth:
Are they eliminating `vecnorm`?

pkofod:
> Did you use femtocleaner here, or sed/search-replace?

the latter

pkofod:
> Are they eliminating vecnorm?

yes. norm is vecnorm now, and matrix norms are now the matrix norm is opnorm instead.

dawbarton:
What is the status of this PR? Do you want any help with it?

anriseth:
Things have slowed down due to the summer holidays for Patrick and me.

I think it's best to tag 0.7 versions of the NLSolvers dependencies first, to run the CI tests properly.

I'll do NLSolversBase 7.0 this weekend and then we can do LineSearches and possibly Optim next week

anriseth:
@dawbarton I've had a look at this now and it seems that the package `PositiveFactorizations` is holding us up. There have been attempts to update the package (https://github.com/timholy/PositiveFactorizations.jl/pull/15), but I don't think the maintainer @timholy has time to fix this at the moment. I wonder if there are others who have admin rights in the package?

antoine-levitt:
Hm, looking at the commit history it doesn't appear so. Not sure what to do here except fork it, unfortunately...

anriseth:
According to Slack, Tim is away without internet for two weeks. So for those who wants to use Optim on 0.7 in the meantime can do the following:
- create a PR on PositiveFactorizations that fixes the `eigfact` and dep issues there
- check out the PositiveFactorization branch along with one of the 0.7 branches we have for Optim.

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.6.0
+julia 0.7-beta
```
anriseth: `beta2` now?

REQUIRE
```diff
@@ -1,4 +1,4 @@
...
 PositiveFactorizations
 Compat 0.18.0
```
anriseth: I think we should remove all `Compat` stuff we stop supporting Julia 0.6

pkofod:
One solution here is for me to actually make the `linsolve` PR asap, and then we can implement algorithm 6.3 in N&W (easy) or 6.5 in N&W (slightly trickier, but it's fully specified so not too tricky) and use that as the default `linsolve`. We can then refer to PositiveFactorizations.jl as an alternativ way of solving for the step size in the docs/a tutorial. It'll be one less "outside" dependency. cc @timholy

mkborregaard:
Anything that needs to be done before merging this and becoming 1.0 compliant? It seems to break StatPlots via KDE

pkofod:
> Anything that needs to be done before merging this and becoming 1.0 compliant? It seems to break StatPlots via KDE

It's your lucky day my friend :)

mkborregaard:
God bless! :pray:

PR #656, opened 2018-08-12
Update .travis.yml

pkofod:


.travis.yml
```diff
@@ -13,8 +14,8 @@ notifications:
 sudo: false
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'Pkg.clone(pwd()); Pkg.test("Optim", coverage=true)'
+    - julia -e 'using Pkg;Pkg.clone(pwd()); Pkg.test("Optim", coverage=true)'
 after_success:
-    - julia -e 'cd(Pkg.dir("Optim")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+    - julia -e 'using Pkg;cd(Pkg.dir("Optim")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
```
fredrikekre: No need to `cd` here; we are already in the package folder.

pkofod: Cool, I was just adding `using Pkg` to get that sweet v1.0 CI ;)

Is there a new reference script, btw? This is basically just what `Pkg.generate` generated ages ago :)

fredrikekre: The default script should work, so you can probably remove the `script:`. The `after_success:` section with coverage looks good, but you can remove the `cd`, especially since `Pkg.dir` does not exist in 1.0.

PR #666, opened 2018-08-29
Document algorithm choice and derivatives

antoine-levitt:
Fix https://github.com/JuliaNLSolvers/Optim.jl/issues/665. Edits welcome. In particular, someone more knowledgeable than me should say something about alphaguess. Also I was not able to build the docs locally, so here's hoping CI does its job.


pkofod:
Great! Havn't read it, but I'll try to find time tomorrow evening.

docs/src/user/algochoice.md
```diff
@@ -0,0 +1,17 @@
...
+
+If you can provide analytic gradients and Hessians, and the dimension of the problem is not too large, then second order methods are very efficient. The Newton method with trust region is the method of choice. 
+
+When you do not have an explicit Hessian or when the dimension becomes large enough that the linear solve in the Newton method becomes the bottleneck, first order methods should be preferred. BFGS is often the most efficient, but also requires a linear system solve. LBFGS usually has a performance very close to that of BFGS, and avoids linear system solves (the parameter `m` can be tweaked: increasing it should result in better convergence behavior, at the expense of memory and time spent in linear algebra operations). The conjugate gradient method usually converges less quickly than LBFGS, but requires less memory. Gradient descent should only be used for testing. Acceleration methods are experimental.
```
anriseth: Are you sure BFGS is more efficient than LBFGS? 

1. It is not always true that increasing `m` reduces the number of iterations/function evaluations, you can see an initial improvement but then for too large `m` it is detrimental to convergence. In these cases LBFGS may be better than BFGS. 
2. With LBFGS we can update the scaling of the Hessian (https://github.com/JuliaNLSolvers/Optim.jl/pull/484), which more often than not result in better initial step lengths for the linesearches.

docs/src/user/algochoice.md
```diff
@@ -0,0 +1,17 @@
...
+
+When the objective function is non-differentiable or you do not want to use gradients, use zeroth-order methods. Nelder-Mead is currently the most robust.
+
+## Linesearches
+
+Linesearches are used in every first- and second- method except for the trust-region Newton method. Linesearch routines attempt to locate quickly an approximate minimizer of the univariate function ``\alpha \to f(x+ \alpha d)``, where ``d`` is the descent direction computed by the algorithm. They vary in how accurate this minimization is. Two good linesearches are BackTracking and HagerZhang, the former being less stringent than the latter. For well-conditioned objective functions and methods where the step is usually well-scaled (such as LBFGS or Newton), a rough linesearch such as BackTracking is usually the most performant. For badly behaved problems or when extreme accuracy is needed (gradients below the square root of the machine epsilon, about ``10^{-8}`` with `Float64`), the HagerZhang method proves more robust. An exception is the conjugate gradient method which requires an accurate linesearch to be efficient, and should be used with the HagerZhang linesearch.
```
anriseth: "second-**order** method"

docs/src/user/algochoice.md
antoine-levitt: I've been more careful with the language here. I also added a summary for users in a hurry.

anriseth:
Looks good to me :+1:

antoine-levitt:
Can you or @pkofod chime in on `alphaguess` before merging?

anriseth:
Here are some general comments re: alphaguess. 

1. The point of alphaguess is to make an estimate for a sensible step length. It often does so by trying to understand the "scaling" of the problem. (There is a discussion about initial step lengths in NW 2006, section 3.5 "Step-length selection algorithms"))
2. (Quasi-)Newton methods, especially L-BFGS and Newton, often adapt well-scaled step sizes and therefore `InitialStatic(alpha=1)` is a good initial guess.
3. The GradientDescent and CG algorithms can generate poorly scaled step directions, and the alphaguess may help the linesearch (as noted in NW). I'd probably try InitialHZ + HZ first for both, but the other alphaguess methods may be useful her too.
2. Never use `InitialPrevious` with `BackTracking` (your`alpha`s will then generate a decreasing sequence and step sizes end up being tiny)

My (limited) experience is that `InitialStatic` works best, with the potential exception of `InitialHagerZhang` in combination with `ConjugateGradient` and `HagerZhang`. 
I believe @cortner was interested in experimenting with different alphaguesses at some point (before we had implemented them), maybe he has something to add too.

cortner:
> My (limited) experience is that InitialStatic works best, ...

I suspect this is *highly* problem-dependent. This kind of alphaguess never worked well for me. My feeling is that for nCG and even for LBFGS for large-scale problems (indenpently of which LS you choose) a good alpha guess is crucial, otherwise most of the time is wasted on line-search.

cortner:
No! For a sufficiently nonlinear problem this would be terrible. I think the advise is to experiment, but not say how. 

The best option would be to implement a tuning code that tests many different settings automatically and makes a recommendation.

pkofod:
> The best option would be to implement a tuning code that tests many different settings automatically and makes a recommendation.

I like the idea, but it's not clear to me how to communicate properties of the problem - are there even such properties / classes identified in the literature beyond things like linear, quadratic, separable, and so  on ?

pkofod:
I think that we should maybe just merge as is (thanks a lot @antoine-levitt ! it's really a valuable addition to the docs), and wait with alphaguess until we might actually be able to test some of our conjectures on a collection of problems. If you agree, I will merge.

antoine-levitt:
Fine by me. I don't have any firsthand knowledge of tuning alphaguess, will try it next time.

PR #668, opened 2018-09-04
Use _ind2sub

### LINKED ISSUE #667: Box Constraints
julia 1.0.  "Optim"              => v"0.17.0"

```
julia> using Optim

julia> myfun(x)= (x[1]-2)^2 + (x[2]-3)^2 + ((x[1]-2)*(x[2]-3))^2;

julia> inner_optimizer= GradientDescent()

julia>  results = optimize( myfun, [-Inf,-Inf], [Inf,Inf], initial_x, Fminbox(inner_optimizer))
 * Algorithm: Fminbox with Gradient Descent
 * Starting Point: [2.0,2.0]
 * Minimizer: [2.0,2.9999999999973572]
 * Minimum: 6.984259e-24
 * Iterations: 1
 * Convergence: true
   * |x - x'| ≤ 0.0e+00: false
     |x - x'| = 1.00e+00
   * |f(x) - f(x')| ≤ 0.0e+00 |f(x)|: false
     |f(x) - f(x')| = 1.43e+23 |f(x)|
   * |g(x)| ≤ 1.0e-08: true
     |g(x)| = 5.29e-12
 * Objective Calls: 3
 * Gradient Calls: 3

julia>  results = optimize( myfun, [-Inf,-Inf], [Inf,1.0], initial_x, Fminbox(inner_optimizer))
ERROR: UndefVarError: ind2sub not defined
Stacktrace:
 [1] optimize(::OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Fminbox{GradientDescent{LineSearches.InitialPrevious{Float64},LineSearches.HagerZhang{Float64,Base.RefValue{Bool}},Nothing,getfield(Optim, Symbol("##12#14"))},Float64,getfield(Optim, Symbol("##46#48"))}, ::Optim.Options{Float64,Nothing}) at /Users/ivo/.julia/packages/Optim/fabGe/src/multivariate/solvers/constrained/fminbox.jl:206
 [2] #optimize#53(::Bool, ::Symbol, ::Function, ::Function, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Fminbox{GradientDescent{LineSearches.InitialPrevious{Float64},LineSearches.HagerZhang{Float64,Base.RefValue{Bool}},Nothing,getfield(Optim, Symbol("##12#14"))},Float64,getfield(Optim, Symbol("##46#48"))}, ::Optim.Options{Float64,Nothing}) at /Users/ivo/.julia/packages/Optim/fabGe/src/multivariate/solvers/constrained/fminbox.jl:164
 [3] optimize at /Users/ivo/.julia/packages/Optim/fabGe/src/multivariate/solvers/constrained/fminbox.jl:163 [inlined] (repeats 2 times)
 [4] top-level scope at none:0
```

[a] what am I doing wrong?

[b] could the error message be more informative?

anriseth:
Fix #667 

Now that `ind2sub` has been removed, I decided to just call `Base._ind2sub`, but one could also do `CartesianIndices(x)[i].I`. That seemed a bit wasteful, but maybe there is a better way to use the new CartesianIndex systems?

### DIFF
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -112,7 +112,7 @@ struct PowerManifold<:Manifold
     outer_dims::Tuple
 end
 function retract!(m::PowerManifold, x)
-    for i=1:prod(m.outer_dims)
+    for i=1:prod(m.outer_dims) # TODO: use for i in LinearIndices(m.outer_dims)?
         retract!(m.inner_manifold,get_inner(m, x, i))
     end
     x
@@ -128,10 +128,6 @@ end
     size_outer = prod(m.outer_dims)
     @assert 1 <= i <= size_outer
     return reshape(view(x, (i-1)*size_inner+1:i*size_inner), m.inner_dims)
-end
-@inline function get_inner(m::PowerManifold, x, i::Tuple)
-    s2i = LinearIndices(m.outer_dims)
-    get_inner(m, x, s2i[i...])
 end
 
 
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -129,13 +129,16 @@ end
     @assert 1 <= i <= size_outer
     return reshape(view(x, (i-1)*size_inner+1:i*size_inner), m.inner_dims)
 end
-@inline get_inner(m::PowerManifold, x, i::Tuple) = get_inner(m, x, ind2sub(m.outer_dims, i...))
+@inline function get_inner(m::PowerManifold, x, i::Tuple)
+    s2i = LinearIndices(m.outer_dims)
+    get_inner(m, x, s2i[i...])
+end
 
 
 """
 Product of two manifolds {P = (x1,x2), x1 ∈ m1, x2 ∈ m2}.
 P is stored as a flat 1D array, and x1 is before x2 in memory.
-Use get_inner(m, x, {1,2}) to access x1 and x2 in their original format.
+Use get_inner(m, x, {1,2}) to access x1 or x2 in their original format.
 """
 struct ProductManifold<:Manifold
     m1::Manifold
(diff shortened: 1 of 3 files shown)


timholy:
Not recommended. `ind2sub` is a worse interface: you have to remember which order arguments come in. (Does the shape of the array come first or the index come first? If you haven't called it recently you always have to check the documentation.) And what about arrays that don't start indexing at 1, how will that work? (Do you pass the `axes` rather than the `size`?)

In contrast:
```julia
i2s = CartesianIndices(x)   # this is only a property of x so there is nothing to remember
i2s[i]    # this only uses i so there is nothing to remember
```
and of course you can user `i2s` in as many places as you want; consequently if you have several conversions it's actually shorter than `ind2sub`.

If you need a tuple use `Tuple(i2s[i])` rather than `i2s[i].I`.

anriseth:
Thank you for the clarification, @timholy :) 

> If you need a tuple use Tuple(i2s[i]) rather than i2s[i].I.

I don't understand the difference between using one or the other here, I thought that `CartesianIndex.I` was always a Tuple?

@antoine-levitt you have used `ind2sub` in `get_inner` for `PowerManifold`s. This will throw an error, but I don't understand exactly what the particular method does; it seems to be calling itself? (`ind2sub` returns a `Tuple`). I have deleted the method; let me know if I have misunderstood (and possibly PR a fix with a test :) )

https://github.com/JuliaNLSolvers/Optim.jl/blob/2b6c6ed1e527d7ed8dfe509db29f7243ac283883/src/Manifolds.jl#L132

timholy:
> I don't understand the difference between using one or the other here, I thought that CartesianIndex.I was always a Tuple?

Right. This is just a stylistic thing, generally we discourage direct field access in favor of calling a function. Not a big deal either way.

anriseth:
> Should it just be getindex instead of get_inner?

That sounds sensible. Would you then want `AbstractProductManifold` types to use `getindex` to return the a `Manifold` representing the inner manifold? [I'll leave this to you in another PR :p ]

pkofod:
This seems done?

pkofod:
Then why doesn’t AV pass? ;)

PR #670, opened 2018-09-09
Use NaNMath to prevent NaNs in CG

### LINKED ISSUE #669: CG errors when current and previous gradients are the same
I tracked down a bug that happens due to these 2 lines:

https://github.com/JuliaNLSolvers/Optim.jl/blob/e12fdbf1a64b8c28df42289d8a962d55b2327e18/src/multivariate/solvers/first_order/cg.jl#L178

https://github.com/JuliaNLSolvers/Optim.jl/blob/e12fdbf1a64b8c28df42289d8a962d55b2327e18/src/multivariate/solvers/first_order/cg.jl#L183

So when `gradient(d)` and `state.g_previous` are the same, `betak` will be `NaN` and that will result in a `NaN`s search direction which will corrupt the next line search. Interestingly, the backtracking line search still "succeeds" but gives a `NaN` step which gives a `NaN`s `x` which gives a `NaN`s objective which is caught by this line:

https://github.com/JuliaNLSolvers/Optim.jl/blob/e12fdbf1a64b8c28df42289d8a962d55b2327e18/src/multivariate/solvers/first_order/cg.jl#L164

anriseth:
Fix #669 

There's a question as to whether we should restart CG when we hit a NaN-situation instead of the current approach that takes the truncation value `eta_k` instead. I can't find any discussion of this in the CG papers as they conveniently consider nice functions or Wolfe-condition steps.

### DIFF
--- a/src/multivariate/solvers/first_order/cg.jl
+++ b/src/multivariate/solvers/first_order/cg.jl
@@ -10,10 +10,11 @@
 # reference to a particular point in this paper.
 #
 # Several aspects of the following have also been incorporated:
-#   W. W. Hager and H. Zhang (2012) The limited memory conjugate
+#   W. W. Hager and H. Zhang (2013) The limited memory conjugate
 #     gradient method.
 #
-# This paper will be denoted HZ2012 below.
+# This paper will be denoted HZ2013 below.
+#
 #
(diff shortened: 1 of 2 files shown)


anriseth:
This is ready to go; the error in AppVayor is fixed in #677

PR #674, opened 2018-09-12
Temporary(?) doc generation fix for 1.0

anriseth:
I assume that `TRAVIS_JULIA_VERSION` is `1.0`.

Maybe we should go towards Julia 1.0 Projects to handle docs. That approach was recently implemented in Literate.jl by @fredrikekre

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -237,6 +237,6 @@ println("Literate examples")
 
 
 # Build the docs
-if get(ENV, "TRAVIS_OS_NAME", "") == "linux" && get(ENV, "TRAVIS_JULIA_VERSION", "") == "0.6"
+if get(ENV, "TRAVIS_OS_NAME", "") == "linux" && get(ENV, "TRAVIS_JULIA_VERSION", "") == "1.0"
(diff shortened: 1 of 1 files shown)


fredrikekre:
I can recommend that, that method will also be the recommended approach for the next Documenter release.

fredrikekre:
You also need to bump https://github.com/JuliaNLSolvers/Optim.jl/blob/master/docs/make.jl#L18 btw. I can submit a PR for a doc-project if you want

anriseth:
> You also need to bump https://github.com/JuliaNLSolvers/Optim.jl/blob/master/docs/make.jl#L18 btw. 

Thanks :)

> I can submit a PR for a doc-project if you want

That would be great, cheers! Before you do, let's just check with @pkofod if he has an opinion on what direction to take doc making for JuliaNLSolvers.

pkofod:
As long as it builds, I honestly don't care. I always seem to make these things break, so I'm more than happy to take advise!

pkofod:
Looks good now, right?

PR #676, opened 2018-09-13
Use docs/Project.toml for building docs.

fredrikekre:


pkofod:
Thanks! Glad to see this being handled outside of `runtests`.

fredrikekre:
Yea, using Travis build stages like this is pretty nice, it will still fail the build, which was the reason to have doc building in `runtests.jl` in the first place.

Evizero:
this is cool

pkofod:
> Yea, using Travis build stages like this is pretty nice, it will still fail the build, which was the reason to have doc building in `runtests.jl` in the first place.

Very happy to have an expert drop by! :)

fredrikekre:
64-bit failures on AV seems unrelated.

PR #683, opened 2018-10-01
At least say (only) Julia 1.0 supported

PallHaraldsson:


PallHaraldsson:
Is it helpful to doc this (and versions in between)? I didn't look at what else changed, it must be more since only 1.0 now supported.

Should it have required only 0.7, to allow people to stay at that version (for a while more)?

pkofod:
I’m not sure what the issue is exactly. Is there no version that will install?

PR #689, opened 2018-11-09
Fix Latex in Rasch example

bdeonovic:


bdeonovic:
Not sure if whatever math markdown is being used has `dcases` might have to switch to regular `cases` environment

pkofod:
Let's see

pkofod:
I made a few changes and the cases now show correctly, but I think we just need a display style and use a normal frac

PR #692, opened 2018-12-07
docs to compute objective, grad and hess simultaneously

cossio:
Update the docs on how to reuse computations in common between objective function, gradient and Hessian as required.


pkofod:
I plan to change some things that will probably change the code parts here, but this is good for now. Thanks

PR #705, opened 2019-03-14
Update newton_trust_region.jl

nan2ge1:
I tried to improve the readability and reduce some overhead for `newton_trust_region`. The changes are submitted for review and discussion. Line comments will be added to less minor changes.
These changes have been tested by minimizing the Rosenbrock function with Julia `v1.1.0` and Optim `v0.17.2+`.
```julia
using Optim
f(x)   = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
result = optimize(f, zeros(2), NewtonTrustRegion())
```
Before change:
```julia
julia> include("script_optim.jl")
 * Algorithm: Newton's Method (Trust Region)
 * Starting Point: [0.0,0.0]
 * Minimizer: [0.9999999926632145,0.9999999853263007]
 * Minimum: 5.382842e-17
 * Iterations: 15
 * Convergence: true
   * |x - x'| = 0.0e+00: false
     |x - x'| = 7.52e-07
   * |f(x) - f(x')| = 0.0e+00 |f(x)|: false
     |f(x) - f(x')| = 4.90e+03 |f(x)|
   * |g(x)| = 1.0e-08: true
     |g(x)| = 4.51e-11
 * Objective Calls: 16
 * Gradient Calls: 16
 * Hessian Calls: 14
```
After change:
```julia
julia> include("script_optim.jl")
 * Algorithm: Newton's Method (Trust Region)
 * Starting Point: [0.0,0.0]
 * Minimizer: [0.9999999926632145,0.9999999853263007]
 * Minimum: 5.382842e-17
 * Iterations: 15
 * Convergence: true
   * |x - x'| = 0.0e+00: false
     |x - x'| = 7.52e-07
   * |f(x) - f(x')| = 0.0e+00 |f(x)|: false
     |f(x) - f(x')| = 4.90e+03 |f(x)|
   * |g(x)| = 1.0e-08: true
     |g(x)| = 4.51e-11
 * Objective Calls: 16
 * Gradient Calls: 16
 * Hessian Calls: 14
```

### DIFF
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -136,7 +136,7 @@ function solve_tr_subproblem!(gr,
                 # I don't think it matters which eigenvector we pick so take
                 # the first.
                 calc_p!(lambda, min_i, n, qg, H_eig, s)
-                s[:] = -s + tau * H_eig.vectors[i, 1]
+                s[:] = -s + tau * H_eig.vectors[:, 1]
             end
         end
 
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -136,7 +136,7 @@ function solve_tr_subproblem!(gr,
                 # I don't think it matters which eigenvector we pick so take
                 # the first.
                 calc_p!(lambda, min_i, n, qg, H_eig, s)
-                p[:] = -p + tau * H_eig.vectors[i, 1]
+                s[:] = -s + tau * H_eig.vectors[i, 1]
             end
         end
 
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -36,7 +36,7 @@ function check_hard_case_candidate(H_eigv, qg)
     hard_case, lambda_index
 end
 
-# Equation 4.38 in N&W
+# Equation 4.38 in N&W (2006)
 function calc_p!(lambda::T, min_i, n, qg, H_eig, p) where T
     fill!( p, zero(T) )
     for i = min_i:n
@@ -46,7 +46,7 @@ function calc_p!(lambda::T, min_i, n, qg, H_eig, p) where T
 end
 
 # Choose a point in the trust region for the next step using
-# the interative (nearly exact) method of section 4.3 of Nocedal and Wright.
+# the interative (nearly exact) method of section 4.3 of N&W (2006).
 # This is appropriate for Hessians that you factorize quickly.
 #
 # Args:
@@ -61,7 +61,7 @@ end
 #  m - The numeric value of the quadratic minimization.
 #  interior - A boolean indicating whether the solution was interior
 #  lambda - The chosen regularizing quantity
-#  hard_case - Whether or not it was a "hard case" as described by N&W
+#  hard_case - Whether or not it was a "hard case" as described by N&W (2006)
 #  reached_solution - Whether or not a solution was reached (as opposed to
 #      terminating early due to max_iters)
 function solve_tr_subproblem!(gr,
@@ -122,7 +122,7 @@ function solve_tr_subproblem!(gr,
             # to find a multiple of an orthogonal eigenvector that lands the
             # iterate on the boundary.
 
-            # Formula 4.45 in N&W
+            # Formula 4.45 in N&W (2006)
             calc_p!(lambda, min_i, n, qg, H_eig, s)
             p_lambda2 = sum(abs2, s)
             if p_lambda2 > delta_sq
@@ -141,8 +141,8 @@ function solve_tr_subproblem!(gr,
         end
 
         if !hard_case
-            # Algorithim 4.3 of N&W, with s insted of p_l for consistency with
-            # Optim.jl
+            # Algorithim 4.3 of N&W (2006), with s insted of p_l for consistency
+            # with Optim.jl
 
             reached_solution = false
             for iter in 1:max_iters
@@ -211,11 +211,11 @@ second-order information in a function's Hessian, but with more stability that
 Newton's method when functions are not globally well-approximated by a quadratic.
 This is achieved by repeatedly minimizing quadratic approximations within a
 dynamically-sized trust region in which the function is assumed to be locally
-quadratic. See Wright and Nocedal and Wright (ch. 4, 1999) for a discussion of
+quadratic. See Wright and Nocedal and Wright (ch. 4, 2006) for a discussion of
 trust-region methods in practice.
 
 ## References
- - Nocedal, J. and S. J. Wright (1999), Numerical optimization. Springer Science 35.67-68: 7.
+ - Nocedal, J., & Wright, S. (2006). Numerical optimization. Springer Science & Business Media.
 """
 NewtonTrustRegion(; initial_delta::Real = 1.0,
                     delta_hat::Real = 100.0,
@@ -296,7 +296,7 @@ function update_state!(d, state::NewtonTrustRegionState, method::NewtonTrustRegi
     value_gradient!(d, state.x)
 
     # Update the trust region size based on the discrepancy between
-    # the predicted and actual function values.  (Algorithm 4.1 in N&W)
+    # the predicted and actual function values.  (Algorithm 4.1 in N&W (2006))
     f_x_diff = state.f_x_previous - value(d)
     if abs(m) <= eps(T)
         # This should only happen when the step is very small, in which case
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -7,8 +7,8 @@
 #
 # Returns:
 #  hard_case: Whether it is a candidate for the hard case
-#  lambda_1_multiplicity: The number of times the lowest eigenvalue is repeated,
-#                         which is only correct if hard_case is true.
+#  lambda_index: The index of the first lambda not equal to the smallest
+#                eigenvalue, which is only correct if hard_case is true.
 function check_hard_case_candidate(H_eigv, qg)
     @assert length(H_eigv) == length(qg)
     if H_eigv[1] >= 0
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -33,17 +33,16 @@ function check_hard_case_candidate(H_eigv, qg)
         end
     end
 
-    hard_case, lambda_index - 1
+    hard_case, lambda_index
 end
 
-# Function 4.39 in N&W
-function p_sq_norm(lambda::T, min_i, n, qg, H_eig) where T
-    p_sum = zero(T)
-    H_eigvals = H_eig.values
+# Equation 4.38 in N&W
+function calc_p!(lambda::T, min_i, n, qg, H_eig, p) where T
+    fill!( p, zero(T) )
     for i = min_i:n
-        p_sum += qg[i]^2 / (lambda + H_eigvals[i])^2
+        p[:] -= qg[i] / (H_eig.values[i] + lambda) * H_eig.vectors[:, i]
     end
-    p_sum
+    return nothing
 end
 
 # Choose a point in the trust region for the next step using
@@ -86,10 +85,7 @@ function solve_tr_subproblem!(gr,
     H_ridged = copy(H)
 
     # Cache the inner products between the eigenvectors and the gradient.
-    qg = similar(gr)
-    for i=1:n
-        qg[i] = dot(H_eig.vectors[:, i], gr)
-    end
+    qg = H_eig.vectors' * gr
 
     # These values describe the outcome of the subproblem.  They will be
     # set below and returned at the end.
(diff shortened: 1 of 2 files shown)


src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -33,17 +33,16 @@ function check_hard_case_candidate(H_eigv, qg)
         end
     end
 
-    hard_case, lambda_index - 1
```
nan2ge1: Returning `lambda_index` saves two flops (see L126 and L142).

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -33,17 +33,16 @@ function check_hard_case_candidate(H_eigv, qg)
...
+    hard_case, lambda_index
 end
 
-# Function 4.39 in N&W
-function p_sq_norm(lambda::T, min_i, n, qg, H_eig) where T
-    p_sum = zero(T)
-    H_eigvals = H_eig.values
+# Equation 4.38 in N&W
```
nan2ge1: This generalizes the computations of `p` or `s` and reduces some overhead (see below).

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -86,29 +85,30 @@ function solve_tr_subproblem!(gr,
     H_ridged = copy(H)
 
     # Cache the inner products between the eigenvectors and the gradient.
-    qg = similar(gr)
-    for i=1:n
-        qg[i] = dot(H_eig.vectors[:, i], gr)
-    end
+    qg = H_eig.vectors' * gr
```
nan2ge1: This one-liner improves the readability slightly IMHO.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -86,29 +85,30 @@ function solve_tr_subproblem!(gr,
...
 
     # These values describe the outcome of the subproblem.  They will be
     # set below and returned at the end.
     interior = true
     hard_case = false
     reached_solution = true
 
-    if min_H_ev >= 1e-8 && p_sq_norm(zero(T), 1, n, qg, H_eig) <= delta_sq
```
nan2ge1: Calling `p_sq_norm` introduces some overhead since `s` will be re-calculated in L104.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -86,29 +85,30 @@ function solve_tr_subproblem!(gr,
...
+    # Unconstrained solution
```
nan2ge1: Using `calc_p!` improves readability IMHO.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -86,29 +85,30 @@ function solve_tr_subproblem!(gr,
...
         reached_solution = true
-        s[:] = -(H_eig.vectors ./ H_eig.values') * H_eig.vectors' * gr
         lambda = zero(T)
     else
         interior = false
 
         # The hard case is when the gradient is orthogonal to all
         # eigenvectors associated with the lowest eigenvalue.
-        hard_case_candidate, min_H_ev_multiplicity =
+        hard_case_candidate, min_i =
```
nan2ge1: Renamed as `min_i` to be consistent with `calc_p!` (and the retired `p_sq_norm`).

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -123,12 +123,10 @@ function solve_tr_subproblem!(gr,
             # iterate on the boundary.
 
             # Formula 4.45 in N&W
-            p_lambda2 = p_sq_norm(lambda, min_H_ev_multiplicity + 1, n, qg, H_eig)
+            calc_p!(lambda, min_i, n, qg, H_eig, s)
```
nan2ge1: Using `calc_p!` to improve readability.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -123,12 +123,10 @@ function solve_tr_subproblem!(gr,
...
+            p_lambda2 = sum(abs2, s)
             if p_lambda2 > delta_sq
                 # Then we can simply solve using root finding.
-                # Set a starting point greater than the minimum based on the
-                # range between the largest and smallest eigenvalues.
-                lambda = lambda_lb + 0.01 * (max_H_ev - min_H_ev)
```
nan2ge1: I find this unnecessary because `lambda_lb` is already larger than `-min_H_ev` by a fraction of the eigenvalue gap, unless a 'significantly' larger value is desired. But this doesn't increase the chance of obtaining a `lambda` estimate larger than `lambda_lb` (or does it?) and the other case is taken care of anyway.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -137,28 +135,23 @@ function solve_tr_subproblem!(gr,
...
                 # I don't think it matters which eigenvector we pick so take
                 # the first.
-                for i=1:n
-                    s[i] = tau * H_eig.vectors[i, 1]
-                    for k=(min_H_ev_multiplicity + 1):n
-                        s[i] = s[i] +
-                               qg[k] * H_eig.vectors[i, k] / (H_eig.values[k] + lambda)
-                    end
-                end
+                calc_p!(lambda, min_i, n, qg, H_eig, s)
```
nan2ge1: Using `calc_p!` improves the readability slightly IMHO.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -137,28 +135,23 @@ function solve_tr_subproblem!(gr,
...
+                p[:] = -p + tau * H_eig.vectors[i, 1]
             end
         end
 
         if !hard_case
             # Algorithim 4.3 of N&W, with s insted of p_l for consistency with
             # Optim.jl
 
-            for i=1:n
```
nan2ge1: This is redundant to L175--178 and therefore merged and moved inside the loop.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -168,14 +161,10 @@ function solve_tr_subproblem!(gr,
 
                 # Check that lambda is not less than lambda_lb, and if so, go
                 # half the way to lambda_lb.
-                if lambda < (lambda_lb + 1e-8)
```
nan2ge1: Adding a fraction here is unnecessary since `lambda_lb` has already been perturbed in L116.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -319,7 +306,7 @@ function update_state!(d, state::NewtonTrustRegionState, method::NewtonTrustRegi
         # This can happen if the trust region radius is too large and the
         # Hessian is not positive definite.  We should shrink the trust
         # region.
-        state.rho = method.rho_lower - 1.0
```
nan2ge1: Setting `rho` to a negative number saves one flop.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -336,11 +323,11 @@ function update_state!(d, state::NewtonTrustRegionState, method::NewtonTrustRegi
         # The improvement is too small and we won't take it.
 
         # If you reject an interior solution, make sure that the next
-        # delta is smaller than the current step.  Otherwise you waste
+        # delta is smaller than the current step. Otherwise you waste
         # steps reducing delta by constant factors while each solution
         # will be the same.
         x_diff = state.x - state.x_previous
-        delta = 0.25 * sqrt(dot(x_diff, x_diff))
+        delta = 0.25 * norm(x_diff)
```
nan2ge1: This change improves the readability slightly IMHO.

pkofod:
I will look forward to reading this! Thanks

nan2ge1:
@pkofod Thanks! I look forward to contributing more to this package.

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: I doubt that matters for anything tbh :) If you remove the -1, does that not change the validity of the "doc string" above? If that's the case, you should update that.

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: Fine

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: Maybe this should just be changed to use booleans that describe the mechanism better? I don't know, let's keep it as it as you did it for now.

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: Is this supposed to say Equation 4.28? Can you explain what you mean by generalizes?

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: True. Two flops don't mater that much. Still I find returning `lambda_index` more intuitive and consistent with N&W. The docstring has been updated. I could also change it back if you would prefer. ;)

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: Is it possible that you are referring to the 1st edition of N&W? I was referring to (4.38) in the 2nd edition (2006). Shall we update on the reference? I can do the dirty work.

nan2ge1: `p` is calculated at different places in an (almost) identical way: `p_sq_norm` (previous L40--46), for the unconstrained solution (previous L104), and for the hard case (previous L140--146, with one tweak though). The idea is to write a generic function which can deal with all cases. This eases the maintenance I think.

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: `state.grow_region`, `state.shrink_region` (at most one of them is `true`) and `state.update_x`? Looks like an overkill to me...

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: No, this is fine. It was just either the comment had to be updated, or this reverted :)

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: That's not what I meant :) Basically what I meant was that we could just calculate rho, and then just combine the two if-elseif-else blocks that basically (but not exactly) checks the same thing twice. Instead of bothering to set rho to 1.0 or -1.0, we might as well just adjust and update right away.

This doesn't have to be now, let's just get your improvements in :)

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: I have the first edition yes. It's fine either way, but we might want to add a comment at the very top saying "throughout, N&W refers to the second edition blabla" :)

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: Understood. But then we'll have to compare `state.eta` with `method.rho_lower` (and maybe even `method.rho_upper`), unless the former is always smaller?

src/multivariate/solvers/second_order/newton_trust_region.jl
pkofod: Maybe I'm daft, but neither of the inputs have changed, have they? So `s`should already be updated.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -137,28 +135,23 @@ function solve_tr_subproblem!(gr,
...
                 # the first.
-                for i=1:n
-                    s[i] = tau * H_eig.vectors[i, 1]
-                    for k=(min_H_ev_multiplicity + 1):n
-                        s[i] = s[i] +
-                               qg[k] * H_eig.vectors[i, k] / (H_eig.values[k] + lambda)
-                    end
-                end
+                calc_p!(lambda, min_i, n, qg, H_eig, s)
+                p[:] = -p + tau * H_eig.vectors[i, 1]
```
pkofod: I guess no tests hit this, but you've changed `s`for `p` so this won't work.

pkofod: Oh, tests DID fail https://travis-ci.org/JuliaNLSolvers/Optim.jl/jobs/511117559#L402

pkofod:
Looks pretty good. I had some minor comments, but other than that, I guess you just need to fix the multiplicity tests (they still test for the old values).

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: Or we just update the reference in the current L218 (and 214)?

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: Fixed. Sorry about this blunder. But having both `p` and `s` in the same block can sometimes be quite confusing. :)

pkofod: If it's canonical to call `s` (which seems the norm in texts on line search based methods) `p` in trust region sub problems, we can call it `p` throughout, I don't care. Or just keep it :)

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: Fixed.

src/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -137,28 +135,23 @@ function solve_tr_subproblem!(gr,
...
                 # the first.
-                for i=1:n
-                    s[i] = tau * H_eig.vectors[i, 1]
-                    for k=(min_H_ev_multiplicity + 1):n
-                        s[i] = s[i] +
-                               qg[k] * H_eig.vectors[i, k] / (H_eig.values[k] + lambda)
-                    end
-                end
+                calc_p!(lambda, min_i, n, qg, H_eig, s)
+                s[:] = -s + tau * H_eig.vectors[i, 1]
```
pkofod: This `i` is not defined, it came from the loop. I guess you mean `:` ?

nan2ge1: Shoot, I must have copy&pasted...Fixed.

src/multivariate/solvers/second_order/newton_trust_region.jl
nan2ge1: I can add it to my to-do list. I think there are still several minor issues which I would rather leave to the next PR.

pkofod:
I think we may be down to the lambda_1_multiplicity tests now :)

nan2ge1:
Should be fixed by now. Next time I'll remember to run local tests first before submitting any PR.

pkofod:
Thanks!

PR #709, opened 2019-04-03
Typo in the documentation: Update nelder_mead.md

eirikeb:
Corrected a small typo. 

In Gao & Han (2010, p. 269) they recommend using gamma = 1 - 1/(2n)

Fortunately, the code seems correct in ` Optim.jl/src/multivariate/solvers/zeroth_order/
nelder_mead.jl`, line 28


pkofod:
Thanks! Unfortunately, I'm pretty sure I wrote that (docs and code), so I have to accept the responsibility :) Good that the code is correct at least.

PR #715, opened 2019-05-05
Initial stepnorm in BFGS

pkofod:
This is mentioned in N&W. They mention that various pieces of software allow a similar choice, but I'm not sure which ones to be honest.

The reason why it can make sense, as far as I understand it, is because extreme search directions (not final steps!) can really mess up the line search. It can be difficult to reach a useful step size in the allowed number of iterations, or a sub-optimal point can be accepted by the inexact line search.

`BFGS(; initial_stepnorm=z)` will scale the initial identity matrix with the supnorm of the gradient times `z`. That number `z` is usually described as the "usual" or "expected" step size. If the initial step size can be guessed and set to `z` it is potentially possible to either a) avoid too many function evaluations in the first iteration and/or b) avoid a weird step into a "bad" region in the first step.

As everything non-linear, this is not guaranteed to help anyone (say with `z=1`), but it's a convenient thing to be able to set by a user who experiences odd first steps.

### DIFF
--- a/docs/src/algo/lbfgs.md
+++ b/docs/src/algo/lbfgs.md
@@ -4,9 +4,18 @@ This page contains information about BFGS and its limited memory version L-BFGS.
 ```julia
 BFGS(; alphaguess = LineSearches.InitialStatic(),
        linesearch = LineSearches.HagerZhang(),
-       initial_invH = x -> Matrix{eltype(x)}(I, length(x), length(x)),
-       manifold = Flat()
+       initial_invH = nothing,
+       initial_stepnorm = nothing,
+       manifold = Flat())
 ```
+
+`initial_invH` has a default value of `nothing`. If the user has a specific initial
+matrix they want to supply, it should be supplied as a function of an array similar
+to the initial point `x0`.
+
+If `initial_stepnorm` is set to a number `z`, the initial matrix will be the
+identity matrix scaled by `z` times the sup-norm of the gradient at the initial
+point `x0`.
 
 ```julia
 LBFGS(; m = 10,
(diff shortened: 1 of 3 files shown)


antoine-levitt:
That makes sense - how come it's just for BFGS?

pkofod:
> That makes sense - how come it's just for BFGS?

Did you mean why not LBFGS as well?

antoine-levitt:
Yes, and all the first-order solvers.

pkofod:
Just the initial step, or?

antoine-levitt:
Well, I've always been confused by all stepsize selection works in Optim. What I'd like to do as a user is to tell the solver some "natural" stepsize (eg, if I want to optimize x^T A x - x^T b where A is the matrix of a finite difference approximation to some elliptic operator, I know how the step should scale as a function of the discretization parameter)

pkofod:
> Well, I've always been confused by all stepsize selection works in Optim. What I'd like to do as a user is to tell the solver some "natural" stepsize (eg, if I want to optimize x^T A x - x^T b where A is the matrix of a finite difference approximation to some elliptic operator, I know how the step should scale as a function of the discretization parameter)

What do you want to happen to that step size? I mean, are you just going to say `step size should be 0.3467` or do you want to provide a guess?

antoine-levitt:
I think the user should be able to provide a guess to help the solver figure out the right scale for the problem. If you really want to say that the step size should be some fixed value, you can switch to the `Static` linesearch.

pkofod:
> I think the user should be able to provide a guess to help the solver figure out the right scale for the problem

okay, but it's already possible to provide the initial step size provided to the line search procedure. Is this not what you're asking for?

antoine-levitt:
Oh sorry, I see that this PR is orthogonal to that mechanism. Forget I said anything.

PR #720, opened 2019-05-24
Fix a use-before-initialisation bug

andrewclausen:
After finding the solution, Optim.optimize() assembles some information, which
includes comparing the solution to the previous iteration.  However, if the
solution was found in one iteration (i.e. the initial guess was right), then
there is no previous guess to compare with.  In this case, Optim.optimize()
crashes, because it tries to use something that has not been initialised.

Here is a simple example that triggers a crash:

	using Optim
	f(x) = x[1]^2
	x0 = BigFloat[0]
	obj = OnceDifferentiable(f, x0; autodiff=:forward)
	result = Optim.optimize(f, x0, BFGS())

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -91,6 +91,12 @@ function optimize(d::D, initial_x::Tx, method::M,
     T = typeof(options.f_tol)
     f_incr_pick = f_increased && !options.allow_f_increases
 
+    if iteration < 2
+        # There is no previous state to compare with.
+        state.x_previous = NaN * initial_x
+        state.f_x_previous = NaN * value(d)
+    end
+
     return MultivariateOptimizationResults(method,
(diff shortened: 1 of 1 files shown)


pkofod:
Huh, thanks . It's wonderful that you're stress testing `optimize` with BigFloats, as I havn't used those too much to be honest, so there will be some corner cases where things like this can happen. This does not happen with most other numbers I'd guess. It has something to do with how a vector of initialized array of BigFloats works I think. We're using `similar(x)` in a lot of the initializers, but it seems that it is safer to init them with something like `fill(Tx(NaN), size(x))` or `similar(x).*Tx(NaN)` instead.

pkofod:
Thanks for reaching out. I've made a PR in https://github.com/JuliaNLSolvers/Optim.jl/pull/722 that should fix this (and added the tests to support that claim).

PR #728, opened 2019-06-15
RMSprop

krzysztofdabkowski1:
New multivariate solver based on GradientDescent and RMSprop algorithm

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -86,6 +86,7 @@ export optimize, maximize, # main function
        ### Acceleration methods
        AcceleratedGradientDescent,
        MomentumGradientDescent,
+       RMSprop,
 
        ### Nonlinear GMRES
        NGMRES,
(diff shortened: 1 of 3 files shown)


pkofod:
Thanks for the PR. I will have a look. I was actually just reading the steps to do Adam last night prompted by 

Is it really standard to perform a line search step?

krzysztofdabkowski1:
Nominally, RMSprop doesn't use line search, but I think it's not efficient (compared to accelerated variants of Gradient Descent). This method is combined of traditional Gradient Descent with line search and element calculated from RMSprop algorithm, which accelerate each iteration.

pkofod:
> but I think it's not efficient (compared to accelerated variants of Gradient Descent)

On what kinds of problems? RMSprop is meant for mini batches right? Are you sure this is good in the context of "stochastic" objectives as well?

krzysztofdabkowski1:
OK, probably you're right. I will correct it and write in traditional way.

pkofod:
> OK, probably you're right. I will correct it and write in traditional way.

I should say that the other accelerated methods (that are largely here for historical reasons) also default to a line search, but I don't think they should necessarily either.

krzysztofdabkowski1:
> I should say that the other accelerated methods (that are largely here for historical reasons) also default to a line search, but I don't think they should necessarily either.

So, you think that implementing it even with default to static doesn't make sense. Am I right?

pkofod:
> So, you think that implementing it even with default to static doesn't make sense. Am I right?

I think it would make sense to chose `Static()` as the standard.

PR #733, opened 2019-07-09
Add compats to Project.toml.

pkofod:


Project.toml
```diff
@@ -18,6 +18,15 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 julia = "1"
+Calculus = "0.3.1"
```
fredrikekre: Does it not work with the current Calculus version? Maybe you want
```suggestion
Calculus = "0.3.1, 0.4"
```

Project.toml
```diff
@@ -18,6 +18,15 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
...
+DiffEqDiffTools = "0.12"
```
fredrikekre:
```suggestion
DiffEqDiffTools = "0.12, 0.13"
```

Project.toml
pkofod: I wrote 3 by mistake, meant 4 :)

pkofod:
Thanks @fredrikekre ! The DiffEqDiffTools was me forgetting the "left most non-zero", that is, it isn't just the actual major version that determines the upper bound, and the Calculus one was a typo.

PR #742, opened 2019-08-23
only_* constructors for newton-krylov

### LINKED ISSUE #737: Document KrylovTrustRegion
Add KrylovTrustRegion + example in the docs

### LINKED ISSUE #738: Inconsistencies with fg! in Krylov vs LBFGS
This may also be an issue with NLSolversBase; I'm happy to try to help but would need guidelines for how to proceed. 

Currently a trick suggested with solvers like LBFGS is to implement a function `fg!` and pass it to `Optim.only_fg!(fg!)`; as per https://julianlsolvers.github.io/Optim.jl/stable/#user/tipsandtricks/ this function must be of the form

```
fg!(f, g, x)
    # common computations
    g === nothing || ... # in place update of g with gradient at x
    f === nothing || ... # return eval function at x
end
```

However when using the `KrylovTrustRegion` optimizer, we have to construct a `TwiceDifferentiableHV` taking `f, fg!, Hv!` where this time `fg!` is of the form

```
fg!(g, x)
    # common computations
    ... # in place update of g with gradient at x
    ... # return eval function at x
end
```

this inconsistency is a bit annoying when trying to expose both Krylov and LBFGS to the user and we have to define an ugly function like

```
fg_for_krylov! = (g, x) -> fg!(0.0, g, x)
```

### suggestions

1. instead of passing `f, fg!, Hv!` to `TwiceDifferentiableHV`, just pass `fg!, Hv!` with the same convention as for `only_fg!`
2. allow `TwiceDifferentiableHV` to be constructed with a single function `fghv!` for the same reason as the `only_fg!` trick: usually the computation of `Hv` can re-use some stuff from the computation of the gradient (in fact it's a common trick in ML where you compute `Hv` by computing `d/dr g(x+rv)` at r=0 (see e.g. http://www.bcl.hamilton.ie/~barak/papers/nc-hessian.pdf) 

thoughts?

tlienart:
This goes with the PR https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/109 to NLSolverBase (and should only be merged after that PR has been merged); it allows the use of

* `only_fg_and_hv!(fg!, hv!)` and
* `only_fghv!(fghv!)` 

where

```
fg!(F, G, x) = (G === nothing || G .= ... ; F === nothing || return some_f(x))
hv!(Hv, x, v) = (Hv .= ...)
```

with Newton-Krylov.

Tests pass locally apart from 1 unrelated failure with LBFGS. 

closes #738 as far as I'm concerned.

### DIFF
--- a/test/multivariate/solvers/second_order/krylov_trust_region.jl
+++ b/test/multivariate/solvers/second_order/krylov_trust_region.jl
@@ -6,18 +6,27 @@
         (x[1] - 5.0)^4
     end
 
-    function fg!(storage::Vector, x::Vector)
-        storage[1] = 4.0 * (x[1] - 5.0)^3
+    function fg!(g, x)
+        g[1] = 4.0 * (x[1] - 5.0)^3
         f(x)
     end
 
-    function hv!(storage::Vector, x::Vector, v::Vector)
-        storage[1] = 12.0 * (x[1] - 5.0)^2 * v[1]
+    function fg2!(_f, g, x)
+        g  === nothing || (g[1] = 4.0 * (x[1] - 5.0)^3)
+        _f === nothing || return f(x)
+    end
+
+
+    function hv!(Hv, x, v)
+        Hv[1] = 12.0 * (x[1] - 5.0)^2 * v[1]
     end
 
     d = Optim.TwiceDifferentiableHV(f, fg!, hv!, [0.0])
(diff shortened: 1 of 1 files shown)


tlienart:
Now that the PR JuliaNLSolvers/NLSolversBase.jl#109  has been merged, this can be considered though it won't work before a patch release of NLSolversBase I reckon

pkofod:
Better late than never. Thanks

PR #745, opened 2019-09-09
Add iterator interface

### LINKED ISSUE #599: Offer iterator interface
Using iterator interface, we can make the optimizer like a white box,

```julia
for info in optimize_iterator(func, x0, solver)
    println("Step = ", info["istep"])
    println("x = ", info["x"])
    println("f = ", info["f"])
   solver.PARAMETER = PARAMETER
    ...
end
```
This kind of interface is very welcome is debugging, tunning optimizer et. al.
Please consider ~

A python version is of white box optimizer is https://github.com/BRML/climin

tkf:
This PR adds an iterator interface for optimizer.  I tentatively call it `Optim.optimizing`.  It provides an iterator-based API such that

```julia
let istate
    for istate′ in optimizing(args...; kwargs...)
        istate = istate′
    end
    OptimizationResults(istate)
end
```

is equivalent to `optimize(args...; kwargs...)` for multivariate optimizations.

close #599


src/multivariate/optimize/optimize.jl
```diff
@@ -27,36 +27,73 @@ function initial_convergence(d, state, method::AbstractOptimizer, initial_x, opt
...
+end
+
+Base.IteratorSize(::Type{<:OptimIterator}) = Base.SizeUnknown()
+Base.IteratorEltype(::Type{<:OptimIterator}) = Base.HasEltype()
+Base.eltype(::Type{<:OptimIterator}) = IteratorState
+
+@with_kw struct IteratorState{IT <: OptimIterator, TR <: OptimizationTrace}
+    # Put `OptimIterator` in iterator state so that `OptimizationResults` can
+    # be constructed from `IteratorState`.
+    iter::IT
```
tkf: We can drop `iter` field from `IteratorState` if we change the API to:

```julia
let istate
    iter = optimizing(args...; kwargs...)
    for istate′ in iter
        istate = istate′
    end
    OptimizationResults(iter, istate)  # need to pass `iter` here
end
```

pkofod:
Interesting, thanks. That took surprisingly few changes. Will have to have a more thorough look!

src/multivariate/optimize/optimize.jl
```diff
@@ -85,7 +125,35 @@ function optimize(d::D, initial_x::Tx, method::M,
...
+    )
+
+    return new_istate, new_istate
+end
+
+function OptimizationResults(istate::IteratorState)
+    @unpack_IteratorState istate
+    @unpack d, initial_x, method, options, state = iter
 
     after_while!(d, state, method, options)
```
tkf: Executing a mutating-function `after_while!` inside non-bang function `OptimizationResults` is not super great.  But it looks like `after_while!` is a no-op mostly except for NelderMead so maybe it's OK?  From a quick look, `after_while!` for `NelderMead` seems to be idempotent (so that, e.g., calling `OptimizationResults(istate)` inside a loop multiple times is OK).  If that's the case, `OptimizationResults(istate)` practically has no side-effect?

But using a function like `result!` sounds good to me as well.

src/multivariate/optimize/optimize.jl
```diff
@@ -85,15 +125,50 @@ function optimize(d::D, initial_x::Tx, method::M,
...
+        iteration,
+        ls_success,
+    )
+
+    return new_istate, new_istate
+end
+
+function OptimizationResults(istate::IteratorState)
+    @unpack d, method, options, state = istate.iter
     after_while!(d, state, method, options)
```
tkf: Executing a mutating-function `after_while!` inside non-bang function `OptimizationResults` is not super great.  But it looks like `after_while!` is a no-op mostly except for NelderMead so maybe it's OK?  From a quick look, `after_while!` for `NelderMead` seems to be idempotent (so that, e.g., calling `OptimizationResults(istate)` inside a loop multiple times is OK).  If that's the case, `OptimizationResults(istate)` practically has no side-effect?

But using a function like `result!` sounds good to me as well.

src/multivariate/optimize/optimize.jl
tkf: Executing a mutating-function `after_while!` inside non-bang function `OptimizationResults` is not super great.  But it looks like `after_while!` is a no-op mostly except for NelderMead so maybe it's OK?  From a quick look, `after_while!` for `NelderMead` seems to be idempotent (so that, e.g., calling `OptimizationResults(istate)` inside a loop multiple times is OK).  If that's the case, `OptimizationResults(istate)` practically has no side-effect?

But using a function like `result!` sounds good to me as well.

src/api.jl
```diff
@@ -23,41 +26,45 @@ function x_upper_trace(r::UnivariateOptimizationResults)
...
     end
     !haskey(tr[1].metadata, "x") && error("Trace does not contain x. To get a trace of x, run optimize() with extended_trace = true")
     [ state.metadata["x"] for state in tr ]
 end
 
-function centroid_trace(r::MultivariateOptimizationResults)
-    if !isa(r.method, NelderMead)
-        throw(ArgumentError("There is no centroid involved in optimization using $(r.method). Please use x_trace(...) to grab the points from the trace."))
+function centroid_trace(r::Union{MultivariateOptimizationResults, IteratorState})
+    tr = trace(r)
```
tkf: I suppose `tr = trace(r)` should be added here? I think it'll throw `UndefVarError` otherwise. There are two more places I did this change.

tkf:
@pkofod After posted the PR, I realized that the first version didn't provide the way to access iteration state (e.g., getting "best minimizer so far") so I added accessor functions for iterator state by reusing the function names/code already defined for the result type.  So the diff may not be as small as you remember now.

Anyway, I think I've implemented the APIs I need.  I'm looking forward to hear your thoughts on this.

src/multivariate/optimize/optimize.jl
tkf: Accessor functions like `iteration_limit_reached(istate)` and `(f|g|h)_calls(istate)` need `.iter`, too.

tkf:
Aside: Using this API a few times, I realized that `istate = istate′` pattern for leaking the iteration variable to the outer scope is rather ugly although I guess there is not much we can do.  For a subset of problem, defining `foreach` like below is kind of useful.

```julia
function Base.foreach(f, iter::OptimIterator)
    local istate
    for istate′ in iter
        istate = istate′
        f(istate)
    end
    return istate
end
```

We can then use it as

```julia
result = foreach(optimizing(args...; kwargs...)) do istate
    @info "Minimizing..." minimizer(istate) minimum(istate)
end |> Optim.OptimizationResults
```

although this does not let us use `break`.

(Or we can implement [transducer-compatible `foldl`](https://tkf.github.io/Transducers.jl/dev/manual/#Base.foreach) so that `foreach` is as powerful as `for` loop :wink:)

tkf:
Friendly ping :)

Is there anything I can do to make it easier to merge?

pkofod:
Sorry, I've had too much on my plate. I'll take a look soon.

tkf:
Thanks!

src/multivariate/optimize/optimize.jl
```diff
@@ -29,42 +29,80 @@ function initial_convergence(d, state, method::AbstractOptimizer, initial_x, opt
...
+    method::M
+    options::O
+    state::S
+end
+
+Base.IteratorSize(::Type{<:OptimIterator}) = Base.SizeUnknown()
+Base.IteratorEltype(::Type{<:OptimIterator}) = Base.HasEltype()
+Base.eltype(::Type{<:OptimIterator}) = IteratorState
+
+@with_kw struct IteratorState{IT <: OptimIterator, TR <: OptimizationTrace}
```
pkofod, line 48:
```suggestion
struct IteratorState{IT <: OptimIterator, TR <: OptimizationTrace}
```

pkofod (approved):

src/multivariate/optimize/optimize.jl
```diff
@@ -29,42 +29,98 @@ function initial_convergence(d, state, method::AbstractOptimizer, initial_x, opt
...
+    f_increased::Bool
+    counter_f_tol::Int
+    g_converged::Bool
+    converged::Bool
+    iteration::Int
+    ls_success::Bool
+end
+
+function Base.iterate(iter::OptimIterator, istate = nothing)
+    @unpack d, initial_x, method, options, state = iter
```
pkofod, line 74:
```suggestion
    (;d, initial_x, method, options, state) = iter
```

src/multivariate/optimize/optimize.jl
```diff
@@ -103,6 +159,36 @@ function optimize(
...
+        iteration,
+        ls_success,
+    )
+
+    return new_istate, new_istate
+end
+
+function OptimizationResults(istate::IteratorState)
+    @unpack_IteratorState istate
+    @unpack d, initial_x, method, options, state = iter
```
pkofod, line 191:
```suggestion
    (;d, initial_x, method, options, state) = iter
```

src/multivariate/optimize/optimize.jl
```diff
@@ -220,3 +305,45 @@ function _termination_code(d, gres, state, stopped_by, options)
...
+function minimizer(istate::IteratorState)
+    @unpack iter, f_increased = istate
+    @unpack options, state = iter
+    f_incr_pick = f_increased && !options.allow_f_increases
+    return pick_best_x(f_incr_pick, state)
+end
+
+function minimum(istate::IteratorState)
+    @unpack iter, f_increased = istate
+    @unpack d, options, state = iter
```
pkofod, line 332:
```suggestion
    (;iter, f_increased) = istate
    (;d, options, state) = iter
```

src/multivariate/optimize/optimize.jl
pkofod: I got rid of after_while! so no more side effects

src/multivariate/optimize/interface.jl
```diff
@@ -313,17 +309,29 @@ function optimize(
...
+end
+
+function optimize(args...; kwargs...)
+    local istate
+    iter = optimizing(args...; kwargs...)
+    for istate′ in iter
+        istate = istate′
+    end
+    # We can safely assume that `istate` is defined at this point.  That is to say,
+    # `OptimIterator` guarantees that `iterate(::OptimIterator) !== nothing`.
```
devmotion: I think JET won't agree with this comment 😄 

Generally, the code above seems a bit unfortunate... Maybe `optimizing` should return the iterator AND the initial state?

I also wonder, is there no utility in Julia for directly obtaining the last state of an iterator?

src/multivariate/optimize/optimize.jl
```diff
@@ -32,40 +30,100 @@ end
...
+    counter_f_tol::Int
+    g_converged::Bool
+    converged::Bool
+    iteration::Int
+    ls_success::Bool
+end
+
+function Base.iterate(iter::OptimIterator, istate = nothing)
+    (; d, initial_x, method, options, state) = iter
+    if istate === nothing
```
devmotion: IMO it would be cleaner to move this code to a separate `Base.iterate(iter::OptimIterator)` definition, I'm not a fan of "dummy" states such as `nothing` for the initial iteration as this is not part of the `Base.iterate` interface.

src/multivariate/optimize/optimize.jl
```diff
@@ -32,40 +30,100 @@ end
...
+    initial_x::Tx
+    method::M
+    options::O
+    state::S
+end
+_method(r::OptimIterator) = r.method
+
+Base.IteratorSize(::Type{<:OptimIterator}) = Base.SizeUnknown()
+Base.IteratorEltype(::Type{<:OptimIterator}) = Base.HasEltype()
+Base.eltype(::Type{<:OptimIterator}) = IteratorState
```
devmotion: Is it a problem that the element type is non-concrete here? Could it be defined in a way that the `eltype` is concrete?

devmotion: Seems below `TR` is initialized as `OptimizationTrace{typeof(value(d)),M}`.

PR #755, opened 2019-09-28
Fix successive_f_tol

tkf:
Considering that `converged` is set as

https://github.com/JuliaNLSolvers/Optim.jl/blob/64eb60060a4c1b310207e56ce5e81d41ddd7c75c/src/utilities/assess_convergence.jl#L48

and how `successive_f_tol` is used

https://github.com/JuliaNLSolvers/Optim.jl/blob/64eb60060a4c1b310207e56ce5e81d41ddd7c75c/src/multivariate/optimize/optimize.jl#L67-L68

it seems that `successive_f_tol` has no effect.  That is to say, if `f_converged == true`, `converged = converged | (counter_f_tol > options.successive_f_tol)` is going to be `true` even if `counter_f_tol <= options.successive_f_tol`.

So, maybe the intent was to use `&&` (or `&`) like this?

https://github.com/JuliaNLSolvers/Optim.jl/blob/59e2007e844bbe3febd9e4f6a4c85e41ef5a1788/src/multivariate/optimize/optimize.jl#L68

However, this patch breaks some tests and I haven't looked into why that's the case.  Maybe it's just me misunderstanding `successive_f_tol`?  It is supposed to be used only with `IPNewton`?


pkofod:
I think the problem here is that f_converged shouldn't be part of converged if we write it like this. I honestly can't even remember this check being added... 

Shouldn't we just put the default value to 1 and check that counter >= option? It was added here btw https://github.com/JuliaNLSolvers/Optim.jl/pull/533

tkf:
> Shouldn't we just put the default value to 1 and check that counter >= option?

Is this what you suggest?

https://github.com/JuliaNLSolvers/Optim.jl/blob/9bfb1f91fbda93b5e394ff7c994ef2be2aa8f0ca/src/multivariate/optimize/optimize.jl#L68

If this is the right approach, should I do this for `ipnewton/interior.jl` (which is added in #608)?

pkofod:
> If this is the right approach, should I do this for `ipnewton/interior.jl` (which is added in #608)?

Exactly! Yes, that sounds reasonable. Tests also pass now (?)

tkf:
> Tests also pass now (?)

I think this is because the semantics of `successive_f_tol` is changed. At the beginning of the PR, it was:

```julia
converged = (x_converged || g_converged) && (counter_f_tol > options.successive_f_tol) 
```
Now it's

```julia
converged = x_converged || g_converged || (counter_f_tol > options.successive_f_tol) 
```

pkofod:
Awesome, thanks!

PR #756, opened 2019-10-03
Fix: linesearch result was not checked in `optimize`

tkf:
`ls_failed = update_state!(d, state, method)` assigned in `optimize` is never used. IIUC we need to bail out once it has failed and I _think_ the code with this PR does this.

Does this patch make sense?  Locally running the test fails with multivariate Fminbox testset so I'm not sure if there is a bug with this patch or there is a glitch in the test.

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -50,12 +50,12 @@ function optimize(d::D, initial_x::Tx, method::M,
     options.show_trace && print_header(method)
     _time = time()
     trace!(tr, d, state, iteration, method, options, _time-t0)
-    ls_success::Bool = true
+    ls_failed::Bool = false
     while !converged && !stopped && iteration < options.iterations
         iteration += 1
 
         ls_failed = update_state!(d, state, method)
-        if !ls_success
+        if ls_failed
             break # it returns true if it's forced by something in update! to stop (eg dx_dg == 0.0 in BFGS, or linesearch errors)
         end
         update_g!(d, state, method) # TODO: Should this be `update_fg!`?
@@ -119,7 +119,7 @@ function optimize(d::D, initial_x::Tx, method::M,
                                         f_calls(d),
                                         g_calls(d),
                                         h_calls(d),
-                                        !ls_success,
+                                        !ls_failed,
                                         options.time_limit,
                                         _time-t0,
                                         )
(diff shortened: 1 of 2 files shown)


src/multivariate/optimize/optimize.jl
```diff
@@ -119,7 +119,7 @@ function optimize(d::D, initial_x::Tx, method::M,
                                         f_calls(d),
                                         g_calls(d),
                                         h_calls(d),
-                                        !ls_success,
+                                        !ls_failed,
                                         options.time_limit,
                                         _time-t0,
                                         )
```
tkf: The argument I changed is the third to the last field of `MultivariateOptimizationResults` which is `ls_success`. So I think flipping the Boolean is correct:

https://github.com/JuliaNLSolvers/Optim.jl/blob/80bbd32d2198b273cd6a581fdc4c82ff3a6cafeb/src/types.jl#L191-L194

Before the PR, it was always `!true` so all the failure was reported as line search failure.

pkofod:
I love it when one os fails and another one doesn't 😑

tkf:
It fails in Linux but succeeds in macOS:
https://travis-ci.org/JuliaNLSolvers/Optim.jl/builds/592911800

It fails in x64 Windows but succeeds in x64:
https://travis-ci.org/JuliaNLSolvers/Optim.jl/builds/592911800

The failure mode is always:

```
┌ Warning: f(x) increased: stopping optimization
└ @ Optim ~/build/JuliaNLSolvers/Optim.jl/src/multivariate/solvers/constrained/fminbox.jl:311
Fminbox: Test Failed at /home/travis/build/JuliaNLSolvers/Optim.jl/test/multivariate/solvers/constrained/fminbox.jl:44
  Expression: Optim.converged(results)
```

https://travis-ci.org/JuliaNLSolvers/Optim.jl/jobs/592911803#L402-L405
https://travis-ci.org/JuliaNLSolvers/Optim.jl/jobs/592911801#L399-L402
https://ci.appveyor.com/project/pkofod/optim-jl/builds/27845368/job/la4559ufy5ucc3on#L181

tkf:
I tried a dumb workaround to pass `allow_f_increases=true` but it didn't work... 822d3635301c99b4604ca3b4ce1612da4851ae1d

pkofod:
> sure if there is a bug with this patch or there is a glitch in the test.

The test is just a regression test for behavior... I will get back to this, I promise.

pkofod:
closed in #790

tkf:
Thanks a lot for fixing this!

PR #764, opened 2019-11-12
Nelder mead stopping

jwscook:
Nelder Mead was taking a long time to converge to a value that doesn't fall below the `f_abstol`. Also the vertices of the simplex are within `x_abstol` of each other without exiting. This pull request fixes those problems and keeps the default behaviour **only** if the default "f" and "x" tolerances are changed to zero. These pull request does not pass the test because of this, I think.

### DIFF
--- a/docs/src/algo/nelder_mead.md
+++ b/docs/src/algo/nelder_mead.md
@@ -27,10 +27,10 @@ we can perform one of four actions: reflect, expand, contract, or shrink. Basica
 the goal is to iteratively replace the worst point with a better point. More information
 can be found in Nelder and Mead (1965), Lagarias, et al (1998) or Gao and Han (2010).
 
-The stopping rule is the same as in the original paper, and is the standard
+The stopping rules encompass that of the original paper, and is the standard
 error of the function values at the vertices. To set the tolerance level for this
 convergence criterion, set the `g_tol` level as described in the Configurable Options
-section.
+section. The solver will also stop if one of the objective function evaluations is below the `f_tol` level, or the simplex locations are within `x_tol`. 
 
 When the solver finishes, we return a minimizer which is either the centroid or one of the vertices.
 The function value at the centroid adds a function evaluation, as we need to evaluate the objection
--- a/docs/src/algo/nelder_mead.md
+++ b/docs/src/algo/nelder_mead.md
@@ -27,10 +27,13 @@ we can perform one of four actions: reflect, expand, contract, or shrink. Basica
 the goal is to iteratively replace the worst point with a better point. More information
 can be found in Nelder and Mead (1965), Lagarias, et al (1998) or Gao and Han (2010).
 
-The stopping rules encompass that of the original paper, and is the standard
+The stopping rules includes that of the original paper, and is the standard
 error of the function values at the vertices. To set the tolerance level for this
-convergence criterion, set the `g_tol` level as described in the Configurable Options
-section. The solver will also stop if one of the objective function evaluations is below the `f_tol` level, or the simplex locations are within `x_tol`. 
+convergence criterion, set the `g_abstol` level as described in the Configurable 
+Options section. The solver will also stop if one of the objective function
+evaluations is below the `f_abstol` level, or the simplex locations are within
+`x_abstol` of each other. Set `f_abstol` and `g_abstol` both to zero to
+replicate the functionality descibed in the original paper.
 
 When the solver finishes, we return a minimizer which is either the centroid or one of the vertices.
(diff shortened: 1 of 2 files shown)


pkofod:
Sorry, I've been very busy. `f` tolerances are not meant to be measured against 0, but is used to control the difference between two consecutive values. That's why it has no effect in NelderMead, as the best value doesn't have to change for the step to be successful.

jwscook:
What about `f_abstol`?

The current behaviour of Nelder Mead is to stop only when the standard deviation of the objective function values of the vertices is below the specified `g` tolerance, in its `assess_convergence` function. It does not account for the fact that any one of the values might be below the value of `f_abstol`. Nlopt does not suffer from this problem.

pkofod:
> The current behaviour of Nelder Mead is to stop only when the standard deviation of the objective function values of the vertices is below the specified `g` tolerance, in its `assess_convergence` function. It does not account for the fact that any one of the values might be below the value of `f_abstol`. Nlopt does not suffer from this problem.

It's not a problem. We don't define the tolerances like that. `f_abstol` is the smallest absolute *change* in objective value we allow between two iterations for the iterations to continue - but *still* not in NelderMead, we only expose the simplex "size" tolerance for that.

If you want, I can show you how to write a simple callback that stops if the objective value is sufficiently small, and we can also add it to the documentation. Or we can potentially add a *new* stopping criterion called `f_target` or something like that, that allows you to specify a target f that you'll be sufficiently happy with.

jwscook:
> It's not a problem.

The implementation in NLopt is more efficient because it does not for all the values on the vertices to get below some threshold, and it does allow the user to stop the optimisation bases on the value of `f`. In my use case NLopt is takes 2hrs, and Optim takes 5 hrs.

> If you want, I can show you how to write a simple callback that stops if the objective value is sufficiently small, and we can also add it to the documentation. Or we can potentially add a _new_ stopping criterion called `f_target` or something like that, that allows you to specify a target f that you'll be sufficiently happy with.

Yes please; how is this done?

pkofod:
> The implementation in NLopt is more efficient because it does not for all the values on the vertices to get below some threshold, and it does allow the user to stop the optimisation bases on the value of `f`. In my use case NLopt is takes 2hrs, and Optim takes 5 hrs.

That has nothing to do with efficiency though - you're comparing two different things. It may be better for your usecase, but in full generality you don't *know* the value at the optimum unless you're in some special situation. Are you doing least squares or something similar where you know that the theoretical bound is 0?

> Yes please; how is this done?

I'm jumping on my bike now, but I'll write up the example when I get home.

pkofod:
You have to ignore the `Statu: failure (line search failed)` part, that's a but that'll be fixed separately. It basically just means that you stopped it manually. Set up a test function
```
function exponential(x)
    return exp((2.0 - x[1])^2) + exp((3.0 - x[2])^2)
end
```
Solve it first with then without a callback
```
julia> optimize(exponential, [7.0, 3.0], NelderMead(), Optim.Options(callback=x->(ifelse(x.value<2.001,true,false))))
 * Status: failure (line search failed)

 * Candidate solution
    Minimizer: [2.00e+00, 3.01e+00]
    Minimum:   2.000063e+00

 * Found with
    Algorithm:     Nelder-Mead
    Initial Point: [7.00e+00, 3.00e+00]

 * Convergence measures
    √(Σ(yᵢ-ȳ)²)/n ≰ 1.0e-08

 * Work counters
    Iterations:    17
    f(x) calls:    39


julia> optimize(exponential, [7.0, 3.0], NelderMead()))
ERROR: syntax: extra token ")" after end of expression
Stacktrace:
 [1] top-level scope at REPL[19]:0

julia> optimize(exponential, [7.0, 3.0], NelderMead())))
ERROR: syntax: extra token ")" after end of expression
Stacktrace:
 [1] top-level scope at REPL[19]:0

julia> optimize(exponential, [7.0, 3.0], NelderMead())
 * Status: success

 * Candidate solution
    Minimizer: [2.00e+00, 3.00e+00]
    Minimum:   2.000000e+00

 * Found with
    Algorithm:     Nelder-Mead
    Initial Point: [7.00e+00, 3.00e+00]

 * Convergence measures
    √(Σ(yᵢ-ȳ)²)/n ≤ 1.0e-08

 * Work counters
    Iterations:    34
    f(x) calls:    71
```
As you can see, it stops early. The actual callback can be written like this for clarity
```
function cb(tr)
    if tr.value < 2.001
        return true # means: should we stop? true
    else
        return false # means: should we stop? false
    end
end
```

PR #774, opened 2020-01-13
add broken test for constrained minimum Gross-Pitaevskii

thisrod:
I'm working on ground states for Bose-Einstein condensates.  Antoine Levitt has been encouraging me to explore using the constrained optimisation routines in Optim to find minimum-energy solutions of the Gross-Pitaevskii equation.

The obvious way doesn't work, so I'm submitting it as a broken test case.

I would welcome advice on how to make these modifications:

1. I considered contributing this to OptimTestProblems, but I don't know how to start.

2. BandedMatrices should be added as a test dependency, but again I don't know how that works.

3. I expect that it would be helpful to provide a solution, not just the problem.  My first thought is to put a numerical solution in a JLD file, and an @assert in vortex.jl to show that it solves the problem.  Is there a better way?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "429524aa-4258-5aef-a3af-852621145aeb"
 version = "0.19.7"
 
 [deps]
+BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LineSearches = "d3d80556-e9d4-5f37-9878-2ab0fcc64255"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/test/multivariate/vortex.jl
+++ b/test/multivariate/vortex.jl
@@ -1,0 +1,64 @@
+# Find a stationary solution to the Gross-Pitaevskii equation by
+# minimizing the energy of a normalised wave function in a rotating
+# frame
+
+@testset "Gross-Pitaveskii vortex" begin
+    C = 10
+    μ = 10
+    Ω = 2*0.575
+    h = 0.2
+    N = 40
+
+    # Finite difference matrices.  ∂ on left is ∂y, ∂' on right is ∂x
+    
+    function op(stencil)
+        mid = (length(stencil)+1)÷2
+        diags = [i-mid=>fill(stencil[i],N-abs(i-mid)) for i = keys(stencil)]
+        BandedMatrix(Tuple(diags), (N,N))
+    end
+    
+    ∂ = (1/h).*op([-1/60, 3/20, -3/4, 0, 3/4, -3/20, 1/60])
+    ∂² = (1/h^2).*op([1/90, -3/20, 3/2, -49/18, 3/2, -3/20, 1/90])
+
+    y = h/2*(1-N:2:N-1)
+    x = y'
+    z = Complex.(x,y)
+    V = r² = abs2.(z)
+    
+    init = Complex.(exp.(-r²/2)/√π)
+    init = conj.(z).*init./sqrt(1 .+ r²)
+    init ./= norm(init)
+    
+    # The energy that the steady state minimizes
+    function E(ψ)
+        Lψ = -∂²*ψ-ψ*∂²+V.*ψ+C*abs2.(ψ).*ψ-1im*Ω*(y.*(ψ*∂')-x.*(∂*ψ))
+        sum(conj.(ψ).*Lψ)/norm(ψ)^2 |> real
+    end
+    
+    # The residual for the GPE, zero in a steady state
+    function rdl(ψ)
+        Lψ = -∂²*ψ-ψ*∂²+V.*ψ+C*abs2.(ψ).*ψ-1im*Ω*(y.*(ψ*∂')-x.*(∂*ψ))
+        E = sum(conj.(ψ).*Lψ)/norm(ψ)^2 |> real
+        norm(Lψ-E*ψ)/norm(ψ)
+    end
+    
+    # break out real and imaginary parts, constraint is unit sphere
+    function reconstruct(xy)
+        n = length(xy) ÷ 2
+        ψ = xy[1:n] + 1im*xy[n+1:end]
+        m = sqrt(n) |> Int
+        ψ = reshape(ψ,m,m)
+    end
+    
+    # norm taken from SOR solution
+    cost(xy) = E(20.31*reconstruct(xy))
+    
+    result = optimize(cost, [real.(init[:]); imag.(init[:])], NelderMead(manifold=Sphere()))
+    ψ₂ = 20.31*reconstruct(result.minimizer)
+    
+    # 3000 steps of successive over-relaxation gets a residual of 1e-3.
+    # The default settings in Optim should beat that.
+    
+    @test_broken rdl(ψ₂) < 1e-3
+
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,8 +13,13 @@ import NLSolversBase
 import NLSolversBase: clear!
 import LinearAlgebra: norm, diag, I, Diagonal, dot, eigen, issymmetric, mul!
 import SparseArrays: normalize!, spdiagm
+import BandedMatrices
+import BandedMatrices: BandedMatrix
 
 debug_printing = false
+
+        println("Vortex")
+        @time include("multivariate/vortex.jl)
 
 special_tests = [
     "bigfloat/initial_convergence",
@@ -91,6 +96,7 @@ multivariate_tests = [
     "arbitrary_precision",
     "successive_f_tol",
     "f_increase",
+    "vortex",
 ]
 multivariate_tests = map(s->"./multivariate/"*s*".jl", multivariate_tests)
 


thisrod:
When I run `julia --project` in `~/.julia/dev/Optim/test`, and do `import BandedMatrices` at the command prompt, it works.  But when `pkg"test Optim"` does the same thing, it gets a `Package BandedMatrices not found in current path` error.  How can that happen?

antoine-levitt:
I do it here, so it's definitely possible : https://github.com/JuliaMolSim/DFTK.jl/pull/93

I don't think neldermead works here. Can you try a gradient based method? You need to pass it the energy gradient (note : this is the unconstrained gradient, so without the lambda/mu term). Look in tests/manifolds. You don't need to break up the real and imaginary part, and optim can work with 2d matrices directly.

pkofod:
There's no nelder mead support for manifolds, so that's one problem. Can you describe what you mean by "broken"?

thisrod:
Thanks for looking at this.

My ultimate goal is to benchmark some theoretically nice optimisation methods against the kludgy but effective Gauss-Seidel type methods that physicists use to find vortex ground states.

> Can you try a gradient based method?

Yes.  When I have a spare afternoon, I'll derive the gradient and try.  It will take a big gain in efficiency to sell that to other physicists, because none of the usual ways to solve this problem require a gradient. 

Since the cost (aka energy) is a non-analytic function of the order parameter, I guess that taking the gradient with respect to real and imaginary parts is the only way to go?

> Can you describe what you mean by "broken"?

I used `@test_broken`, because a solution exists, but my test code fails to find it.

> There's no nelder mead support for manifolds, so that's one problem.

It would have helped if an error was thrown when I tried that.

antoine-levitt:
My god, anybody still doing Gauss Seidel? For sure using a better optimizer will help, so will using a preconditionner. I have a paper on that. 

Taking the gradient wrt real and imag can be done implicitly by taking the gradient "wrt zbar", see in the optim docs.

> because none of the usual ways to solve this problem require a gradient.

Sure they do: the gradient is nothing else than the mean field hamiltonian applied to psi. 

We implement it all with several other bells and whistles in DFTK in a plane wave basis (aka pseudospectral method), exploiting the fact that gross pitaevskii is mathematically similar to Kohn Sham with 1 electron. I used LBFGS for the optimization. The code is in the branch in the above PR. Let me know if you are interested and need help getting it to work.

pkofod:
> It would have helped if an error was thrown when I tried that.

what happened?

thisrod:
> > It would have helped if an error was thrown when I tried that.
> what happened?

When I ran

    result = optimize(cost, guess, NelderMead(manifold=Sphere()))

I got a result object, and a message saying that no solution had been found.  `result.minimizer` was similar to `guess`, but not identical.

If `NelderMead(manifold=Sphere())` had thrown an error, I would have realised what I was doing wrong.

thisrod:
> My god, anybody still doing Gauss Seidel?

It's simple and it works.  None of the current alternatives have both those properties.

> Taking the gradient wrt real and imag can be done implicitly by taking the gradient "wrt zbar", see in the optim docs.

> Sure they do: the gradient is nothing else than the mean field hamiltonian applied to psi.

Thanks.  Wirtinger derivatives make this a lot easier, and I'll try it later today.

> The code is in the branch in the above PR. Let me know if you are interested and need help getting it to work.

Thanks for the offer.  I looked at that and couldn't see how to start, so I'll try Optim first.  Also, the vortex experts I work with are suspicious of Fourier pseudospectral derivatives, because of the seized gears problem at the periodic boundary.

thisrod:
Thanks for the help.  I've gotten the conjugate gradient solver to work.  It runs impressively fast, and finds a wave function two vortices, energy 7.03 and GPE residual 5e-5.  The solver reports that it hasn't found a minimum, but that sounds pretty close to me.

[Conjugate gradient solution](https://github.com/JuliaNLSolvers/Optim.jl/files/4062703/resp200115a.pdf)

However, Gauss-Seidel does better, finding a wave function with a single vortex, energy 6.98 and residual 1e-3 after 4000 steps.

[Gauss-Seidel solution](https://github.com/JuliaNLSolvers/Optim.jl/files/4062705/resp200115b.pdf)

Here's the code.  Would you like me to set up a few vortex calculations as test problems?

[opbench.jl](https://gist.github.com/thisrod/896d60bbad411f8feaed87d03cbbe038)

I found a couple of glitches in Optim while doing this:

1. I had to reshape the N×N grid into a N²×1 vector.  Is Optim supposed to accept general arrays, or just vectors?

2. `optimize` threw an error when it tried to display the solution.  It looks like arrays of complex numbers might not be handled correctly.

antoine-levitt:
> Thanks for the offer. I looked at that and couldn't see how to start, so I'll try Optim first. Also, the vortex experts I work with are suspicious of Fourier pseudospectral derivatives, because of the seized gears problem at the periodic boundary.

I've got your problem specs from https://gist.github.com/thisrod/896d60bbad411f8feaed87d03cbbe038, so I'll try and get it working and report back. What do you mean by "seized gears"? Fourier methods implicitly periodize the function. It's no worse or better than FD-type methods that implicitly truncate it. I've never found it to be a problem either in theory or in practice.

> It's simple and it works. None of the current alternatives have both those properties.

Fair enough. CG isn't that complicated though, and a little sophistication goes a long way for these methods.

> However, Gauss-Seidel does better, finding a wave function with a single vortex, energy 6.98 and residual 1e-3 after 4000 steps.

I'm afraid that's life: all these methods try to find local minima, and it's very hard to say a priori which one they're going to find. That's especially true in these vortex problems. It might be that CG is a bit agressive at the beginning, and therefore "changes valleys" more quickly than other methods. Perhaps you could do a few steps of gradient descent then switch to CG? Maybe that's a stupid idea though... Have you tried LBFGS?

> I had to reshape the N×N grid into a N²×1 vector. Is Optim supposed to accept general arrays, or just vectors?

It's supposed to accept general arrays (and it does work, see eg tests/manifolds.jl). What exactly failed?

> optimize threw an error when it tried to display the solution. It looks like arrays of complex numbers might not be handled correctly.

Ah, I actually encountered this issue but was too lazy to report/debug it... Would you mind opening an issue/PR?

antoine-levitt:
So I tested with DFTK, and get solutions like this (abs2 shown):
Figure_1

Code here: https://gist.github.com/antoine-levitt/98a8b4a377ab1b8abd8eba7aa552a160 to run on this branch: https://github.com/JuliaMolSim/DFTK.jl/pull/93.
It looks like the normalization is off in your code: you're taking `norm(psi)=1`, shouldn't it be `h^2*norm(psi)=1`? You can easily change the normalization by defining a new sphere manifold with a different radius: see the definition of Sphere in Manifolds.jl. Also you use `-Delta` whereas I use `-Delta/2`, but that should be simple enough to convert.

thisrod:
> [Optim is] supposed to accept general arrays (and it does work, see eg tests/manifolds.jl). What exactly failed?

I tried to constrain the norm of a 2D array with `Manfold=sphere()`.  That might be asking for too much.  The error was

    MethodError: no method matching normalize!(::Array{Float64,2}) 

> Ah, I actually encountered this issue but was too lazy to report/debug it... Would you mind opening an issue/PR?

#775

antoine-levitt:
Ah, that's a good point. The offending call go normalize in Manifolds.jl has to be replaced by an explicit renormalizatuon.

antoine-levitt:
I'm not concerned about E, I'm concerned about the normalization `int psi^2 = 1`. You get a solution of `H[psi] psi = mu psi, norm(psi)=1`, but you're rather looking for a solution of `H[psi] psi = mu psi, h*norm(psi) = 1`, no? Those are not the same (the weighing of the interaction term differs)

antoine-levitt:
Also you solve slightly different problems in SOR vs CG: in SOR you fix the chemical potential free and let the normalization free, in CG you fix the normalization and let the chemical potential mu free. Not sure if that changes anything (both problems should be of equal hardness), but doesn't the SOR method cause problems? Eg psi = 0 is a solution.

thisrod:
> I'm not concerned about E, I'm concerned about the normalization `int psi^2 = 1`.

Oh sorry, the comment is wrong.  I actually adjusted the interaction coefficient to solve `L[psi] psi = mu psi, norm(psi)=1`.  I'll correct that and update the gist.

> doesn't the SOR method cause problems?

Good question.  This is why I called it kludgy but effective.  SOR is meant to solve a system of equations, and you would expect many things to go wrong when it tries solve an eigenproblem.  For SOR to work (and it does work: it converged to a solution) the eigenproblem has to be nonlinear in the right way.

> Eg psi = 0 is a solution.

It would be a solution if this was a linear eigenproblem.  But the nonlinearity makes two big differences.  First, every positive μ is an eigenvalue.  Second, μ determines the norm of the eigenvector.  (Physically, the chemical potential determines the particle number, and visa-versa.)  So, in this problem, psi = 0 is only a solution for μ = 0.

Those two points make this different enough from a linear eigenproblem, and similar enough to a linear system of equations, for SOR to work.

antoine-levitt:
Right. Psi=0 is still a solution for any mu, but an unstable one I guess. SOR could probably be adapted to eigenvalue problems by normalizing psi and updating mu, but using optim is better. 

Note that in SOR you divide by the diagonal, which is good because it acts as a preconditionner for the potential energy. You can emulate that in optim with a preconditionner. In DFTK we do something similar but in reciprocal space, which preconditions for the kinetic energy instead. 

Keep me posted when you have something with the right normalization and I'll test with DFTK.

thisrod:
> Keep me posted when you have something with the right normalization and I'll test with DFTK.

I've updated [the gist](https://gist.github.com/thisrod/896d60bbad411f8feaed87d03cbbe038).  As a bonus, I rounded off the constants, and now SOR and Optim agree to 6 figures.

> Right. Psi=0 is still a solution for any mu, but an unstable one I guess.

You're right, of course.

> SOR could probably be adapted to eigenvalue problems by normalizing psi and updating mu

That's how it is usually done.

> Note that in SOR you divide by the diagonal, which is good because it acts as a preconditionner for the potential energy.

Interesting.  I should finish reading Trefethen & Bau and learn the theory behind preconditioners.

Every physicist who tries this comes up with the same idea: how about we use a basis of Fock wave functions?  Those are eigenstates of the total energy for a harmonic oscillator, so they constrain the kinetic, potential, and every other kind of energy.  Unfortunately, no one knows how to do a fast Fock transform, and that elegant idea keeps losing to brute force.  However, a small basis of Fock states might be a good preconditioner.

antoine-levitt:
> I've updated the gist

Thanks, will take a look.

> Every physicist who tries this comes up with the same idea: how about we use a basis of Fock wave functions? 

Exactly. One thing I've wondered is if it's problematic to not include the magnetic field in this basis, but I think it should be fine. As you say, there is no fast transform for this. However we don't need the full transform, we just need to apply the inverse H^-1 where H = -Delta/2 + x^2/2. H^-1(x,y) is the Green's function, which can be computed explicitly (through special functions). Then there is the question of how to apply H^-1 efficiently to a vector. I think that can be done in O(N), eg with H-matrices, exploiting the regularity of the Green's function. I thought about exploring that but I don't care enough about gross-pitaevskii to do it; if you're interested I think this can be very efficient though.

antoine-levitt:
https://gist.github.com/antoine-levitt/72916a06bbbe342a5648240c19d577ee produces 
GPE in ~1s, with ~260 iterations to get to a residual of 1e-8 (the code is pretty unoptimized for this use case, but the orders of magnitude should be right). I did the proper normalization `\int psi^2 = 1`, and compensated by multiplying `g` by `h^2`.

antoine-levitt:
The iterations sometimes converge to a one-vortex solution, which is higher in energy (they are seeded from a completely random initial guess)

thisrod:
The 2D and 3D harmonic oscillators are spherically symmetric.  If you choose the right set of degenerate energy eigenstates, they are also angular momentum eigenstates, and magnetic fields are handled automatically.

I'll read up on preconditioning.  Thanks for the help.

thisrod:
One second!  That's pretty good.  I'll try the molecular dynamics code tomorrow.

antoine-levitt:
Ah yes, of course, that makes sense. Preconditioning for this kind of problems is not very well explored in the literature, but the basic idea is the same as for other linear / optimization problems: you want to rebalance the relationship between error and residual.

thisrod:
I think I understand.  When a solver is asked to minimize E(ψ), it wastes a lot of time fiddling with tiny components of high-energy Fock states.  The solution is to minimize the function φ ↦ E(H⁻¹φ), where those Fock states have unit energy.
  
Presumably that fails when the nonlinear term is small, E∘H⁻¹ is almost constant, and there is nothing left to minimize?

I noticed that SOR takes more steps when it's working with a finer grid.  The lack of preconditioning on momentum might explain that.

Thanks for your help.

thisrod:
The phase plot of the DFTK wave function is interesting.  It shows the superfluid rushing through the gaps between the vortices and the edge of the condensate.

phase plot

antoine-levitt:
> I think I understand. When a solver is asked to minimize E(ψ), it wastes a lot of time fiddling with tiny components of high-energy Fock states.

That's the essence of ill-conditioning. We describe that for the particular case of gross-pitaevskii in our JCP paper. There are many ways to look at preconditioning, but the simplest is: you want to solve F(x) = 0, and you do that by something like xn+1 = xn + alpha F(xn), where alpha is a fixed constant selected appropriately (more advanced methods like CG are more fancy, but still have the same issues). Let's linearize things around a solution x*, F(x*) = 0, which I take wlog to be 0. Then the iteration is xn+1 = xn + alpha F'(0) xn = (1 - alpha F'(0)) xn. If F'(0) is a multiple of the identity, then you get the solution in one shot with an appropriate alpha. If however F'(0) has very high and very low eigenvalues, you're screwed: you're forced to take alpha small to ensure stability, but that leads to very slow convergence (as you say, you fiddle with high-energy components). The problem is that the error-residual relationship r = F'(x*) e, where r = F(x) and e = x - x*, is distorted, so that r is not a faithful representation of e. The solution to that is to take a preconditioned step xn+1 = xn + alpha P F(xn), where P is a preconditioner. The ideal preconditioner is the Newton step P = -F'(xn)^-1.

For GP the situation is a bit complicated because of the constraint, but the general idea remains the same. The problem is still there even with `g=0`. The problem is still nonlinear because of the constraint. Ill-conditioning comes from 1) the large number of local minima 2) the high-momentum states, for which the kinetic energy diverges 3) the extended states, for which the potential energy x^2 diverges. 1) is pretty hard to do anything about 2) is solved by preconditionning for the laplacian as we do in DFTK 3) is solved by preconditionning for the potential energy as you do implicitly in SOR. If you do KE preconditioning, your convergence rate is independent of the mesh size, but depends on the size of the system; if you do PE preconditioning it's the reverse. We have a finer analysis in our paper. Solving 2) and 3) at the same time is an open research problem: this is where ideas such as fast transforms, hierarchical matrices etc. could be explored.

> As a proof of concept, would it make sense to numerically diagonalize the 1D harmonic oscillator with the relevant boundary condition, then rescale the high-energy components by projection and brute force?

Yep. Managing to do that in O(N), where N is the number of DOF, independently of the system or mesh size, is already hard. With a nonzero `g` the solution is not analytic. The two immediate approaches are to discretize by FD/pseudospectral (in this case the problem is finding a good preconditioner), or to discretize using the eigenfunction basis (in this case the problem is evaluating the nonlinear term efficiently). I did the former with Hmats a few years ago and it seemed to work, although with a pretty large prefactor.

thisrod:
Sorry for taking a while to notice your reply.  We had a long weekend in Australia.

I was wondering how scaled axes would interact with the normalization, but that makes it clear.  The preconditioner acts on the step, not the axes.

In the problems I'm working with, I suspect that potential energy is a bigger problem than kinetic, so I'll try using V as a preconditioner.

Do you know of a hierarchical matrix library for Julia?

thisrod:
Sorry for taking a while to notice your reply.  We had a long weekend in Australia.

I was wondering how scaled axes would interact with the normalization, but that makes it clear.  The preconditioner acts on the step, not the axes.

In the problems I'm working with, I suspect that potential energy is a bigger problem than kinetic, so I'll try using V⁻¹ as a preconditioner.  I have two technical questions about Optim.  With a complex ψ, does V "approximate the Hessian" of V|ψ|²?  Secondly, is there a way to supply a preconditioner for a 2D array, or is it simplest to splat everything to vectors?

Do you know of a hierarchical matrix library for Julia?

antoine-levitt:
> With a complex ψ, does V "approximate the Hessian" of V|ψ|²? 

In this particular case yes, because the gradient turns out to be an analytic function of psi. Do keep in mind that what's to be approximated is not just the Hessian of the energy, but rather the Lagrangian (plus some trickery).

> Secondly, is there a way to supply a preconditioner for a 2D array

Optim only interfaces with the preconditionner through `ldiv!`, `dot` and `precondprep!`. You can overload these for your own type and act on whatever you want. See `DFTK/src/scf/direct_minimization.jl` for an example.

> Do you know of a hierarchical matrix library for Julia?

There are a couple:
https://discourse.julialang.org/t/kernelmatrices-jl-a-software-package-for-working-with-hierarchical-matrices/21696
https://github.com/JuliaMatrices/HierarchicalMatrices.jl
and you can `ccall` into any C library.

thisrod:
I tried `ConjugateGradient` by itself, and with a potential energy preconditioner.  The problem was a faster rotating order parameter, which should have a vortex lattice.

The preconditioner seemed to speed up convergence, but in both cases the solver gives up before it is satisfied with the solution.  (There was a message about increasing error and line search failure.)  The solution looks right near the origin, but becomes random noise outside a certain radius.

resp200130a
resp200130b
resp200130c
resp200130d

SOR converges to a different solution, and is much less efficient.  On the other hand, the solution is smooth and physically plausible everywhere, so it works better as an initial condition for dynamics.

resp200130e
resp200130f

I guess the issue is the tiny density at that large radius, which allows the velocity/phase gradient to be large without increasing the total energy by much.  SOR acts locally, so it smooths out those phase gradients to reduce the local energy, even though that doesn't have a significant effect on the total energy.

antoine-levitt:
> The preconditioner seemed to speed up convergence, but in both cases the solver gives up before it is satisfied with the solution. (There was a message about increasing error and line search failure.)

Ah, this is annoying, you need to pass `allow_f_increases=true`. @pkofod didn't we talk about making that the default at some point?

> I guess the issue is the tiny density at that large radius, which allows the velocity/phase gradient to be large without increasing the total energy by much.

It's not obvious to me, I would have thought CG preconditioned and SOR have the same behavior here. Definitely for the unpreconditioned CG you expect bad things to happen at the edge. It seems that the energy for CG is lower though?

thisrod:
> I would have thought CG preconditioned and SOR have the same behavior here. Definitely for the unpreconditioned CG you expect bad things to happen at the edge.

Good point.  I'll check that I got the preconditioning right.

> It seems that the energy for CG is lower though?

Yes.  In the SOR solution, I kept μ fixed, found an energy minimizer with a slightly different number of atoms to the CG solution, then normalised it.  SOR might converge to the CG solution if I was more careful about adjusting μ to solve the same problem.

I have an idea about preconditioning.  The operators ∂/∂x and ∂/∂y can be discretized by one-sided finite difference formulae, on the side of larger x and y.  Then, when ψ is flattened to a vector, the operator H = ∇²+V+ΩL will discretize to an upper-triangular matrix (or maybe lower-triangular), and H⁻¹ can be evaluated by back substitution.  Has that been tried before?

antoine-levitt:
> I have an idea about preconditioning. The operators ∂/∂x and ∂/∂y can be discretized by one-sided finite difference formulae, on the side of larger x and y. Then, when ψ is flattened to a vector, the operator H = ∇²+V+ΩL will discretize to an upper-triangular matrix (or maybe lower-triangular), and H⁻¹ can be evaluated by back substitution. Has that been tried before?

I'm like 95% sure that won't work at all, even in 1D. Decentering the FD changes the operator dramatically. But I'd be interested in seeing how that fails (then that becomes a very nice problem to show students)

thisrod:
> Ah, this is annoying, you need to pass `allow_f_increases=true`.

When I do that, the solver reports success.  However, the minimizer has random phase and constant amplitude in the corners of the square domain.  That seems wierd.

resp200204zs
resp200204zt
resp200204zu

[Source code](https://gist.github.com/thisrod/a338c913eca1963bf457766bb0de4146)

antoine-levitt:
Looks about right to me? Note you don't precondition here

thisrod:
> Looks about right to me?

Yes.  I was wondering why the solver stopped there, instead of smoothing the noise in the corners.

The result has `x_converged == false`, `f_converged == true` and `g_converged == false`, with `f_abstol == 0.0` and `f_abschange == 0.0`.  I interpret that to mean that, along the line defined by the solution and gradient, there is a minimum within numerical epsilon of the solution, so the solver needs a preconditioner to tell it where to look next.  Is that right?

thisrod:
The potential energy preconditioner makes things worse!  On a larger domain, the default `ConjugateGradient` succeeds after 4260 steps, but the preconditioned one gives up at 10000 steps with a worse solution.

Now it gets wierd.  Preconditioning with 1/V succeeds after 4675 steps, with a slightly worse solution than the default.

With the sphere constraint, I guess that the preconditioner and the cost should project onto the tangent plane in compatible ways, and everything should just work?

resp200205a
resp200205b
resp200205c

thisrod:
I've added some of this discussion to the manual.  Is it possible to reopen the PR?

antoine-levitt:
This one has gone on for too long, can you open a new PR?

antoine-levitt:
> The result has x_converged == false, f_converged == true and g_converged == false, with f_abstol == 0.0 and f_abschange == 0.0. I interpret that to mean that, along the line defined by the solution and gradient, the energy of the solution is within numerical epsilon of the minimum, so the solver needs a preconditioner to tell it where to look next. Is that right?

Very likely. You can disable the stopping criterion on the energy. That'll possibly mess up the stepsize routines, I'm not sure

antoine-levitt:
Weird that preconditionning doesn't work... it definitely does in a more complicated case I've tried. Possibly sqrt(1+V^2) is too steep and you should play around with parameters, eg sqrt(alpha + V^2)? Or maybe V is not an accurate representation of the self-consistent potential (you can improve that using precondprep! to update the preconditioner)? It should be, though... I think we discuss the parameters in our paper.

pkofod:
Just commenting here to say: pretty pictures, but I have absolutely no idea what you're discussing :D but do keep it going :)

thisrod:
> Possibly sqrt(1+V^2) is too steep ... Or maybe V is not an accurate representation of the self-consistent potential

Good idea.  I replaced that with `sqrt(rc^4 + V^2)`, where `rc = 5` is roughly the condensate radius.  This should be a good approximation: it's a flat chemical potential in the condensate (except in the vortex cores), and V in the vacuum outside.  The result was convergence in 2540 iterations to a residual of 2e-8.  So that preconditioner roughly doubles the rate of convergence, but finds the same solution.

Something bothers me about these results.  Our hypothesis was that, without preconditioning, the solver would spend too much time adjusting the parts of the solution that have a lot of energy but few particles.  The solutions suggest it's the opposite.  They're OK near the trap centre, but wrong by 40 orders of magnitude where the potential energy is largest.

I'm reading your 2017 JCP more closely now that I have a better idea what a preconditioner is.  I'm a bit puzzled by the definition in Section 2 that ∇E = 2Hψ, then the steepest descent step in Equation 5.1 being ψ → ψ - α(H-λ)ψ.  I'm using Optim with a gradient of Hψ.  Will that factor of 2 make a difference to the conjugate gradient solver?  (I don't understand how conjugate gradient works.)

thisrod:
> do keep it going

Thanks!  I was a bit worried that I'm wasting your and Antoine's time.

thisrod:
Now this is interesting: gradient descent preconditioned with √(rc⁴+V²) gets the edges right, it just takes a long time.  These results are for 50_000 steps, and have a residual of 1e-5.

There are some stray vortices near the edge of the domain, whose cores mess up the straight line on the density graph.

resp200106a
resp200106b

antoine-levitt:
For conjugate gradient, look it up : the descent direction is a combination of the gradient and past search directions (to make them conjugate and avoid zigzags). Re factor of 2, do check that it does not matter, I'm not sure (the linesearch probably does not like it at all, and CG is pretty reliant on good linesearches). Also try LBFGS. The gradient is 2Hpsi, but you want to project on the tangent space, so it's (1- x x*) 2 H psi, hence the lambda. Optim does the projection by itself so you just need to supply the gradient 2 H psi

antoine-levitt:
Also it's possible that the edges get lost because of linesearches. There are tricks you can play but they are a bit tricky to implement. But maybe you can try playing with the linesearches (eg try static and backtracking)

thisrod:
I've fixed the factor of 2, but it didn't make a difference to the solution.

I understand how CG works to minimize a quadratic form by excluding directions it has already searched.  I still don't quite understand how it works in the general case, where the Hessian changes with each step, and the algorithm must forget about the directions it searched 1000 steps ago, that are now optimized for the wrong quadratic form.

For the 7 vortex problem, CG goes wrong at the grid corners very quickly.  I'm going to experiment with different boundary conditions.  But it's interesting that steepest descent avoids this.

Red dots are the initial guess, for comparison.

resp200210a
resp200210b
resp200210e
resp200210c
resp200210f
resp200210d
resp200210g

antoine-levitt:
For cg, that's why there are usually periodic resets imposed (or at least countermeasures). Also optimization doesn't usually require thousands of steps, gross pitaevskii is unusually bad because of a nasty combination of hard and soft modes. Have you tried LBFGS? Otherwise possibly an option is Newton, especially for 2D finite differences discretization. Resolving the edges is most likely a very badly conditioned problem though.

thisrod:
Thanks for your help; I've learnt a lot.  I'm going overseas next week, but I'll get back to this when I'm back.

> Have you tried LBFGS? Otherwise possibly an option is Newton, especially for 2D finite differences discretization.

I think I tried LBFGS, but I should try again.  I haven't tried Newton.

> Resolving the edges is most likely a very badly conditioned problem though.

I'm not sure about that.  The Hessian/Hamiltonian eigenstates with support at the edge are the ones with large energy, so it seems like the condensate should be the badly conditioned part.

PR #775, opened 2020-01-16
Breaking tests with Complex input

thisrod:
This adds two tests, that currently throw errors for the following reasons:

1. (I think) The `show` or `repr` method for solutions tries to `@printf` complex numbers.

2. The default solver attempts to sort complex numbers.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -239,7 +239,8 @@ function Base.show(io::IO, r::MultivariateOptimizationResults)
 
     @printf io " * Candidate solution\n"
     nx = length(minimizer(r))
-    str_x_elements = [@sprintf "%.2e" _x for _x in take(minimizer(r), min(nx, 3))]
+    str_x_elements = [repr(_x, context=:compact => true)
+        for _x in take(minimizer(r), min(nx, 3))]
     if nx >= 4
         push!(str_x_elements, " ...")
     end
@@ -252,7 +253,8 @@ function Base.show(io::IO, r::MultivariateOptimizationResults)
     @printf io " * Found with\n"
     @printf io "    Algorithm:     %s\n" summary(r)
     nx = length(initial_state(r))
-    str_x_elements = [@sprintf "%.2e" _x for _x in take(initial_state(r), min(nx, 3))]
+    str_x_elements = [repr(_x, context=:compact => true)
+        for _x in take(initial_state(r), min(nx, 3))]
     if nx >= 4
         push!(str_x_elements, " ...")
     end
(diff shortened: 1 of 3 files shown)


pkofod:
Interesting :) Good catach... I will think about a better way of printing complexs (or just not print these)

antoine-levitt:
Can this be merged?

thisrod:
I've fixed the first one, I think.

thisrod:
I'm looking at point 2.  The error occurs at `Optim/src/multivariate/solvers/zeroth_order/nelder_mead.jl:156`, where the method `initial_state(method::NelderMead, options, d, initial_x)` assumes that the type of the minimum is the same as the eltype of the minimizer.

I presume that the code is supposed to handle a cost function that returns any subtype of `Real`, though I don't fully understand the design intent here.  Is there a reliable way to determine the type of the minimum?  I'd guess that a reliable and extensible implementation of Optim will need one.  Should be a method to return that type for a given problem?

thisrod:
At a higher level, I think there's a choice to be made between the following designs:

1. Every gradient-free solver is implemented to handle complex minimizers, or

2. Gradient-free solvers are only implemented for real minimizers, and a generic constructor method for complex minimizers splits them into their real and imaginary parts.

pkofod:
Yeah, I think maybe gradient free solvers will have to use the splitting and or special cases for the complex case.The problem with nelder mead (if I fixed the type stuff you found above) for example is that the simplex will be of order two times too low, and it will not be able to disentangle the real and imaginary part the way it's written right now.

thisrod:
> I think maybe gradient free solvers will have to use the splitting and or special cases for the complex case.

Then the next step might be to merge the display fix, and raise the gradient free solvers as an issue that will take longer to fix.

What's the best way to rebase this PR onto the master branch?  Should I just do it and force push to my github?

thisrod:
I have a plan for zeroth order complex.  Please tell me if there is an obvious problem or better way.

1. In `NLSolversBase`, implement `ComplexNonDifferentiable`.  This wraps the `NonDifferentiable` methods with `Complex`, to combine real and imaginary parts.

2. In `Optim`, implement this pseudo code:

    function optimize(
        d::AbstractObjective,
        initial_x::AbstractArray{<:Complex},
        method::ZerothOrderOptimizer,
        options
    )
        c = ComplexObjective(d)
        # split initial_x into real and imaginary parts
        results = optimize(c, initial_parts, method, options)
        # combine real and imaginary results into complex
        results
    end

PR #776, opened 2020-01-16
Normalized multidimensional arrays

thisrod:
The option `Manifold=sphere()` is supposed to handle multidimensional arrays.  This adds a failing test as a reminder that it doesn't yet.

### DIFF
--- a/docs/src/algo/manifolds.md
+++ b/docs/src/algo/manifolds.md
@@ -20,7 +20,7 @@ All first-order optimization methods are supported.
 
 The following manifolds are currently supported:
 * Flat: Euclidean space, default. Standard unconstrained optimization.
-* Sphere: spherical constraint `||x|| = 1`
+* Sphere: spherical constraint `||x|| = 1`, where `x` is a real or complex array of any dimension.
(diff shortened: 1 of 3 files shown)


antoine-levitt:
Just fix it : https://github.com/JuliaNLSolvers/Optim.jl/blob/92e3910ee41e51580a4dd101a99ae24696ab6b2c/src/Manifolds.jl#L69 change to `x ./= norm(x)`

thisrod:
> Just fix it :

Done, I think.

antoine-levitt:
Well actually: https://github.com/JuliaLang/julia/pull/34239 so this would have been fixed by itself.

thisrod:
Is it worth adding a Julia version test, and using `normalise!` if it supports multidimensional arrays?

antoine-levitt:
No I think the current version is OK. The tests seem to fail though?

thisrod:
Is there a way to run the tests again?  I suspect that I confused Travis by force pushing my branch while it was running.

antoine-levitt:
I think close and reopen?

pkofod:
Seems to be working :)

pkofod:
Thanks!

pkofod:
note to self: it's a good idea to put a seed in such a test :)

PR #781, opened 2020-01-29
Handle Inf values in NewtonTrustRegion

andreasnoack:
The will lead to NaNs in the gradient and Hessian which have to be handled explicitly.

### DIFF
--- a/test/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/test/multivariate/solvers/second_order/newton_trust_region.jl
@@ -196,4 +196,9 @@ end
     Optim.solve_tr_subproblem!([0, 1.], [-1000 0; 0. -999], 1e-2, ones(2))
 end
 
+@testset "Handle Inf without erroring" begin
+    o = optimize(t -> Inf, ones(10), NewtonTrustRegion())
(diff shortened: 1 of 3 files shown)


test/multivariate/solvers/second_order/newton_trust_region.jl
```diff
@@ -196,4 +196,9 @@ end
     Optim.solve_tr_subproblem!([0, 1.], [-1000 0; 0. -999], 1e-2, ones(2))
 end
 
+@testset "Handle Inf without erroring" begin
```
pkofod: Is this test valid, or will it hit the problem we saw earlier with literals?

pkofod:
Seems like a good idea. I'm wondering about that test failure :| Can you verify that the test doesn't produce 0 gradients and therefore doesn't fail?

andreasnoack:
The test works as intended locally

test/multivariate/solvers/second_order/newton_trust_region.jl
andreasnoack: I'm using FD here so it produces NaNs as it should.

PR #785, opened 2020-02-12
Change initial_state in ipnewton.jl to allow broader class of AbstractArrays

jonniedie:
Allows custom array types that have the ```similar``` method implemented to be passed through the ```IPNewton``` solver. For example [labeled arrays](https://github.com/JuliaDiffEq/LabelledArrays.jl).

### DIFF
--- a/src/multivariate/solvers/constrained/ipnewton/ipnewton.jl
+++ b/src/multivariate/solvers/constrained/ipnewton/ipnewton.jl
@@ -100,7 +100,7 @@ function Base.convert(::Type{IPNewtonState{T,Tx}}, state::IPNewtonState{S, Sx}) 
                   )
 end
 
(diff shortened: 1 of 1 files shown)


pkofod:
I hadn't seen this. Thanks. Did you actually try it out? The changes are sufficiently simple that it isn't super important, but if you have a simple test case that would be great.

jonniedie:
Here is an example using LabelledArrays with the documentation example for nonlinear constrained optimization.

```
using LabelledArrays
using Optim

fun(u) = (1.0 - u.x)^2 + 100.0 * (u.y - u.x^2)^2

u₀ = LVector(x= 0.0, y= 0.0)
lb = LVector(x=-0.5, y=-0.5)
ub = LVector(x= 0.5, y= 0.5)

df = TwiceDifferentiable(fun, u₀; autodiff=:forward)
dfc = TwiceDifferentiableConstraints(lb, ub)

val = optimize(df, dfc, u₀, IPNewton())

```

pkofod:
Sorry, I was very busy and forgot. Thanks!

PR #786, opened 2020-02-13
Use Compat for `dot(x, P, y)`

devmotion:
This PR replaces the custom fixes for `dot(x, P, Y)` on older Julia versions with the fixes provided by Compat to avoid warnings such as
```julia
WARNING: Method definition dot(Any, Any, Any) in module Compat at /home/travis/.julia/packages/Compat/if8Gv/src/Compat.jl:111 overwritten in module Optim at /home/travis/.julia/packages/Optim/SFpsz/src/multivariate/precon.jl:24
```
with Compat >= 3.2 (see, e.g., [Travis CI tests for MCMCChains](https://travis-ci.org/TuringLang/MCMCChains.jl/jobs/649969906#L438)).

When testing locally Pkg could not resolve the dependencies correctly due to some bound on Compat introduced by Suppressor, hence I removed Suppressor (Suppressor still only uses a REQUIRE file, maybe that's an issue as well?). If the output of the tests seems to be too verbose now, one could maybe adjust the keyword arguments to not print all traces?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "429524aa-4258-5aef-a3af-852621145aeb"
 version = "0.20.1"
 
 [deps]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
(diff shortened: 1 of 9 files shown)


pkofod:
This should be fixed.

PR #788, opened 2020-02-24
Fix swapped relative and absolute tolerances

gwater:
Report tolerances of the objective function correctly in results by following their order given in 
https://github.com/JuliaNLSolvers/Optim.jl/blob/74794f21b131825c0cd16882276b103dcf94e723/src/types.jl#L179-L180


pkofod:
Oops, thanks!

Issue #789, opened 2020-02-25
Unused return value from update_state!() in optimize()


gwater:
The variable `ls_failed` is assigned but never used. At the same time `ls_success` is never changed during optimization and is always set to `true` no matter what happens.
https://github.com/JuliaNLSolvers/Optim.jl/blob/74794f21b131825c0cd16882276b103dcf94e723/src/multivariate/optimize/optimize.jl#L57

pkofod:
Thanks, I merged the PR

gwater:
Thanks!

PR #790, opened 2020-02-25
Actually stop at line search failure

gwater:
addresses #789

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -54,7 +54,7 @@ function optimize(d::D, initial_x::Tx, method::M,
     while !converged && !stopped && iteration < options.iterations
         iteration += 1
 
-        ls_failed = update_state!(d, state, method)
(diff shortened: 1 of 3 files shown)


gwater:
I dont understand why Windows x64 breaks and everything else works.

pkofod:
Thanks!

Yeah, some times there can be subtle differences between Linux and Windows. I will try to see if I can figure out what is going on.

gwater:
I think we can deduce nevertheless that the differencent behavior between Windows x64 and other operating systems are not due to this change. Therefore I think it should be adressed in a separate issue / PR

gwater:
The failing test is FMinBox using Nonlinear GMRES preconditioned with Gradient Descent

pkofod:
It's fine to remove that specific run then.

gwater:
ok, I'll make a patch to mark it as broken on Windows x64

gwater:
done

pkofod:
Thanks a lot, this was a very stupid oversight on my part.

PR #791, opened 2020-02-27
Use Compat.jl for dot.

pkofod:


timholy:
What's the motivation for `show_trace=true` in particle swarm and simulated annealing? I find the length of the test output makes it hard to assess the overall results, it would be nice to shorten it somehow.

PR #792, opened 2020-03-03
fix printf syntax in error message

tpapp:
This is a trivial fix about error printing (currently prints `%d` instead of the coordinate index).


pkofod:
Good call, thanks

PR #795, opened 2020-03-10
Update the documentation of the callback function. 

hcarlsso:
I thought some more details on the callback function would be helpful.


pkofod:
Thanks. The callback interface has been a pain point for some time, but I don't use it much myself, so I havn't gotten around to it.  IT would probably be better if it was just passed a namedtuple or dictionary or similar.

hcarlsso:
I am using it for [ProgressMeter.jl](https://github.com/timholy/ProgressMeter.jl)

pkofod:
Thanks!

PR #799, opened 2020-03-15
fix Brent for flat functions

bjarthur:
fixes https://github.com/JuliaNLSolvers/Optim.jl/issues/798

still need to add tests

### DIFF
--- a/src/univariate/solvers/brent.jl
+++ b/src/univariate/solvers/brent.jl
@@ -143,7 +143,7 @@ function optimize(
         new_f = f(new_x)
         f_calls += 1
 
-        if new_f <= new_minimum
+        if new_f < new_minimum
             if new_x < new_minimizer
                 x_upper = new_minimizer
                 best_bound = "upper"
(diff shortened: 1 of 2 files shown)


bjarthur:
this is ready too i think.  comments?

pkofod:
Wow, I did not see this until just now. I will review. Thanks

pkofod:
You said in OP that tests were needed, but I only see one commit. Was this the force push? It looks reasonable. I can see that with the old branch, you'd be successively be discarding the lower bound, end. Thanks for this.

bjarthur:
push force, yes.  thanks.

and shoot!  we still have problems.  as you merged i was thinking that we really need to test the reversed case, where the function is `-sign(x)` in addition to `sign(x)`.  and sure enough, my "fix" to the latter results in an error to the former:

```
julia> result = optimize(x->-sign(x), -2, 1)
 * Algorithm: Brent's Method
 * Search Interval: [-2.000000, 1.000000]
 * Minimizer: -8.541020e-01
 * Minimum: 1.000000e+00
 * Iterations: 38
 * Convergence: max(|x - x_upper|, |x - x_lower|) <= 2*(1.5e-08*|x|+2.2e-16): true
 * Objective Function Calls: 39
```

so the tests really need to be:

```
for f in [x->sign(x), x->-sign(x)]
     for x in [[-2,2],[-1,2],[-2,1]]
         result = optimize(f, x...)
         @test Optim.converged(result)
         @test Optim.minimum(result) == -1.0
     end
end
```

do you have any insight as to the proper way to fix this?

pkofod:
Oh.. Well. Let me have a look :)

PR #808, opened 2020-04-26
Support creation of `TwiceDifferentiable` from `IPNewton` methods

timholy:
This just adds a couple of convenience methods. CC @ChantalJuntao.

I also found I needed to add `sitename` to get the docs to build. Not sure if you want that or not, but I thought I'd offer. Happy to strip that commit from the PR if necessary.


pkofod:
Oh, excellent, thanks. Seems like we have a problem with preconditioning on v1.4, I have to look into that.

pkofod:
Now that I have you here :) If I want to e-mail you, is it still the wustl e-mail?

PR #812, opened 2020-04-30
Add simple InitialStatic alpha choice by providing just a number to t…

### LINKED ISSUE #805: numerical values for alphaguess
I think it would be very convenient for the `alphaguess` keyword, defined for many algorithms, to accept numerical values. Then, `alphaguess` would be transformed to an `InitialStatic` object :

```julia 
if isnumeric(alphaguess):
   alphaguess = LineSearches.InitialStatic(alpha=alphaguess)
end
```
This is in order to make smoother the most common tweak one would perform to tune the optimization, i.e. changing the step size. Now, one has to write
```julia
optimize(fg!, x0, Optim.LBFGS(alphaguess=LineSearches.InitialStatic(alpha=η)))
```
while I'd like to have
```julia
optimize(fg!, x0, Optim.LBFGS(alphaguess=η))
```
As a side note, I found LineSearches documentation a bit lacking and had to go look into the source to understand how to construct an InitialStatic object.

pkofod:
…he alphaguess keyword in method constructors.

Fixes #805

### DIFF
--- a/src/multivariate/solvers/first_order/accelerated_gradient_descent.jl
+++ b/src/multivariate/solvers/first_order/accelerated_gradient_descent.jl
@@ -18,7 +18,7 @@ function AcceleratedGradientDescent(;
                                     alphaguess = LineSearches.InitialPrevious(), # TODO: investigate good defaults
                                     linesearch = LineSearches.HagerZhang(),        # TODO: investigate good defaults
                                     manifold::Manifold=Flat())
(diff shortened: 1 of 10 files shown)


CarloLucibello:
thanks for fixing this. Maybe this should be documented somewhere

pkofod:
Certainly. Where would be a good place to put it? In the method docstrings? Maybe an early mention in the Optim docs as well.

PR #815, opened 2020-05-08
Fix minor bugs

amanbh:
Fixes the following two (unrelated) bugs:
1. simplex_trace and centroid_trace functions are missing call to trace method
2. pandoc has replaced --latex-engine flag with --pdf-engine


pkofod:
Wow, that's a major oooops. Thanks a lot!

PR #843, opened 2020-08-14
WIP: Parallel SAMIN

MatFi:
My problems often involve functions that are expensive to evaluate. However, Julia IMHO offers  hardly any choice of native packages with parallel working methods for (global) optimization (besides [BlackBoxOptim.jl](https://github.com/robertfeldt/BlackBoxOptim.jl)). Now I made your SAMIN method cluster-compatible and would like contribute this to Optim.jl.  What do you think ?

To do:
- [X]  Worker based parallelization
- [ ] Threaded parallelization (needed?)
- [ ] Correct implementation of `f_calls`-stats (now it just uses  the iteration counter)
- [ ] …

From my first benchmarks I can see that the pay-off comes when the f_call costs more than 1e-5 s (all workers on the same machine)

image

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,6 +31,7 @@ OptimTestProblems = "cec144fc-5a64-5bc6-99fb-dde8f63e154c"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 
 [targets]
-test = ["Test", "OptimTestProblems", "Random", "RecursiveArrayTools"]
+test = ["Test", "OptimTestProblems", "Random", "RecursiveArrayTools", "Distributed"]
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "0.22.0"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LineSearches = "d3d80556-e9d4-5f37-9878-2ab0fcc64255"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
This is great! However, you doing too much work. Instead of trying to implement every form of parallelism, which you won't do (what about GPUs? TPUs? ...), it would be good if this was just a batch interface, i.e. give the user an array of arrays or a matrix of `x` and have them return a whole vector of objective functions. If you do it like that, all forms of parallelism are implemented.

pkofod:
I have made something like what Chris mentions for Particle Swarm that you can have a look at https://gist.github.com/pkofod/c6f0dc28588e1d65b521ba785405aff2

pkofod:
I also have the same algorithm in a slightly different flavor here: https://github.com/pkofod/NLSolvers.jl/blob/bc8b1941218fea9349dc9797ed32bb577c59289c/src/optimize/randomsearch/particleswarm.jl#L48 the idea there is that the objective type determines how the objective receives the states, either as a batch or one at a time through the `batched_value` call.

MatFi:
> it would be good if this was just a batch interface

I definitely see the advantages of doing this, but keep in mind that some functions can vary a lot in timing (e.g. integration of stiff differential equations). Therefore I would like to keep the possibility of asynchronous evaluation, because a batch based implementation is too much idle in such cases (Of course, it is always possible to synchronize the asynchronous procedure with a full batch to allow for both).

> If you do it like that, all forms of parallelism are implemented.
Actually the implementation is essentially shifted to the user then, but of course one can cover the most common cases using a suitable dispatch.

> I have made something like what Chris mentions for Particle Swarm that you can have a look at 

Thanks, I was already about to ask for examples... Extremely helpful :+1:

ChrisRackauckas:
>I definitely see the advantages of doing this, but keep in mind that some functions can vary a lot in timing (e.g. integration of stiff differential equations). Therefore I would like to keep the possibility of asynchronous evaluation, because a batch based implementation is too much idle in such cases (Of course, it is always possible to synchronize the asynchronous procedure with a full batch to allow for both).

That's exactly the reason why it should be on the caller's side. Sometimes that's the case, and integration of stiff stochastic differential equations I've seen whopping 3 orders of magnitude timing differences with the same equation (due to switching off a steady state). However, you can't rely on this because the spawn cost for threads is high, if you have an ODE that finishes in like 1ms the spawning a task per thread for dynamic scheduling or using `pmap` is far too slow. So you can't just have "threading", you need 

1. threading with dynamic scheduling
2. threading with static scheduling
3. threading with partial static scheduling (clumping but not the number of threads)
3. distributed dynamic
4. distributed static
5. distributed + threads in all combinations
6. GPU via CuArray
7. GPU via DiffEqGPU
8. GPU via KernelAbstractions
9. MPI (since Distributed doesn't scale all that well)
10. MPI+CUDA (the Clima setup)

Those are 10 forms of parallelism we are actively using in projects right now with stiff differential equations, all for different purposes due to the trade-offs. It's just so much easier to tell users who want asynchronous to just loop over `@spawn` than to try and handle parallelism efficiently, since that means something different to every user.

lrnv:
Hi, 

Would it be difficult ot integrate @pkofod gist into Optim.jl ? I dont think i have enough comprehension of Julia's internals to do it myself, as the code @pkofod produced is quite different from the content of the ParticleSwarm.jl file in this repo...

But if you think it is doable easily and point me to the right direction, i might do it :)

PR #849, opened 2020-08-25
Fix project_tangent! and retract! args in manual

thisrod:


antoine-levitt:
Looks good, thanks!

thisrod:
I've been playing with the manifold of wave functions whose vortices are pinned at fixed positions, but whose density and phase can otherwise relax.  When it's a bit more stable, I'll write it up as an example.

antoine-levitt:
Ah nice I'd be interested in knowing more

pkofod:
Oh, good catch. Thanks! And yes, we're always happy to hear about usecases and user-written manifolds!

antoine-levitt:
I've been meaning to interface with https://github.com/JuliaManifolds/Manifolds.jl/ for a while, but never got around to it

Issue #834, opened 2020-07-13
Fminbox optimization does not exit when callback returns true


clintonTE:
MWE (borrowing some snippets from the docs):
```Julia
using Revise, Optim
mutable struct AdditionalState
  fevalctr
end

function f(x, Ψ)
  Ψ.fevalctr += 1

  return (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
end

function g!(G, x)
  G[1] = -2.0 * (1.0 - x[1]) - 400.0 * (x[2] - x[1]^2) * x[1]
  G[2] = 200.0 * (x[2] - x[1]^2)
end

function optimcallback(opt,Ψ)

  @info "fevalctr: $(Ψ.fevalctr)"
  return Ψ.fevalctr > 3
end

function mwe()
  Ψ=AdditionalState(0)

  #prep for the optimization and set parameters
  obj(x)=f(x,Ψ)
  callback(opt) = optimcallback(opt,Ψ)
  optimoptions = Optim.Options(iterations = 10, callback=callback)
  x = zeros(2)
  lower = [-100.,-100.]
  upper = [Inf, Inf]

  res=optimize(obj, g!, lower, upper, x, Fminbox(LBFGS()), optimoptions)
  println(res)
end

mwe()
```

Output:
```
[ Info: fevalctr: 2
[ Info: fevalctr: 11
[ Info: fevalctr: 13
... Omitting a bunch...
[ Info: fevalctr: 1561
[ Info: fevalctr: 1563
[ Info: fevalctr: 1565
 * Status: success

 * Candidate solution
    Final objective value:     1.163764e-16

 * Found with
    Algorithm:     Fminbox with L-BFGS

 * Convergence measures
    |x - x'|               = 2.95e-10 ≰ 0.0e+00
    |x - x'|/|x'|          = 2.09e-10 ≰ 0.0e+00
    |f(x) - f(x')|         = 4.36e-17 ≰ 0.0e+00
    |f(x) - f(x')|/|f(x')| = 3.74e-01 ≰ 0.0e+00
    |g(x)|                 = 8.49e-09 ≤ 1.0e-08

 * Work counters
    Seconds run:   1  (vs limit Inf)
    Iterations:    380
    f(x) calls:    232203
    ∇f(x) calls:   232203
```

mhangaard:
Related to #822 ?
I also have this issue, running GradientDescent and Optim v0.22.0.
My Optim.Options
```
[...]

Issue #818, opened 2020-05-18
Use AbstractConstrainedOptimizer as the parent type for all the constrained optimizers?


abhigupta768:
Hey, I wanted to know if there has been discussion on using AbstractConstrainedOptimizer uniformly within Optim as the parent type for the constrained optimizers. Currently only `Fminbox` follows this where as `SAMIN<:ZerothOrderOptimizer` and `IPNewton` defines a new abstract `ConstrainedOptimizer` type.

If you agree I can put together the PR for this. Thanks.

pkofod:
Hi! The very short ansewr is: no there hasn't been any real discussion around this. Can you flesh out your plan a bit? Thanks

PR #856, opened 2020-09-05
Fix some outstanding Fminbox issues

### LINKED ISSUE #821: Confusing behavior of f_abstol with Fminbox
I'm having a hard time understanding why having a higher (more lenient) tolerance results in _more_ iterations when using Fminbox. 
```
using Optim
tols = @. 10. ^ - (1:12)  # .1, .01 ...
map(tols) do tol
    # opt = optimize(f, zeros(2), LBFGS(), Optim.Options(f_abstol=tol))
    opt = optimize(f, -2ones(2), 2ones(2), zeros(2),  Fminbox(LBFGS()), Optim.Options(f_abstol=tol))
    opt.iterations
end |> println
# prints: [11, 6, 6, 4, 4, 4, 3, 3, 3, 3, 3, 3]
```
Note that if I switch to unbounded optimization (switch the comment) I get the expected behavior of increasing iterations: `[3, 18, 21, 21, 23, 23, 23, 24, 24, 24, 24, 24]`

Is this expected behavior?

More pragmatically, I have a bounded optimization problem where I only need three digits of precision and I can see that the optimizer is spending the vast majority of iterations within this range of the final result, no matter what I set f_abstol to.

pkofod:
This PR will fix #821 #818 and #834 .


fredcallaway:
Has this update been pushed to Pkg? (i.e can I get it with `] update Optim`)

PR #858, opened 2020-09-06
Make it easier to use IPNewton if you only have bounds.

pkofod:
As an alternative to Fminbox when you have second order information.


pkofod:
Fixes https://github.com/JuliaNLSolvers/Optim.jl/issues/600 by making it easier to use the primal-dual alg when you have the hessian.

PR #878, opened 2020-10-26
Loosen types with ldiv! and dot of InverseDiagonal

jonniedie:
All that was required to make [this example](https://discourse.julialang.org/t/nonlinear-optimization-help-optim-jump-something-else/47163/2) work was to patch in a method for `ldiv!`. Loosening this type restriction should let even more array types work here. I've tested this and it doesn't look like it introduces any method ambiguities.

### DIFF
--- a/src/multivariate/precon.jl
+++ b/src/multivariate/precon.jl
@@ -51,8 +51,8 @@ end
 mutable struct InverseDiagonal
    diag
 end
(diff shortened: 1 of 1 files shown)


pkofod:
Thanks, though I fixed this while also not doing type piracy in a recent PR :)

PR #887, opened 2020-12-14
Fix compat with Julia version

briochemc:
I'm not sure but I thought I should suggest moving the compact from Julia v1.2 to Julia v1 (so that current stable v1.5.3 and future versions are in)


pkofod:
I feel like there's a reason why this was specifically 1.2 because there were some package issues on earlier versions. If you just specify a number it respects semver, so "1.2" means `[1.2.0, 2.0.0)`. Please reopen if you intended for v1.0.x and v1.1.y to be supported. Thanks though.

briochemc:
No, you're right, I got confused, sorry for the noise!

pkofod:
Hey, every contribution is welcome. Made me double-check if nothing else :)

PR #888, opened 2020-12-14
Add Github Action CI

briochemc:
Adds CI for Linux, Windows, and OSX via GitHub actions (timely considering the Travis recent move to credits). There are 6 files (two per OS, for the current stable versions `v1` and the `nightly` version).

Note you will need to add you CODECOV token to the secrets for the coverage during these runs to be accounted for.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,40 @@
+name: Linux
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


pkofod:
Huh, that's weird. It looks like there are ambiguities on Linux?

pkofod:
Oh, that was all on nightly. Is it possible to allow nightly to fail?

briochemc:
Well I’m not sure (there was no easy way back when I made those for another
package of mine) but that’s why the actions are split in 6 scripts — so
that you can make a badge for each OS on the stable version and “ignore”
the CI status of the runs on nightly (but still run them for your own
information).

pkofod:
Okay. It might also force me to fix the problem :) Thanks so much for this. Saved me a bunch of time.

PR #889, opened 2020-12-15
make some utils gpu friendly

piever:
With this, I can do:

```julia
julia> using Optim, CUDA

julia> init = CUDA.rand(10, 10);

julia> function fg!(_, G, w)
           res = sum(abs2, w)
           if !isnothing(G)
               @. G = 2w
           end
           return res
       end
fg! (generic function with 1 method)

julia> CUDA.allowscalar(false)

julia> optimize(Optim.only_fg!(fg!), init)
 * Status: success

 * Candidate solution
    Final objective value:     0.000000e+00

 * Found with
    Algorithm:     L-BFGS

 * Convergence measures
    |x - x'|               = 9.90e-01 ≰ 0.0e+00
    |x - x'|/|x'|          = Inf ≰ 0.0e+00
    |f(x) - f(x')|         = 3.48e+01 ≰ 0.0e+00
    |f(x) - f(x')|/|f(x')| = Inf ≰ 0.0e+00
    |g(x)|                 = 0.00e+00 ≤ 1.0e-08

 * Work counters
    Seconds run:   5  (vs limit Inf)
    Iterations:    1
    f(x) calls:    3
    ∇f(x) calls:   3
```

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -23,7 +23,7 @@ after_while!(d, state, method, options) = nothing
 
 function initial_convergence(d, state, method::AbstractOptimizer, initial_x, options)
     gradient!(d, initial_x)
-    norm(gradient(d), Inf) < options.g_abstol
+    maximum(abs, gradient(d)) < options.g_abstol
 end
 initial_convergence(d, state, method::ZerothOrderOptimizer, initial_x, options) = false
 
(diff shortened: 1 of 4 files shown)


pkofod:
Thanks! It's so annoying that those norm calls don't just work on gpus. Do you know why?

piever:
Looks like the fallbacks are done with for loops instead of `mapreduce`, e.g., [here](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L464). I'll try a PR as soon as I have a bit more time (I hope there aren't too many weird corner cases like `NaN` and such to cover).

Issue #865, opened 2020-09-07
BarrierWrapper has no field f_calls


grahamas:
I don't have a MWE at the moment; I'll try to get one later.  For the moment, I've fixed things by reverting to Optim@0.22.

```
ERROR: LoadError: type BarrierWrapper has no field f_calls                                                                                                                                     
Stacktrace:                                                                                    
 [1] getproperty(::Optim.BarrierWrapper{NLSolversBase.OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}},Optim.BoxBarrier{Array{Float64,1},Array{Float64,1}},Float64,Float64,Array{F
loat64,1}}, ::Symbol) at ./Base.jl:33                                                                                                                                                          
 [2] value!!(::Optim.BarrierWrapper{NLSolversBase.OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}},Optim.BoxBarrier{Array{Float64,1},Array{Float64,1}},Float64,Float64,Array{Float
64,1}}, ::Array{Float64,1}) at /home/graham/.julia/packages/NLSolversBase/5oIMo/src/interface.jl:7                                                                                             
 [3] initial_state(::Optim.NelderMead{Optim.AffineSimplexer,Optim.AdaptiveParameters}, ::Optim.Options{Float64,Nothing}, ::Optim.BarrierWrapper{NLSolversBase.OnceDifferentiable{Float64,Array{
Float64,1},Array{Float64,1}},Optim.BoxBarrier{Array{Float64,1},Array{Float64,1}},Float64,Float64,Array{Float64,1}}, ::Array{Float64,1}) at /home/graham/.julia/packages/Optim/BBM06/src/multiva
riate/solvers/zeroth_order/nelder_mead.jl:158                                                  
 [4] optimize(::NLSolversBase.OnceDifferentiable{Float64,Array{Float64,1},Array{Float64,1}}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Optim.Fminbox{Optim.NelderMead{Optim
.AffineSimplexer,Optim.AdaptiveParameters},Float64,Optim.var"#45#47"}, ::Optim.Options{Float64,Nothing}) at /home/graham/.julia/packages/Optim/BBM06/src/multivariate/solvers/constrained/fminb
ox.jl:288                                                                                      
[...]

PR #891, opened 2020-12-17
Fix reset of LBFGS.

### LINKED ISSUE #874: `optimize` passes NaN as x value to function
*Edit:* I forgot to mention the version; I have version 1.2.0 installed in the environment I am running in.

I encountered a similar situation to that described in #581 : the function I am trying to optimize receives NaN as input from `optimize`. The cause appears to be non-deterministic, receiving the NaN only happens occasionally, but I encountered it several times because I am evaluating performance and thus running with the same inputs repeatedly. Pasting the code below into a blank Pluto notebook and running the cell repeatedly will eventually produce the following output:

````
* Status: success

 * Candidate solution
    Final objective value:     9.999990e+05

 * Found with
    Algorithm:     Fminbox with L-BFGS

 * Convergence measures
    |x - x'|               = NaN ≰ 0.0e+00
    |x - x'|/|x'|          = NaN ≰ 0.0e+00
    |f(x) - f(x')|         = 0.00e+00 ≤ 0.0e+00
    |f(x) - f(x')|/|f(x')| = 0.00e+00 ≤ 0.0e+00
    |g(x)|                 = NaN ≰ 1.0e-08

 * Work counters
    Iterations:    3
    f(x) calls:    39
    ∇f(x) calls:   39
````

The final objective value of 999999 indicates that a NaN value was received. The summary reports success, as does the succeeded() method, which seems wrong (I don't think it succeeded)? Would there be a problem with checking whether a NaN was produced in succeeded() etc?

MWE:

````julia
begin
	using Optim
	using QuadGK

	function fdead(a :: Real, b :: Real, μ :: Real, rec :: Integer)
		y, err = quadgk(x -> (μ * exp(-μ * x)) / (exp(-a + b * x) + 1), 0, rec)
		y * exp(μ * rec)
	end

	function loglike3(μ :: Real, a, b, rec, T)
		if isnan(μ)
			return 999999
		end
		dead = fdead(a, b, μ, rec)
		live = 1. / (1 + exp(-a + b * rec))
		ll = -μ * T + log(live + dead)
		-ll
	end

	optimize(x -> loglike3(x[1], 0.625505, 0.03187, 246, 95), [1e-8], [1.], [0.5])
end
````

pkofod:
Fixes #874

### DIFF
--- a/src/multivariate/solvers/first_order/l_bfgs.jl
+++ b/src/multivariate/solvers/first_order/l_bfgs.jl
@@ -154,7 +154,7 @@ function reset!(method, state::LBFGSState, obj, x)
     value_gradient!(obj, x)
     project_tangent!(method.manifold, gradient(obj), x)
 
-    pseudo_iteration = 0
+    state.pseudo_iteration = 0
 end
 function initial_state(method::LBFGS, options, d, initial_x)
     T = real(eltype(initial_x))
--- a/test/multivariate/solvers/constrained/fminbox.jl
+++ b/test/multivariate/solvers/constrained/fminbox.jl
@@ -138,3 +138,7 @@ end
         optimize(x -> sqrt(x[1]), [0.0], [10.0], [1.0], Fminbox(m); autodiff=:forwarddiff)
     end
 end
+
+@testset "#865"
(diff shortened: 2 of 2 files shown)


pkofod:
Fixes #865

PR #898, opened 2021-01-16
Change order of ldiv and - in sparse Newton.

mikkoku:
Negating the search direction instead of the whole hessian saves some 
computation and allocations.


pkofod:
Makes sense, thanks.

PR #904, opened 2021-01-29
Updated code to match online documentation

krislock:
The documentation https://julianlsolvers.github.io/Optim.jl/stable/#algo/simulated_annealing/ has
```julia
neighbor!(x_proposal::Array, x::Array)
```
so I propose changing the order of `x_proposal` and `x` in the simulated annealing code.

### DIFF
--- a/src/multivariate/solvers/zeroth_order/simulated_annealing.jl
+++ b/src/multivariate/solvers/zeroth_order/simulated_annealing.jl
@@ -2,7 +2,7 @@ log_temperature(t) = 1 / log(t)
 
 constant_temperature(t) = 1.0
 
-function default_neighbor!(x::AbstractArray{T}, x_proposal::AbstractArray) where T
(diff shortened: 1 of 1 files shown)


pkofod:
Thanks! Good catch. The "problem" with this change is that it's almost impossible to deprecate properly, and this would be a breaking change. So while it goes against Julia conventions I think we might have to wait with this change (we can keep the PR open) until the next breaking tag. For now I think we could change the docs to reflect the correct order. Does that make sense? :)

krislock:
Yes, I agree. It would be bad if this change causes things to break. In that case, changing the documentation would be better. I could submit the pull request to change the documentation, if you would like.

pkofod:
> I could submit the pull request to change the documentation, if you would like.

That would be much appreciated! But please keep this PR open as it will help me remember to fix it :)

PR #913, opened 2021-04-08
relax type constraints to support Measurements

### LINKED ISSUE #823: how to make optim work with Measuements
Hi:
I want to use some Measurements variable in optim problem.
but have some error,how do I handle it,wish to get reply

using Measurements,Optim
f(x)=(1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
x=zeros(Measurement{Float64},2)
optimize(f,x)

longemen3000:
paves the way to use Optim with Measurements. calculations runs and seems to propagate the uncertainty. fixes #823 . fixing the show problem could be done in two ways:
- trying to use other way of formatting numbers, compatible with `Measurements`.
- displaying only the value. this can be done on a two liner:
```
using Printf,Measurements
Printf.tofloat(x::Measurement) = Printf.tofloat(Measurements.value(x))
```
if the user doesnt want to do this, they can directly use Optim.minimizer(res) or Optim.minimum(res) and not show the summary of the optimization procedure.
as this type relaxation doesnt seem to affect other number types, this is an strict improvement over the current situation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,10 +27,11 @@ StatsBase = "0.29, 0.30, 0.31, 0.32, 0.33"
 julia = "1.2"
 
 [extras]
+Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7"
 OptimTestProblems = "cec144fc-5a64-5bc6-99fb-dde8f63e154c"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "OptimTestProblems", "Random", "RecursiveArrayTools"]
+test = ["Test", "Measurements", "OptimTestProblems", "Random", "RecursiveArrayTools"]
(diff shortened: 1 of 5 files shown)


longemen3000:
on my local machine, only tests with Momentum Gradient Descent seem to fail,seems unrelated

longemen3000:
@pkofod can you check this PR? The printing mechanism on measurements has been already fixed

pkofod:
> on my local machine, only tests with Momentum Gradient Descent seem to fail,seems unrelated

Momentum gradient descent is very unstable. Last unit differences can cause it to explode. I'll have a look at the PR :)

pkofod:
@longemen3000 can you add a test?

longemen3000:
something like the example problem in #823 ?

longemen3000:
i added the #823 example problem, work well but im testing. after that i commit

longemen3000:
@pkofod can you check on this? It seems that the new github policy is blocking the test from running

pkofod:
Yes, but unfortunately the "approve and run" button leads me to a error 500 page ..

longemen3000:
The error in the tests is unrelated (convergence on particle swarm)

pkofod:
> The error in the tests is unrelated (convergence on particle swarm)

Yes that's just because I have to use StableRNGs for the tests. Thanks for the contribution once again :)

Boxylmer:
> paves the way to use Optim with Measurements. calculations runs **and seems to propagate the uncertainty.** 

Has anyone verified the resulting uncertainty is correctly propagated?

PR #929, opened 2021-06-20
Add MOI wrapper

blegat:
With this MOI wrapper, one should be able to use Optim with JuMP doing:
```julia
model = Model(Optim.Optimizer)
```
If there are nonlinear constraints created with `@NLconstraint`, it would select `IPNewton`.
Otherwise, this would select the method using `fallback_method` which is wrapped in `Fminbox` if there are variable bounds.
The objective need to be created with `@NLobjective`.

```julia
julia> model = Model(Optim.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: Optim

julia> @variable(model, x[1:2])
2-element Vector{VariableRef}:
 x[1]
 x[2]

julia> @NLobjective(model, Min, (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2)

julia> @NLconstraint(model, x[1]^2 + x[2]^2 <= 0.5^2)
(x[1] ^ 2.0 + x[2] ^ 2.0) - 0.5 ^ 2.0 ≤ 0

julia> optimize!(model)

julia> objective_value(model)
0.22226228958462926

julia> value.(x)
2-element Vector{Float64}:
 0.5287447092556342
 0.2782265698193508
```

*Note*: Quadratic and affine objective created with `@objective` and constraints created with `@constraint` are not supported directly by the wrapper but it might be usable with Optim without the need to change anything in this wrapper with the idea suggested in https://github.com/jump-dev/MathOptInterface.jl/issues/1397

This addresses the third item of https://github.com/JuliaNLSolvers/Optim.jl/issues/609


pkofod:
Thanks, this seems like a lot of work you've done. I'll review it

blegat:
@pkofod I just updated it to MOI v1. Could you get a look ?

blegat:
Bump :)

pkofod:
> :)

Sorry, I had not noticed any of these bumps.

blegat:
One of the questions of this talk: https://meetings.siam.org/sess/dsp_talk.cfm?p=129039 was whether Optim could be used from JuMP

ChrisRackauckas:
Optimization.jl wraps both Optim.jl and the JuMP methods (through MOI), so it effectively achieves the same goal.

joaquimg:
I dont see how it is the same goal.
The user wanted to call Optim from JuMP.

pkofod:
> One of the questions of this talk: https://meetings.siam.org/sess/dsp_talk.cfm?p=129039 was whether Optim could be used from JuMP

want to get it finished if I promise to review and get it in this time? :)

pkofod:
You are way too patient. Thanks @blegat !

PR #938, opened 2021-08-03
Fix edit link at bottom of each page

ArnoStrouwen:
Edit links point to
https://github.com/JuliaNLSolvers/Optim.jl/edit/master/docs/
and not to
https://github.com/JuliaNLSolvers/Optim.jl/edit/master/docs/src/


pkofod:
Thanks, I had not realized this.

PR #939, opened 2021-08-04
Start Newton-Krylov docs

ArnoStrouwen:
Still needs some details to be filled in.

### DIFF
--- a/docs/src/algo/newton_krylov_trust_region.md
+++ b/docs/src/algo/newton_krylov_trust_region.md
@@ -1,0 +1,46 @@
+# Hessian-Free Newton's Method With a Trust Region
+## Constructor
+```julia
+KrylovTrustRegion(; initial_radius = 1.0,
+                    max_radius = 100.0,
+                    eta = 0.1,
+                    rho_lower = 0.25,
+                    rho_upper = 0.75,
+                    cg_tol = 0.01)
+```
+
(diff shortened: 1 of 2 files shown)


docs/src/algo/newton_krylov_trust_region.md
```diff
@@ -0,0 +1,46 @@
...
+
+Newton's method optimizes a quadratic approximation to a function. When a function is well approximated by a quadratic (for example, near an optimum), Newton's method converges very quickly by exploiting the second-order information in the Hessian matrix.  However, when the function is not well-approximated by a quadratic, either because the starting point is far from the optimum or the function has a more irregular shape, Newton steps can be erratically large, leading to distant, irrelevant areas of the space.
+
+The standard Newton method requires repeatedly calculating the Hessian of the optimization objective to solve a linear system, which for large problems can require a lot of time and memory. A completely different way to solve large linear systems is to use a Krylov subpsace method [2]. This method does not require the calculation of the full Hessian, instead it uses Hessian vector products to reduce memory and time requirements.
```
prbzrg, line 26:
```suggestion
The standard Newton method requires repeatedly calculating the Hessian of the optimization objective to solve a linear system, which for large problems can require a lot of time and memory. A completely different way to solve large linear systems is to use a Krylov subspace method [2]. This method does not require the calculation of the full Hessian, instead it uses Hessian vector products to reduce memory and time requirements.
```

prbzrg (changes requested):
I just fixed a typo

Issue #941, opened 2021-08-12
Minor issues in documentation and traces for Nelder Mead


eirikbrandsaas:
Hi,

While trying to understand what is going on in my own code I found some minor errors. I'll fix a few of them in a pull request, but other's I tried but wasn't able too.

- [ ] Some of the traces are not mentioned at https://julianlsolvers.github.io/Optim.jl/stable/#user/minimization/. (`centroid_trace`, `simplex_trace`, `simplex_values_trace`)
- [ ]  `function simplex_value_trace(r::MultivariateOptimizationResults)` doesn't work, missing ` tr = trace(r)`
- [x] The output from `simplex_trace` and `simplex_values_trace` only give the values for the last iteration:

For the last one, here is a MWE:
```Julia
using Optim
f(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
x0 = [0.0, 0.0]
res = optimize(f, x0,iterations=3,show_trace=true,store_trace=true,extended_trace=true,trace_simplex=true );

Optim.centroid_trace(res) # Works
Optim.simplex_trace(res) # Doesn't work
```
```
julia> res = optimize(f, x0,iterations=3,show_trace=true,store_trace=true,extended_trace=true,trace_simplex=true );
Iter     Function value    √(Σ(yᵢ-ȳ)²)/n 
------   --------------    --------------
     0     9.506641e-01     4.576214e-02
 * simplex_values: [1.0, 0.9506640624999999, 1.0625]
 * time: 0.0009310245513916016
 * step_type: initial
 * centroid: [0.0125, 0.0]
 * simplex: [[0.0, 0.0], [0.025, 0.0], [0.0, 0.025]]
     1     9.506641e-01     2.023096e-02
 * simplex_values: [1.0, 0.9506640624999999, 0.9793678283691405]
 * time: 0.0014400482177734375
 * step_type: outside contraction
 * centroid: [0.0125, 0.0]
 * simplex: [[0.0, 0.0], [0.025, 0.0], [0.018750000000000003, -0.0125]]
     2     9.506641e-01     2.172172e-02
 * simplex_values: [0.9262175083160399, 0.9506640624999999, 0.9793678283691405]
 * time: 0.0019030570983886719
 * step_type: expansion
 * centroid: [0.021875000000000002, -0.00625]
 * simplex: [[0.065625, -0.018750000000000003], [0.025, 0.0], [0.018750000000000003, -0.0125]]
[...]

PR #942, opened 2021-08-12
Fixes 2/3 of problems with Nelder Mead related traces

eirikbrandsaas:
Fixes the first two the issues raised in #941 

I tried to fix the last one, but that was beyond my capabilities. If you give me a pointer I should be able to figure it out.

### DIFF
--- a/docs/src/user/minimization.md
+++ b/docs/src/user/minimization.md
@@ -214,6 +214,12 @@ Defined for multivariate optimization:
 * `g_converged(res)`
 * `initial_state(res)`
 
+Defined only for Nelder Mead multivariate optimizatoin:
+
+* `centroid_trace(res)`
+* `simplex_trace(res)`
+* `simplex_value_trace(res)`
+
 ## Input types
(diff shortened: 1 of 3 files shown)


pkofod:
Thanks for looking into this. It looks fine. I would appreciate if you could add a test for the missing `tr` thing, just to make sure that the code path is run as part of the tests.

eirikbrandsaas:
Sure, at least if you can slightly more specific on what you want. You just want me write a test that simply checks that the command works? In which of the many `test/` files should it go?

pkofod:
> Sure, at least if you can slightly more specific on what you want. You just want me write a test that simply checks that the command works? In which of the many `test/` files should it go?

Yes, just checking that what you changed does what you intended.  Here would be a good place https://github.com/JuliaNLSolvers/Optim.jl/blob/master/test/multivariate/solvers/zeroth_order/nelder_mead.jl. Just add a testset with title `#942`

eirikbrandsaas:
Okay, I'll get it done sometime "soon".

eirikbrandsaas:
"Soon" became 6 months, but hey, who is counting?

I see now that you wanted me to post a separate pull request, but hope this works.

Note that, as mentioned in #941 that the stored trace is wrong, it replaces all simplex values with the last one. (That's why there are two `@test_broken` in the pull request. 

Too see this:
```julia
using Optim
x0 = [0.0,0.0]
res = optimize(f, x0, NelderMead(),Optim.Options(store_trace=true,trace_simplex=true,extended_trace=true,iterations=2,show_trace=true))
display(res.trace)
```
and compare the output of the simplexes (both y-values and x-simplexes) from the "running" trace and the stored res.trace.

eirikbrandsaas:
@pkofod Just pinging. 

(See comment above)

PR #946, opened 2021-09-28
BFGS - GPU

vpuri3:
BLAS-ified update_h! call in BFGS() so it can work on GPU

### DIFF
--- a/src/multivariate/solvers/first_order/bfgs.jl
+++ b/src/multivariate/solvers/first_order/bfgs.jl
@@ -161,11 +161,13 @@ function update_h!(d, state, method::BFGS)
 
         # TODO BLASify this
         # invH = invH + c1 * (s * s') - c2 * (u * s' + s * u')
-        for i in 1:n
-            @simd for j in 1:n
-                @inbounds state.invH[i, j] += c1 * state.dx[i] * state.dx[j]' - c2 * (state.u[i] * state.dx[j]' + state.u[j]' * state.dx[i])
-            end
-        end
+        state.invH .+= c1.*(state.dx * state.dx') .-
+                       c2.*(state.u  * state.dx' .+ state.dx * state.u')
+#       for i in 1:n
+#           @simd for j in 1:n
+#               @inbounds state.invH[i, j] += c1 * state.dx[i] * state.dx[j]' - c2 * (state.u[i] * state.dx[j]' + state.u[j]' * state.dx[i])
+#           end
+#       end
     end
 end
 


src/multivariate/solvers/first_order/bfgs.jl
```diff
@@ -161,11 +161,13 @@ function update_h!(d, state, method::BFGS)
 
         # TODO BLASify this
         # invH = invH + c1 * (s * s') - c2 * (u * s' + s * u')
-        for i in 1:n
-            @simd for j in 1:n
-                @inbounds state.invH[i, j] += c1 * state.dx[i] * state.dx[j]' - c2 * (state.u[i] * state.dx[j]' + state.u[j]' * state.dx[i])
-            end
-        end
+        state.invH .+= c1.*(state.dx * state.dx') .-
```
ChrisRackauckas, line 164: this introduces quite a bit of intermediate allocations. You probably want to keep a fast `ArrayInterface.fast_scalar_indexing(state.u)` check and keep the branch

vpuri3:
@ChrisRackauckas  `ArrayInterface.can_setindex(state.u)` evaluates to `true`. I also replaced the oneliner with calls to `mul!`. what are next steps?

ChrisRackauckas:
It's true for arrays

longemen3000:
the errors are relevant, as the `update_h` function is tested against matrix and tensor types. maybe adding `vec(x)` to `state.dx` and `state.u` could solve the problem. Finally, there are some new unused dependencies (`CUDA`, `ArrayInterface`) in the `Project.toml`

ChrisRackauckas:
The code is also not correct because it's just overwriting `state.invH` instead of accumulating.

And it adds a Manifest.toml.

killah-t-cell:
I'm new here, but based on this MWE (which I wrote based on the Rosenbrock OptimTestProblems), the code with mul! seems to do be doing the same thing:

```julia
# with loop
c1 = 41.772696387872294
c2 = 0.0319955499083424
invH1 = [1.0 -0.0; -0.0 1.0]
dx = [0.14330445374562797, 0.0584916137737257]
u = [186.9231445284796, 76.37722725998003]

for i in 1:2
    @simd for j in 1:2
        @inbounds invH1[i, j] += c1 * dx[i] * dx[j]' - c2 * (u[i] * dx[j]' + u[j]' * dx[i])
    end
end
@show invH1

# with mul!
c1 = 41.772696387872294
c2 = 0.0319955499083424
invH2 = [1.0 -0.0; -0.0 1.0]
dx = [0.14330445374562797, 0.0584916137737257]
u = [186.9231445284796, 76.37722725998003]

mul!(invH2,dx,dx', c1,1)
mul!(invH2,u ,dx',-c2,1)
mul!(invH2,dx,u' ,-c2,1) 
@show invH2

invH1 - invH2
(invH1 .- invH2) .< 1e-6
```

Am I missing something?

longemen3000:
> EDIT: It seems like the new code can't handle tensors or manifolds.

for the tensor part, i think wrapping the tensor in `vec` would suffice. i don't know if doing the same with the manifold would help. the thing is that, by iterating over one dimension, the old code assumes implicitly vector form independent of the actual shape, but with an explicit `mul` this assumption is made explicit and for Arrays that are not in vector form, the new code fails. 
Something like this would solve the dimensionality problem:
```
mul!(invH2,vec(dx),vec(dx)', c1,1)
mul!(invH2,vec(u) ,vec(dx)',-c2,1)
mul!(invH2,vec(dx),vec(u)' ,-c2,1) 

```

ChrisRackauckas:
```julia
mul!(invH2,dx,dx', c1,1)
mul!(invH2,u ,dx',-c2,1)
mul!(invH2,dx,u' ,-c2,1) 
```

is equivalent to

```julia
mul!(invH2,dx,u' ,-c2,1) 
```

because it's just overwriting the same cache. You need to add the previous calculation to the last.

killah-t-cell:
@ChrisRackauckas I don't think so. As per [the definition a 5-argument mul!](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Low-level-matrix-operations), `mul!(C,A,B,α,β)` combines inplace matrix-matrix or matrix-vector multiply-add of the form `ABα+Cβ` and stores it in `C`. So an addition is already happening here. I wrote this short program to convince myself that I understood this right, and it seems that both approaches are equivalent (though they sometimes vary by 1e-16 which I assume is just a numerical rounding error).

```julia
# Declarations
invH1 = [2.0 40; -4.0 7.0]
invH2 = [2.0 40; -4.0 7.0]
invH3 = [2.0 40; -4.0 7.0]
c1 = 41.772696387872294
c2 = 0.0319955499083424
dx = [0.14330445374562797, 0.0584916137737257]
u = [186.9231445284796, 76.37722725998003]

# with loop
for i in 1:2
    @simd for j in 1:2
        @show c1 * dx[i] * dx[j]' - c2 * (u[i] * dx[j]')
        @show -c2*u[j]' * dx[i]
        @inbounds invH1[i, j] += c1 * dx[i] * dx[j]' - c2 * (u[i] * dx[j]' + u[j]' * dx[i])
    end
end

# with mul!
mul!(invH2,dx,dx', c1,1) 
mul!(invH2,u ,dx',-c2,1)
mul!(invH2,dx,u' ,-c2,1) 

# with A*B*α + C -> C
invH3 += dx*dx'*c1
invH3 += u*dx'*-c2
invH3 += dx*u'*-c2

# Test equivalence
@show invH1
@show invH2
@show invH3

invH1 ≈ invH2
invH2 ≈ invH3
invH1 - invH2
invH2 == invH3

# prove inequivalence of https://github.com/JuliaNLSolvers/Optim.jl/pull/946#issuecomment-939242924
invH4 = [2.0 40; -4.0 7.0]
mul!(invH4,dx,u' ,-c2,1) 

invH5 = [2.0 40; -4.0 7.0]
mul!(invH5,dx,dx', c1,1) 
mul!(invH5,u ,dx',-c2,1)
mul!(invH5,dx,u' ,-c2,1) 
invH4 ≈ invH5
```

Once again LMK if there is something I don't get.

killah-t-cell:
I opened a PR that continues this work here https://github.com/JuliaNLSolvers/Optim.jl/pull/950. I didn't know how to add to this one. I hope this isn't rude to @vpuri3. This is mostly his work!

ChrisRackauckas:
You'll still want to keep the loops on CPU to fuse the operations, and use LoopVectorization there.

killah-t-cell:
So you’re saying I should create a mul! Function using LoopVectorization like in the example here? 

https://juliasimd.github.io/LoopVectorization.jl/stable/examples/matrix_multiplication/#Matrix-Multiplication

```
function A_mul_B!(C, A, B)
    @turbo for n ∈ indices((C,B), 2), m ∈ indices((C,A), 1)
        Cmn = zero(eltype(C))
        for k ∈ indices((A,B), (2,1))
            Cmn += A[m,k] * B[k,n]
        end
        C[m,n] = Cmn
    end
end
```

ChrisRackauckas:
No, that `@simd` loop fuses the 3 operations, so I would expect that a LoopVectorization form would outperform the BLAS one on Arrays for at least "small enough" equations. It's worth testing and possibly specializing.

ChrisRackauckas:
>Would you recommend moving the arrays to CPU with Array for the update?

No, it should branch on the type.

killah-t-cell:
I would probably just remove the new manifest.toml you added and maybe delete the `# TODO BLASify this` and `@show ArrayInterface.can_setindex(state.u)` lines.

vpuri3:
@ChrisRackauckas is this PR good to go?

pkofod:
the nightlies are unrelated I think

PR #947, opened 2021-10-01
Clarify discussion of initial simplex in NelderMead

eirikbrandsaas:
- Added the default values for the `AffineSimplexer` (I think it is `a=0.025`,`b=0.5`)?
- Split up the Simplex discussion into two parts, first on the default in `Optim.jl` and then how to change them]
- Added a small example on how to change the default values of `AffineSimplexer`


eirikbrandsaas:
(Finally, the docs aren't building/redeploying correctly). For example, the NelderMead page (https://julianlsolvers.github.io/Optim.jl/stable/#algo/nelder_mead) still doesn't have the correct Matlab simplexer even though they are correct on master.

PR #956, opened 2021-10-30
fix for ML example

mcreel:
The ML example in the documentation (https://julianlsolvers.github.io/Optim.jl/stable/#examples/generated/maxlikenlm/) has some problems. The t-statistics are not correct, as can be verified by computing the OLS estimates. The error is that the transformation from log sigma to sigma is not correctly done. The Hessian matrix is computed after directly exponentiating the log sigma parameter, while the likelihood function is not prepared to accept sigma rather than log sigma, so the resulting Hessian is not correct. This PR uses the delta method to do the transformation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "1.5.0"
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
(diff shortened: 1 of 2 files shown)


pkofod:
Thanks, Looks good. But ForwardDiff appear to not be in the manifest.

mcreel:
Yes, I saw the error. Where should I add it, at the top level of Optim.jl?

pkofod:
Great, thanks! I'll fix the nightly errors shortly.

PR #961, opened 2021-11-08
check for nothing with ===/!== instead of ==/!=

st--:
As pointed out in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/240#discussion_r744598754 - I couldn't find good documentation in base julia beyond a bullet point in https://docs.julialang.org/en/v1/manual/noteworthy-differences/, but there's also https://stackoverflow.com/questions/56852880/comparing-julia-variable-to-nothing-using-or ...

Optim also requires Julia >= 1.2, so if you prefer we could change the checks to `isnothing`/`!isnothing`?

### DIFF
--- a/docs/src/user/tipsandtricks.md
+++ b/docs/src/user/tipsandtricks.md
@@ -64,14 +64,15 @@ For example, here we define a function `fg!` to compute the objective function a
 the gradient, as required:
 
 ```julia
-function fg!(F,G,x)
+function fg!(F, G, x)
   # do common computations here
   # ...
-  if G != nothing
+  if G !== nothing
     # code to compute gradient here
     # writing the result to the vector G
+    # G .= ...
   end
-  if F != nothing
+  if F !== nothing
     # value = ... code to compute objective function
     return value
   end
@@ -91,10 +92,10 @@ Optim.optimize(Optim.only_fg!(fg!), [0., 0.], Optim.LBFGS())
 Similarly, for a computation that requires the Hessian, we can write:
 
 ```julia
-function fgh!(F,G,H,x)
-  G == nothing || # compute gradient and store in G
-  H == nothing || # compute Hessian and store in H
-  F == nothing || return f(x)
+function fgh!(F, G, H, x)
+  G === nothing || # compute gradient and store in G
+  H === nothing || # compute Hessian and store in H
+  F === nothing || return f(x)
   nothing
 end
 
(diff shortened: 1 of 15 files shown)


docs/src/user/tipsandtricks.md
```diff
@@ -91,10 +92,10 @@ Optim.optimize(Optim.only_fg!(fg!), [0., 0.], Optim.LBFGS())
...
 
 ```julia
-function fgh!(F,G,H,x)
-  G == nothing || # compute gradient and store in G
-  H == nothing || # compute Hessian and store in H
-  F == nothing || return f(x)
+function fgh!(F, G, H, x)
+  G === nothing || # compute gradient and store in G
+  H === nothing || # compute Hessian and store in H
+  F === nothing || return f(x)
```
st--, line 98:
```suggestion
  isnothing(G) || # compute gradient and store in G
  isnothing(H) || # compute Hessian and store in H
  isnothing(F) || return f(x)
```
if you prefer

docs/src/user/tipsandtricks.md
```diff
@@ -64,14 +64,15 @@ For example, here we define a function `fg!` to compute the objective function a
 the gradient, as required:
 
 ```julia
-function fg!(F,G,x)
+function fg!(F, G, x)
   # do common computations here
   # ...
-  if G != nothing
+  if G !== nothing
```
st--, line 70: or
```suggestion
  if !isnothing(G)
```

docs/src/user/tipsandtricks.md
```diff
@@ -64,14 +64,15 @@ For example, here we define a function `fg!` to compute the objective function a
...
     # code to compute gradient here
     # writing the result to the vector G
+    # G .= ...
   end
-  if F != nothing
+  if F !== nothing
```
st--, line 75: or
```suggestion
  if !isnothing(F)
```

jd-foster:
I don't have a view for Optim but note that it's just [`isnothing(x) = x === nothing`](https://github.com/JuliaLang/julia/blob/48ed43fc8e47d1e0cbe768cb4b691eaba697eebc/base/some.jl#L70) in Base.

st--:
I don't mind, I'd be happy for it to be merged as is, I just wanted to point it out in case you thought it'd be neater:)

st--:
Hi, could one of the maintainers please approve the workflow run?

st--:
@pkofod would you say this counts as a bugfix (so just have to bump the patch version)?

pkofod:
Looks good, thanks.

PR #966, opened 2021-12-21
Allow univariate nelder mead.

### LINKED ISSUE #965: To solve 1D problem by adding 5% change at each step in Optim.jl methods
I am trying to use equivalent for fminsearch in Matlab for 1D problem. There are two issues here:
1- fminsearch uses Neldermead for scalar and works well. It looks like optim.jl cannot be called to solve 1d problem using Nelder mead
2- fminsearch gets an initial guess. At each step it changes the initial guess by 5% until convergence. In julia there are ways of changing NelderMead to comply with that (initial_simplex). Optim.jl also offers Brent() and Goldensection() for 1d, but, neither of those accept initial guess. They both need a bound.

I have a complicated code (both in Matlab and Julia). User can choose among some parameters to optimize some of them. This works well in Matlab since single parameter can be picked as well. In Julia I used "if condition" to switch to Brent(). Based on the problem the bound might change. I am looking for a way of using 5% change from initial guess (or 5% change from mid point of a more general bound). I think Brent uses the mid point of the bound as initial value. Which is totally fine. But the problem happens when for the next step the new guess is significantly different from former one. This could cause NAN in my calculation (due to the number being physically unacceptable). So the question is: Is there any 1D solver which operates by changing 5% from initial guess?

pkofod:
Fixes #965

```
julia> optimize(x->x[1]^2, [4.0])
 * Status: success

 * Candidate solution
    Final objective value:     3.187452e-09

 * Found with
    Algorithm:     Nelder-Mead

 * Convergence measures
    √(Σ(yᵢ-ȳ)²)/n ≤ 1.0e-08

 * Work counters
    Iterations:    11
    f(x) calls:    25
```

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -32,9 +32,6 @@ end
 function optimize(d::D, initial_x::Tx, method::M,
                   options::Options{T, TCallback} = Options(;default_options(method)...),
                   state = initial_state(method, options, d, initial_x)) where {D<:AbstractObjective, M<:AbstractOptimizer, Tx <: AbstractArray, T, TCallback}
-    if length(initial_x) == 1 && typeof(method) <: NelderMead
-        error("You cannot use NelderMead for univariate problems. Alternatively, use either interval bound univariate optimization, or another method such as BFGS or Newton.")
-    end
 
     t0 = time() # Initial time stamp used to control early stopping by options.time_limit
     tr = OptimizationTrace{typeof(value(d)), typeof(method)}()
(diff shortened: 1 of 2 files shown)


cossio:
It was only a matter of removing the coded restriction and it just works. Nice.

Assuming all tests pass, can you tag a release with this after merging?

pkofod:
> Assuming all tests pass, can you tag a release with this after merging?

I plan to yes. It appears that the nightly tests are very very slow to finish but I won't let that stop this PR. maybe it's some inference thing going bad.

Issue #915, opened 2021-04-14
Newlines in two-argment show


andreasnoack:
These should be avoided in e.g. https://github.com/JuliaNLSolvers/Optim.jl/blob/adc5b277b3f915c25233b45f8f2dd61006815e63/src/types.jl#L134-L143 according the the `show` docs. Only the three-argument show methods should have newlines.

PR #970, opened 2022-01-20
 fix(show)!: produce executable string for Options

gwater:
Re-implement two-argument Base.show() such that it produces an executable string of Optim.Options() as specified in Base docs [1].

Move previous implementation to three-argument Base.show() which should be human-readable [2].

[1] https://docs.julialang.org/en/v1/base/io-network/#Base.show-Tuple{IO,%20Any}
[2] https://docs.julialang.org/en/v1/base/io-network/#Base.show-Tuple{IO,%20Any,%20Any}

Refs: #915

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -131,7 +131,19 @@ function Options(;
         Int(show_every), callback, Float64(time_limit))
 end
 
+_show_helper(output, k, v) = output * "$k = $v, "
+_show_helper(output, k, ::Nothing) = output
+
 function Base.show(io::IO, o::Optim.Options)
+    content = foldl(fieldnames(typeof(o)), init = "Optim.Options(") do output, k
+        v = getfield(o, k)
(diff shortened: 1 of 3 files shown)


gwater:
I dont really understand why the coverage is reduced. As you can see I added a test case for the three-argument show() method.

gwater:
The problem seems to be that `test/general/types.jl` is not actually executed during testing https://github.com/JuliaNLSolvers/Optim.jl/blob/cdebfcacb58ad52256342d8e8e8754b1bb92dcde/test/runtests.jl#L35

pkofod:
> The problem seems to be that `test/general/types.jl` is not actually executed during testing

that's stupid, thanks for catching this

> Re-implement two-argument Base.show() such that it produces an executable string of Optim.Options() as specified in Base docs [1].

Thanks!

PR #972, opened 2022-01-27
fix BFGS loop order

vpuri3:
MWE: https://github.com/JuliaNLSolvers/Optim.jl/commit/331a26fd76011b1e4598fc622a86c4cc6797307f#commitcomment-64720042
```julia
using LinearAlgebra, BenchmarkTools

n = 200
A = rand(n,n)
x = rand(n)
u = rand(n)
c1= 1.0
c2= 1.0


function f1!(A,x,u,c1,c2)
  for i in 1:n
    @simd for j in 1:n
      A[i,j] += c1 * x[i] * x[j]
      A[i,j] -= c2 * u[i] * x[j]
      A[i,j] -= c2 * x[i] * u[j]
    end
  end
  return
end

function f2!(A,x,u,c1,c2)
  for i in 1:n
    @simd for j in 1:n
      @inbounds A[i, j] += c1 * x[i] * x[j]' - c2 * (u[i] * x[j]' + u[j]' * x[i])
    end
  end
  return
end 
  
function f3!(A,x,u,c1,c2)
  mul!(A,vec(x),vec(x)', c1,1.0)                  
  mul!(A,vec(u),vec(x)',-c2,1.0)
  mul!(A,vec(x),vec(u)',-c2,1.0)
  return
end
f1!(A,x,u,c1,c2)
f2!(A,x,u,c1,c2)
f3!(A,x,u,c1,c2)

@btime f1!($A,$x,$u,$c1,$c2)
@btime f2!($A,$x,$u,$c1,$c2)
@btime f3!($A,$x,$u,$c1,$c2)
```
```julia
julia> @run t
  19.693 ms (720601 allocations: 11.00 MiB)
  18.030 ms (640601 allocations: 9.78 MiB)
  57.246 μs (0 allocations: 0 bytes)
```

### DIFF
--- a/src/multivariate/solvers/first_order/bfgs.jl
+++ b/src/multivariate/solvers/first_order/bfgs.jl
@@ -160,9 +160,19 @@ function update_h!(d, state, method::BFGS)
         c2 = 1 / dx_dg
 
         # invH = invH + c1 * (s * s') - c2 * (u * s' + s * u')
-        mul!(state.invH,vec(state.dx),vec(state.dx)', c1,1)
-        mul!(state.invH,vec(state.u ),vec(state.dx)',-c2,1)
-        mul!(state.invH,vec(state.dx),vec(state.u )',-c2,1)
+        if(state.invH isa Array) # i.e. not a CuArray
+            for j in 1:n
+                @simd for i in 1:n
+                    @inbounds state.invH[i, j] += c1 * state.dx[i] * state.dx[j]' -
+                                                  c2 * (state.u[i] * state.dx[j]' +
+                                                        state.u[j]' * state.dx[i])
+                end
+            end
+        else
+            mul!(state.invH,vec(state.dx),vec(state.dx)', c1,1)
+            mul!(state.invH,vec(state.u ),vec(state.dx)',-c2,1)
+            mul!(state.invH,vec(state.dx),vec(state.u )',-c2,1)
+        end
     end
 end
 
--- a/src/multivariate/solvers/first_order/bfgs.jl
+++ b/src/multivariate/solvers/first_order/bfgs.jl
@@ -160,19 +160,9 @@ function update_h!(d, state, method::BFGS)
         c2 = 1 / dx_dg
 
         # invH = invH + c1 * (s * s') - c2 * (u * s' + s * u')
-        if(state.invH isa Array) # i.e. not a CuArray
-            for i in 1:n
-                @simd for j in 1:n
-                    @inbounds state.invH[i, j] += c1 * state.dx[i] * state.dx[j]' -
-                                                  c2 * (state.u[i] * state.dx[j]' +
-                                                        state.u[j]' * state.dx[i])
-                end
-            end
-        else
-            mul!(state.invH,vec(state.dx),vec(state.dx)', c1,1)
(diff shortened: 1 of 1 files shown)


vpuri3:
error is in the `arraypartition input` test

vpuri3:
```julia
using RecursiveArrayTools
H = [1.0 0.0 0.0; 0.0 1.0 0.0; 0.0 0.0 1.0]                                                            
x = ArrayPartition{Float64, Tuple{Vector{Float64}, Vector{Float64}}}(([0.00014378264534851268], [0.006636
295688791828, 35.74003500398774]))                                                                              
u = ArrayPartition{Float64, Tuple{Vector{Float64}, Vector{Float64}}}(([0.00028756529069795533], [2.9764724
291374023, 3.4695914310129094e6]))                                                                              

mul!(H, u,x') # errors
```

vpuri3:
somewhere in `LinearAlgebra/transpose.jl`, it is accessing `x[3,1]`
```julia
julia> checkbounds(Bool,u,3,1)
true

julia> u[3,1]
ERROR: BoundsError: attempt to access Tuple{Vector{Float64}, Vector{Float64}} at index [3]
Stacktrace:
 [1] getindex(A::ArrayPartition{Float64, Tuple{Vector{Float64}, Vector{Float64}}}, i::Int64, j::Int64)
   @ RecursiveArrayTools ~/.julia/dev/RecursiveArrayTools.jl/src/array_partition.jl:151
```

pkofod:
the `simd` way is probably not optimal. That said, it all probably depends on the size of the problem, wouldn't you think? @chriselrod

chriselrod:
> the `simd` way is probably not optimal. That said, it all probably depends on the size of the problem, wouldn't you think? @chriselrod

Loops are in the wrong order, and use the global variable `n`.

vpuri3:
Thanks @chriselrod , i had no idea global variables could have such an impact. it's really cool to see native julia code run so fast!

chriselrod:
Still doesn't quite match the `@turbo` version, but getting there:
```julia
julia> function f2muladd!(A,x,u,c1,c2,n)
         @inbounds for j in 1:n
           c1xj = c1 * x[j]'
           c2xj = c2 * x[j]'
           c2uj = c2 * u[j]'
           for i in 1:n
             A[i, j] = muladd(x[i], c1xj, muladd(-u[i], c2xj, muladd(c2uj, x[i], A[i, j])))
           end
         end
         return
       end
f2muladd! (generic function with 1 method)

julia> @btime f2muladd!($A,$x,$u,$c1,$c2,$n)
  5.614 μs (0 allocations: 0 bytes)
```
I suggest going with the explicit `muladd`s and hoisting.

pkofod:
@vpuri3 let me know how you want to proceed here. Do you just want to do the loop re-ordering, or do you want to include Chris' suggestions?

vpuri3:
thanks @chriselrod , @pkofod . im unavailable for a few days due to travel. ill just do loop reordering for now, and leave it open for later improvements.

vpuri3:
running CI. weird that tests are only failing on ubuntu

pkofod:
Thanks for noticing this. I've merged Chris' PR

Issue #971, opened 2022-01-21
Interior-point Newton seems to be non-deterministic


ForceBru:
# Code

```julia
import Pkg
Pkg.activate(temp=true, io=devnull)
Pkg.add(name="Optim", version="1.6.0", io=devnull)
Pkg.status()

const AV = AbstractVector{T} where T
using DelimitedFiles, Optim

normal_pdf(x::Real, μ::Real, var::Real) =
    exp(-(x-μ)^2 / (2var)) / sqrt(2π * var)

function D(θ::AV{<:Real}, b::Real, r::AV{<:Real})
	@assert length(θ) % 3 == 0

	N = length(r)
	K = length(θ) ÷ 3
	α, μ, σ = θ[1:K], θ[K+1:2K], θ[2K+1:end]
	var = σ .^2

	sum(
		α[k] * (
			sum(
				α[h] * normal_pdf(μ[k], μ[h], 2b + var[k] + var[h])
				for h ∈ 1:K
			)
			- 2*sum(
				1/N * normal_pdf(μ[k], r[j], 2b + var[k])
				for j ∈ eachindex(r)
			)
		)
		for k ∈ 1:K
	)
end

data = open("sample_data.csv", "r") do io
	DelimitedFiles.readdlm(io, ',')
end
data = data[:, 1]

result = let
	θ0 = Float64[1/3,1/3,1/3, 0,0,0, 1,1,1]
	obj(θ::AV{<:Real}) = D(θ, 1.0, data)

	con_c!(c, θ::AV{<:Real}) = (c[1] = sum(θ[1:(length(θ)÷3)]) - 1; c)
	con_jac!(J, θ::AV{<:Real}) = (J[1, 1:(length(θ)÷3)] .= 1; J)
	con_h!(H, θ, λ) = H

	lx = Float64[0,0,0, -Inf,-Inf,-Inf, 0,  0,  0]
	ux = Float64[1,1,1,  Inf, Inf, Inf, Inf,Inf,Inf]
	lc = [0.0]; uc = [0.0]
	dfc = TwiceDifferentiableConstraints(
		con_c!, con_jac!, con_h!,
		lx, ux, lc, uc
	)
	optimize(
		obj, dfc, θ0, IPNewton(),
		Optim.Options(store_trace=true, show_trace=true, show_every=1, iterations=10),
		autodiff=:forward
	)
end

@show Optim.minimizer(result)
```

[...]

PR #973, opened 2022-02-02
fix(ipnewton): NaN instead of undef allocation

gwater:
Prevent non-deterministic behavior in interior point Newton method

Ref: #971


johnmyleswhite:
LGTM, but others should approve

pkofod:
Thanks, but let me just see the original issue. I think the appropriate thing here is to fill with NaNs because we expect users to take responsibility of all elements of the partial derivatives. You are right that this may be something we should emphasize more in the docs, but I only skimmed the issue, let me read it in detail.

src/multivariate/solvers/constrained/ipnewton/ipnewton.jl
```diff
@@ -108,7 +108,7 @@ end
 function initial_state(method::IPNewton, options, d::TwiceDifferentiable, constraints::TwiceDifferentiableConstraints, initial_x::AbstractArray{T}) where T
     # Check feasibility of the initial state
     mc = nconstraints(constraints)
-    constr_c = Array{T}(undef, mc)
+    constr_c = zeros(T, mc)
```
pkofod, line 111: I would prefer this to be `fill(T(NaN), mc)`. The reason is that the user would immediately realize if they forgot to set all the elements they want to set. If it starts as zero, you might not realize that you're not setting element number 4 ever, because the 0 will not ruin the calculations.

src/multivariate/solvers/constrained/ipnewton/ipnewton.jl
```diff
@@ -124,13 +124,13 @@ function initial_state(method::IPNewton, options, d::TwiceDifferentiable, constr
     f_x_previous = NaN
     f_x, g_x = value_gradient!(d, initial_x)
     g .= g_x # needs to be a separate copy of g_x
-    H = Matrix{T}(undef, n, n)
-    Hd = Vector{Int8}(undef, n)
+    H = zeros(T, n, n)
```
pkofod, line 127: Same here

pkofod (changes requested):

src/multivariate/solvers/constrained/ipnewton/ipnewton.jl
gwater: ok, i think the best solution here is to use `collect` for H rather than splitting allocation and initialization.

src/multivariate/solvers/constrained/ipnewton/ipnewton.jl
gwater: agreed

gwater:
It looks like this test problem does not fill all components of the constraint Jacobian https://github.com/JuliaNLSolvers/OptimTestProblems.jl/blob/45650b6fbbfb07892f827dd74aee03f5e32afc20/src/optim_tests/multivariate/constrained.jl#L81

gwater:
Basically, all tests fail now because the test case actually leaves some fields of the Jacobian buffer unset. I've opened a PR to fix the test case https://github.com/JuliaNLSolvers/OptimTestProblems.jl/pull/20

gwater:
as far as i can tell, its still checking with the old test cases

pkofod:
Thanks!

Issue #609, opened 2018-05-29
IPNewton / ConstrainedOptim TODOs


anriseth:
*This is just copied from https://github.com/JuliaNLSolvers/ConstrainedOptim.jl/issues/4*

Here's a list of things that I believe we should focus on next. If anyone is interested in helping out here that would be very much appreciated.

**Interface**

- [ ] Enable `value!` et al. for the constraints in NLSolversBase **top priority**
- [ ] Add a quasi-Newton algorithm
- [ ] ~Update MathProgBase interface~ Add `MathOptInterface.jl` functionality
- [ ] Add a Newton-Krylov scheme

**Interior / IPNewton**

- [ ] Support sparse Hessian and sparse constraint Jacobian **top priority**
- [ ] Figure out how to enable the autodiff-tests in `constraints.jl` again
- [ ] Move "show_linesearch`  to Optim.options
- [ ] Rethink how to deal with `allow_f_increases` for constrained problems (it should "always" be allowed)
- [ ] There are lots of functions from Tim's PR that are not used anywhere. Figure out what to delete.

**Linesearch**

- [ ] Figure out how Tim inteded to use `backtrack_constrained` (only `backtrack_constrained_grad` works at the moment)
- [ ] Move line searches to LineSearches.jl (after we change the interface in LineSearches)

**Testing**

- [ ] Add more constrained optimization tests to OptimTestProblems
- [ ] Update OptimTests to work with ConstrainedOptim (We need sparse Hessian + Jacobian support before going all in on CUTEst problems)
- [ ] Compare against Ipopt

kkmann:
Hey, just wanted to bump this since I just recently used the constrained optimization functionality in Optim.jl - thanks a lot for implementing this! 
I wanted to use ForwardDiff and followed the IPNewton tutorial as close as possible. I.e. I implemented all gradients, Jacobians, and Hessians myself via ForwardDiff - is there a similar functionality like `autodiff = :forward` for non-constrained problems (if not, that would be cool ;))? Also, I had a hard time figuring out that the mu parameter is quite important - had to chose it manually to get a solution that respected the constraints.

anriseth:
Happy to hear that you've used this:)

There is an automatic differentiation convenience call for first order, but not for second order. See https://github.com/JuliaNLSolvers/NLSolversBase.jl/blob/master/src/objective_types/constraints.jl#L122

[...]

PR #977, opened 2022-02-04
IPNewton() sparsity support

schrimpf:
This makes IPNewton() compatible with alternate types (e.g. sparse) of Hessians and Jacobians. It addresses one part of #609 

This is a work in progress. Some remaining issues:

- [ ]  The passing of array type for the Jacobian of the constraints is ugly. I made it an argument to IPNewton, but it would be better as a field of `TwiceDifferentiableConstraints`, just like `H` is a field of `TwiceDifferentiable`. That requires changing NLSolversBase
- [ ] Add en example of a sparse problem to the documentation
- [ ] Ensure IPNewton() benefits from sparsity. The changes make IPNewton() work with a sparse Hessian, but I haven't examined whether its appropriately preserved and exploited throughout the algorithm. Should also bench mark the performance versus dense matrices.
- [ ] Improve tests.  The only test is a simple equality constrained problem. It is not organized the same way as tests for other aspects of the package.


pkofod:
Appreciate it! Let me know when it's ready for review.

Vaibhavdixit02:
@schrimpf do you plan to continue this?

PR #978, opened 2022-02-07
use muladd and hoist some multiplications

chriselrod:
Given @vpuri3 won't be able to make the change I suggested here: https://github.com/JuliaNLSolvers/Optim.jl/pull/972#issuecomment-1027028046 for a few days:

> thanks @chriselrod , @pkofod . im unavailable for a few days due to travel. ill just do loop reordering for now, and leave it open for later improvements.

I figured I'd make it here.


pkofod:
thanks

PR #1044, opened 2023-06-16
Math rendering in the README

DanielVandH:
Since LaTeX can now be used in the README, I've changed some of the math in the README. This also helps with the rendering of the Rosenbrock function, which on a dark theme is unreadable:

image


pkofod:
Appreciate it!

PR #1049, opened 2023-08-29
Only warn for early stop if `options.show_trace`

MilesCranmer:
Addresses discussion in https://github.com/JuliaNLSolvers/Optim.jl/pull/1046 with @mohamed82008 

Users of PySR and SymbolicRegression.jl were suddenly getting spammed with warning messages (e.g., https://github.com/MilesCranmer/PySR/discussions/416) so I had to hotfix things to use an earlier Optim.jl.

This PR addresses this so that the warning only shows up if the user requested that level of information with `show_trace = true`. Otherwise the warning does not show. But it will still quit early which I agree is the correct behavior.

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -95,11 +95,11 @@ function optimize(d::D, initial_x::Tx, method::M,
         end
 
         if g_calls(d) > 0 && !all(isfinite, gradient(d))
-            @warn "Terminated early due to NaN in gradient."
+            options.show_trace && @warn "Terminated early due to NaN in gradient."
             break
         end
         if h_calls(d) > 0 && !(d isa TwiceDifferentiableHV) && !all(isfinite, hessian(d))
-            @warn "Terminated early due to NaN in Hessian."
+            options.show_trace && @warn "Terminated early due to NaN in Hessian."
             break
         end
     end # while


mohamed82008 (approved):
The decision is Patrick's but just showing my support.

MilesCranmer:
Pinging @pkofod

pkofod:
I'm wondering if this is the right approach. If you have `optimize` in a loop where you don't want logging information, have you considered using `with_logger(NullLogger) do .... end` ?

MilesCranmer:
I don’t want to skip all logging information, just the NaNs in the Optim loop. I want all logging info from any user-specified objective, or other optimization warnings, to be displayed. 

Throwing a warning for NaNs in all optimization loops is a subjective design decision. For some applications where NaNs are frequently observed (like my symbolic regression use cases), it results in a bad user experience with tons of irrelevant warnings being printed. So I’d like to be able to turn it off.

pkofod:
Okay. I personally think the options struct is way overloaded as it is, but maybe it's better to introduce a `show_warnings` keyword here instead? I'm not sure I agree the two are similar in nature.

MilesCranmer:
Sounds good. I didn't want to add yet another option which is why I included it in `show_trace`. But happy to change it.

MilesCranmer:
Let me know what your decision is?

MilesCranmer:
Ping @pkofod. What do you prefer?

MilesCranmer:
Ping @pkofod. This is holding up SymbolicRegression.jl and PySR from upgrading.

MilesCranmer:
Okay I still haven't heard any response so I'm just going to submit a second PR with `show_warnings` instead and you can pick the PR you want.

MilesCranmer:
Ping @mohamed82008 @pkofod. Sorry for being annoying but I've been waiting for this since August and it is still blocking me from upgrading.

pkofod:
went with #1058

PR #1051, opened 2023-10-05
restore allocation free maxdiff

annuges:
The maxdiff function used in assess_convergence allocates which causes a non negligible time increase in a large-ish LBFGS problem for me.

#889 switched maxdiff to a mapreduce based implementation to support gpu arrays.
Unfortunately the mapreduce now allocates for standard arrays.

This change restores the previous allocation free code for normal Arrays while still allowing gpu arrays using mapreduce


pkofod (approved):

pkofod:
thanks

PR #1052, opened 2023-10-08
Update to new MOI nonlinear interface

blegat:
JuMP v1.15 now supports providing the nonlinear objective and constraints with `@objective` and `@constraint`.
The old syntax `@NLobjective` and `@NLconstraint` is kept for backward compatibility only.
Since the MOI wrapper in Optim is new, we don't need to support the old interface so this PR replace it the new one which works with `@objective` and `@constraint`.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -255,11 +255,15 @@ function MOI.optimize!(model::Optimizer{T}) where {T}
     method = model.method
     nl_constrained = !isempty(nlp_data.constraint_bounds)
     features = MOI.features_available(evaluator)
+    has_bounds = any(vi -> isfinite(model.variables.lower[vi.value]) || isfinite(model.variables.upper[vi.value]), vars)
     if method === nothing
         if nl_constrained
             method = IPNewton()
         elseif :Grad in features
-            if :Hess in features
+            # FIXME `fallback_method(f, g!, h!)` returns `Newton` but if there
+            # are variable bounds, `Newton` is not supported. On the other hand,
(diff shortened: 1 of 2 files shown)


pkofod:
lgtm

blegat:
Should be ready from my side as well

pkofod:
I'll review it, thanks

pkofod:
Okay, fairly small change compared to the original. LGTM

odow:
This PR did not appropriately update the compat in Project.toml: https://github.com/JuliaNLSolvers/Optim.jl/pull/1060

ccoffrin:
@blegat, is there a way this can support both the old and the new interface?

blegat:
It's not too hard, what's the motivation ?

ccoffrin:
Legacy code may not yet be updated to the new NLP interface.

Now that I think of it, I think it would be ok if there was one version that worked with the old NLP interface and another version that only supported the new one.

When I tried to use the old NLP interface with the current tagged release of Optim it did not seem to work for me.

blegat:
Yes, the addition of the old NLP was merged a few months ago so after the latest release. I thought that since it's a new feature there is no need to maintain support for the old interface

ccoffrin:
Ya I agree no need to maintain support for the old interface, but can we make sure there is a release where the old interface works?

blegat:
The issue is that v1 has already been tagged so if we release a version with the old interface and then remove it, it's a breaking change

ccoffrin:
Good point. I was very confused when I made my first comment on this.  I thought that in a previously release this package had been working with the legacy NL interface, but that is not the case.

I 100% agree going for the new NL interface, folks should be targeting that for new code development.

PR #1054, opened 2023-10-12
also show the minimizer

MariusDrulea:
Currently the optimize function does show a candidate solution, but only the minimum of the function is shown. We shall also show the minimizer of the function, such that the provided info is complete.

Looks like this now:

```
 * Status: success (objective increased between iterations)

 * Candidate solution
    Minimizer:     [214.63875045325074, 207.70250063762063, 214.6387496819325, 60.99374379542592, -1.6462211405820447e-10, -2.329416881307505e-9, -2.8537271470976752e-8]
    Final objective value:     8.456923e-13

 * Found with
    Algorithm:     L-BFGS

 * Convergence measures
    |x - x'|               = 2.49e-12 ≤ 1.0e-07
    |x - x'|/|x'|          = 1.16e-14 ≰ 0.0e+00
    |f(x) - f(x')|         = 8.45e-19 ≰ 0.0e+00
    |f(x) - f(x')|/|f(x')| = 9.99e-07 ≰ 0.0e+00
    |g(x)|                 = 6.38e-06 ≰ 1.0e-07

 * Work counters
    Iterations:    46
    f(x) calls:    228
    ∇f(x) calls:   228
```

### DIFF
--- a/docs/src/user/config.md
+++ b/docs/src/user/config.md
@@ -41,7 +41,7 @@ In addition to the solver, you can alter the behavior of the Optim package by us
 * `x_tol`: Absolute tolerance in changes of the input vector `x`, in infinity norm. Defaults to `0.0`.
 * `f_tol`: Relative tolerance in changes of the objective value. Defaults to `0.0`.
 * `g_tol`: Absolute tolerance in the gradient, in infinity norm. Defaults to `1e-8`. For gradient free methods, this will control the main convergence tolerance, which is solver specific.
-* `f_calls_limit`: A soft upper limit on the number of objective calls. Defaults to `0` (unlimited). Does not count the objective calls in gradient and hessian estimation.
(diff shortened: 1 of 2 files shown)


annuges:
Just another user here so someone else will need to make that call but I don’t think printing the minimizer like that is a good idea.

While it might be fine for smaller problems dumping the entire Parameter vector of say a +100k parameter model into the console would be completely unusable.

pkofod:
We used to have this, and I must admit that I removed it because we had to default to showing at max 5 values or something like that, and at that point it gets quite useless for many real use cases. An alternative may of course be to show it in vertical alignment, but I think it's better for the user to just print `Optim.minimizer(res)` after the optimization run. I'm open to counter arguments of course. You could show it in more of a `n x 5` grid that gets wrapped such that you only show 5 columns, but still, if you have hundreds of thousands of variables it will kill your terminal.

pkofod:
You're welcome to open this if you want to argue the case that it should be shown.

PR #1060, opened 2023-12-12
Fix compat bounds in Project.toml

odow:
#1052 needed to update the lower bound of MOI.

In addition, since that version MOI supports only v1.6, Optim now supports only 1.6 and later also.

Also added compat bounds for standard libraries, since these are necessary for auto-merge in the registry.

The `<0.0.1` is needed for any packages on 1.6 that do package operations in the test suite.


pkofod:
Thanks

PR #1073, opened 2024-01-29
Start in interior of variable bounds

blegat:
As highlighted in https://github.com/JuliaNLSolvers/Optim.jl/issues/1071, most tests were failing because the starting point was infeasible.
With this PR, most are now passing.
The remaining ones to investigate are:
* `test_nonlinear_expression_hs071` : this one gives an infeasible solution, not sure if it's due to the fact that the starting values does not satisfy the constraints, no warning is thrown though.
* `test_nonlinear_expression_hs071_epigraph` : the variables are in the interior of the bounds but some of the constraints are not satisfied hence the warning is thrown. Then it converges to an incorrect solution
* `test_objective_FEASIBILITY_SENSE_clears_objective` : this one seems it may be a bug in the MOI wrapper, let me investigate first

### DIFF
--- a/ext/OptimMOIExt.jl
+++ b/ext/OptimMOIExt.jl
@@ -182,11 +182,32 @@ function MOI.add_constraint(
 end
 
 function starting_value(optimizer::Optimizer{T}, i) where {T}
-    if optimizer.starting_values[i] !== nothing
-        return optimizer.starting_values[i]
+    start = optimizer.starting_values[i]
+    v = optimizer.variables
+    if isfinite(v.lower[i])
+        if isfinite(v.upper[i])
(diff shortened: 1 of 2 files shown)


blegat:
I suggesting investigating the remaining failing tests in separate PRs

blegat:
@pkofod This PR is good to go from my end

pkofod:
> @pkofod This PR is good to go from my end

thanks I will look. Busy feb and march for me 😓

blegat:
Thanks, I just rebased to fix conflicts

pkofod:
Wow, I had totally forgotten about this. Thanks

PR #1080, opened 2024-02-21
Fixing Nelder-Mead in box optimization

lalondeo:
This fixes the issue raised in https://github.com/JuliaNLSolvers/Optim.jl/issues/912. When Nelder-Mead's initial state is created, some of the points in the simplex lie outside of the box, so that the value of the corresponding logarithm is infinite. Since the barrier parameter is initially zero, this results in the objective's value at those points being zero * Inf = NaN, which trips up the optimizer. Starting out with a small but nonzero barrier parameter instead fixes the issue (resulting in the objective being infinite outside the box) and the code snippets that were given in the aforementioned post now give the correct results. However, this is more of a band-aid than anything else, as it seems reasonable to think that the points in Nelder-Mead's simplex should always satisfy the box constraints. Enforcing this would require changing large portions of the code.


pkofod:
Thank you, I will have a look.

EdsterG:
+1 for this to be fixed.

PR #1088, opened 2024-03-27
Fix precompilation issue

### LINKED ISSUE #1087: `eval` in OptimMOIExt fails to precompile
```
fatal: error thrown and no exception handler available.
InitError(mod=:OptimMOIExt, error=ErrorException("Evaluation into the closed module `Optim` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `Optim` with `eval` during precompilation - don't do this."))
```

MWE:
With a `optim-test/OptimTest.jl` file that has:
```julia

module OptimTest

using MathOptInterface
using Optim

end
```
```julia
pushfirst!(LOAD_PATH, "optim-test")
using OptimTest
```

devmotion:
Fixes #1087 (basically in the same way as the implementation in NLOpt, the tiny difference is just that the check in the `__init__` function in the extension is also based on `isdefined(Base, :get_extension)` to ensure there are no cases when `Optim.Optimizer` is not defined).

This reintroduces MOI as a regular dependency on Julia < 1.9, but IMO this is fine since on Optim <= 1.9.2 MOI was a regular dependency already. On Julia >= 1.9, this PR actually reduces the number of dependencies since PackageExtensionsCompat is removed.


pkofod:
thanks David!

PR #1089, opened 2024-03-28
 Fix type of initial_scale in reset_search_direction!

odow:
x-ref https://discourse.julialang.org/t/error-loaderror-methoderror-no-method-matching-init-identity-matrix-componentvector-float32-vector-float32-tuple-axis-float64/112211

I don't really know the implications of this, but I assume this is the fix.

### DIFF
--- a/src/utilities/perform_linesearch.jl
+++ b/src/utilities/perform_linesearch.jl
@@ -15,7 +15,7 @@ function reset_search_direction!(state, d, method::BFGS)
         if method.initial_stepnorm === nothing
             state.invH .= _init_identity_matrix(state.x)
         else
-            initial_scale = method.initial_stepnorm * inv(norm(gradient(d), Inf))
+            initial_scale = T(method.initial_stepnorm) * inv(norm(gradient(d), Inf))
             state.invH .= _init_identity_matrix(state.x, initial_scale)
         end
     else


devmotion:
I think it might be more consistent to apply the pattern in https://github.com/JuliaNLSolvers/Optim.jl/blob/7cc8328a590a0c815e23005321459712382b5b42/src/multivariate/solvers/first_order/bfgs.jl#L81 to https://github.com/JuliaNLSolvers/Optim.jl/blob/7cc8328a590a0c815e23005321459712382b5b42/src/utilities/perform_linesearch.jl#L18. Generally, isn't the error caused by an incorrectly typed `initial_stepnorm` provided by the user? If you deal with `Float32` data, you can avoid the problem by specifying also `initial_stepnorm` as a `Float32`. That being said, I guess https://github.com/JuliaNLSolvers/Optim.jl/blob/7cc8328a590a0c815e23005321459712382b5b42/src/multivariate/solvers/first_order/bfgs.jl#L81 makes it a bit more convenient for users.

pkofod:
> I think it might be more consistent to apply the pattern in

I agree

pkofod:
> Generally, isn't the error caused by an incorrectly typed `initial_stepnorm` provided by the user? If you deal with `Float32` data, you can avoid the problem by specifying also `initial_stepnorm` as a `Float32`. That being said, I guess

Sure, but usually I think it should be enough to supply x0 as the type you want even if there might be some cases where it's harder to prepare for that.. but I think we can apply the "correct" pattern here for the user..

pkofod:
@odow would you like to update this PR? Or would you rather not do more here. I think @devmotion 's assessment is correct here.

odow:
Updated. I don't have a test or a reproducible example; I was just following the stack trace from the discourse post.

pkofod:
> Updated. I don't have a test or a reproducible example; I was just following the stack trace from the discourse post.

okay, thanks. Yes it can be a bit hard here to have a user-facing example (from the top level call to optimize) because you need it to hit a reset... I will have a look again, maybe I can write a test than artificially calls internals as if it had iterated and had to reset

PR #1090, opened 2024-03-28
Run CI also on oldest compatible Julia version and update GH actions

devmotion:
I noticed that currently CI only tests the latest stable Julia release and nightly. To avoid accidentally breaking support for older, officially declared to be compatible, Julia versions I think it would be good to run CI test also with the oldest compatible Julia version (currently 1.6).

I used the opportunity and also updated and simplified the GH actions.


pkofod:
Thanks. Let's see if it works :)

Issue #1104, opened 2024-10-24
Issues affecting changes in line search methods


timholy:
In an attempt to fix https://github.com/JuliaNLSolvers/LineSearches.jl/issues/173 and https://github.com/JuliaNLSolvers/LineSearches.jl/issues/175 (see https://github.com/JuliaNLSolvers/LineSearches.jl/pull/180), I've concluded that the best option is to rewrite `HagerZhang` from scratch. I've actually done this, and the new implementation passes all the tests in LineSearches.jl.

However, many tests in Optim fail. Here's a to-be-updated list of issues that need discussion and/or resolution:

- [X] some tests depend on every minor detail of the algorithms. In [this case where we deliberately limit the number of iterations](https://github.com/JuliaNLSolvers/Optim.jl/blob/658c39a316bda911cddf442702b11bf086f1b5ba/test/general/api.jl#L116-L129), the new implementation gets to a lower minimum with fewer function calls, and yet because of how that test is written it's counted as a failure. I've already addressed this one (it's easy).
- [ ] there are a *ton* of tests that have bespoke exceptions granted with no obvious rhyme or reason. One example: https://github.com/JuliaNLSolvers/Optim.jl/blob/658c39a316bda911cddf442702b11bf086f1b5ba/test/multivariate/solvers/first_order/accelerated_gradient_descent.jl#L14-L32 The `test/multivariate` directory is full of examples like these. Since `HagerZhang` is the default line search algorithm, any subtle change can cause shifts in the things that require exceptions. I will wager that something things that previously failed will pass, and I know that some things that previously passed will fail.
  **Proposed solutions**:
  + echoing #1102, one option is that the convergence criteria should be part of the problem formulation, stored as a field in [`OptimizationProblem`](https://github.com/JuliaNLSolvers/OptimTestProblems.jl/blob/c1fba66c90b44934d13cd11b2502573f1c11fab8/src/optim_tests/multivariate/MultivariateProblems.jl#L19). Or at least, something that would allow you to compute problem-specific convergence criteria.
  + alternatively, perhaps we could compute the criteria from the Hessian evaluated at the minimum, if all the problems are in fact second-order differentiable. This doesn't adequately incorporate roundoff error (which can have a big impact), so it may not actually work in practice. Perhaps we could do something clever with IntervalArithmetic to estimate the impact of roundoff error, though again there are caveats.

[...]

PR #1105, opened 2024-10-25
Use `@test_broken` instead of skipping

timholy:
A global flag, `test_broken = false`, determines whether this runs during CI.
Locally it's useful to be able to tally the failures, e.g., #1104

This also fixes one failure in L-BFGS that causes CI to fail.


timholy:
xref #1104

timholy:
Green! Yay!

Issue #1102, opened 2024-09-25
warn people that g_tol often needs to be set


stevengj:
You set an absolute tolerance of `g_tol = 1e-8` by default.   I don't know how you picked this — it depends on the scaling of the problem, because **gradients are dimensionful**.

This implicitly assumes that the function is scaled so that the **typical derivatives are of order unity**.      This makes it pretty dangerous to use the default tolerances, in my opinion.

A safer default would have been to set a relative tolerance `g_reltol` on the gradient, e.g. by setting `g_tol = g_reltol * initial_g`.

At the very least, the manual should highlight this implicit scaling assumption, and let people know that they are likely to need a problem-specific `g_tol`.

timholy:
Agreed (and I didn't choose those settings). My favorite criterion is `abs(f(x+Δx) - f(x)) < f_rel_tol * (abs(f(x + Δx)) + abs(f(x))`, as it is fully scale-invariant. Unfortunately, it can be unsatisfying for functions that achieve `f(zeros(d)) == 0` at the minimum: due to the compression of floating-point numbers near zero, it will iterate until you start getting into the subnormals. Short of an absolute tolerance on `f` (which is sensitive to the scale of `f`) or a tolerance on `x` or `g` (both of which are dimensionful), I'm not aware of a good solution.

stevengj:
Yeah, in practice many people don't realize that problem-specific tuning of tolerances is something you should almost always do.

If you're going to have a default, it should be relative tolerances, even though these can be insufficient upon occasion as well (or default to no tolerances, and throw an error if the user doesn't choose a stopping criterion).   Maybe you can't change this now for backwards-compatibility reasons.

But in almost all of the manual's examples, no tolerance is passed, which is misleading.    This, at least, can be fixed.

pkofod:
Thank you for this discussion, and I agree with everything said so far. I don't remember the history of the default tolerance, and it doesn't matter if it was me or someone else.

I think this is actually a surprisingly hard truth about numerical optimization, that I also have difficulty handling in my profession life. I mean, when doing work myself, but also when communicating with customers.

Users and customers want to know if the procedure converged, but it's actually quite hard to say anything meaningfully about in general, because we have no natural scale to accept "0" at. Not converged? Just multiply the objective with 0.0001 and suddenly you "converged". 

[...]

PR #1107, opened 2024-10-27
Convergence criteria: improve docs, add per-component gradient

timholy:
This was inspired by #1102, and expands the discussion of convergence criteria in the documentation.

However, it also goes beyond that by optionally allowing the user to specify that the `g_abstol` criterion should change from
```julia
norm(g) <= g_abstol
```
to
```julia
all(abs(g[i]) <= g_abstol[i] for i = 1:n)
```
That is, it avoids taking the norm. As pointed out in #1102 with the scaling of the objective coefficient, it's also the case that scaling the coordinates (e.g., using meters vs kilometers) changes the scaling of the gradient components. Thus, there is an argument to be made that we need a theory of optimization over non-metric vector spaces. I'm not aware of such a theory existing, and basically all the algorithms in this package assume a metric at some place or another. But at least this allows useful convergence criteria to be specified in a scale-invariant manner.

This is fully-opt in: by default `g_abstol = 1e-8`, and that uses the previous criterion. But if you specify, e.g., `g_abstol = [1e-8, 1e-6]` (for an optimization problem with two variables), then you use the new one.

CC @stevengj

### DIFF
--- a/docs/src/user/config.md
+++ b/docs/src/user/config.md
@@ -40,7 +40,7 @@ In addition to the solver, you can alter the behavior of the Optim package by us
 
 * `x_tol`: Absolute tolerance in changes of the input vector `x`, in infinity norm. Defaults to `0.0`.
 * `f_tol`: Relative tolerance in changes of the objective value. Defaults to `0.0`.
-* `g_tol`: Absolute tolerance in the gradient, in infinity norm. Defaults to `1e-8`. For gradient free methods, this will control the main convergence tolerance, which is solver specific.
+* `g_tol`: Absolute tolerance in the gradient. If `g_tol` is a scalar (the default), convergence is achieved when `norm(g, Inf) ≤ g_tol`; if `g_tol` is supplied as a vector, then each component must satisfy `abs(g[i]) ≤ g_tol[i]`. Defaults to `1e-8`. For gradient-free methods (e.g., Nelder-Meade), this gets re-purposed to control the main convergence tolerance in a solver-specific manner.
 * `f_calls_limit`: A soft upper limit on the number of objective calls. Defaults to `0` (unlimited).
 * `g_calls_limit`: A soft upper limit on the number of gradient calls. Defaults to `0` (unlimited).
 * `h_calls_limit`: A soft upper limit on the number of Hessian calls. Defaults to `0` (unlimited).
--- a/docs/src/user/config.md
+++ b/docs/src/user/config.md
@@ -35,12 +35,15 @@ Special methods for bounded univariate optimization:
 * `Brent()`
 * `GoldenSection()`
 
-### General Options
+### [General Options](@id config-general)
+
 In addition to the solver, you can alter the behavior of the Optim package by using the following keywords:
 
-* `x_tol`: Absolute tolerance in changes of the input vector `x`, in infinity norm. Defaults to `0.0`.
-* `f_tol`: Relative tolerance in changes of the objective value. Defaults to `0.0`.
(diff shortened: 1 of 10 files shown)


pkofod:
I suppose this partially addresses @stevengj 's concern in-so-far it is now a bit more clearly described. I suppose the point about the relative tolerance is saved for later.

timholy:
I didn't do the relative tolerance because that would be a breaking change. But we could do that if you're OK with making a breaking change.

I agree that relative is better, but only if it's applied to each component individually (it's not scale-invariant otherwise). But that does have one weakness: what if some component of the gradient starts at 0? One could of course keep a running tally of the largest absolute value one has ever encountered.

stevengj:
There *are* algorithms that adaptively learn a per-component scaling (e.g. CCSA does this) and hence are more flexible about dimensionful variables.  But I agree that many optimization algorithms seem to implicitly assume that $x^T x$ is a reasonable norm, corresponding in dimensionful cases to an implicit weighted norm $x^T W x$ where the weight $W$ is a diagonal matrix of the inverses of the units of each component. 

However, I doubt that asking the user to provide even more information here (a per-component gradient tolerance) is so helpful (though it doesn't hurt to have the option) … they barely know how to set tolerances as it is.  Instead, I would recommend (1) rescaling the components of the unknowns so that they have comparable magnitudes, ideally of order unity and (2) then setting an appropriate gradient tolerance.

timholy:
> However, I doubt that asking the user to provide even more information here (a per-component gradient tolerance) is so helpful (though it doesn't hurt to have the option) … they barely know how to set tolerances as it is.

In some sense this was motivated by hoping to rationalize the test suite (#1104) as much as anything else. If you know where the minimum is, you can evaluate the Hessian at the minimum, and thus estimate how much roundoff error should contribute to a nonzero gradient. But there's a simpler way to achieve something similar: put a threshold on the actual value of `f`, i.e., if `f(x) <= thresh` then you've converged. That would be easy to set, might work robustly for the test suite, and may be more useful in practice than this PR. For example, if you're doing least-squares, you know that any point that makes `f(x)` "tiny" (by some user-specified definition of tiny) must be very close to a minimum.

pkofod:
> > However, I doubt that asking the user to provide even more information here (a per-component gradient tolerance) is so helpful (though it doesn't hurt to have the option) … they barely know how to set tolerances as it is.
> 
> In some sense this was motivated by hoping to rationalize the test suite (#1104) as much as anything else. If you know where the minimum is, you can evaluate the Hessian at the minimum, and thus estimate how much roundoff error should contribute to a nonzero gradient. But there's a simpler way to achieve something similar: put a threshold on the actual value of `f`, i.e., if `f(x) <= thresh` then you've converged. That would be easy to set, might work robustly for the test suite, and may be more useful in practice than this PR. For example, if you're doing least-squares, you know that any point that makes `f(x)` "tiny" (by some user-specified definition of tiny) must be very close to a minimum.



do you want to add the threshold in another PR?

Ickaser:
I, as a user, would be in favor of a per-component gradient, even if it's something that takes tinkering on my part to get right.

pkofod:
> I, as a user, would be in favor of a per-component gradient, even if it's something that takes tinkering on my part to get right.

I'm curious how you would know what to set the values at?

Ickaser:
> > I, as a user, would be in favor of a per-component gradient, even if it's something that takes tinkering on my part to get right.
> 
> I'm curious how you would know what to set the values at?

In short: trial and error. I am using a lot of this for, essentially, curve fitting with ODE parameters, which I do over and over for different sets of experimental data, which sometimes have noise or plant-model mismatch. So a per-component gradient would get set from experience of "yep, that's converged as much as we can expect"

PR #1113, opened 2024-10-29
Docs: adjust callback function explanation

### LINKED ISSUE #1111: Documentation about callback functions
As far as I can tell, a callback function given to `optimize` is supposed to return a boolean controlling whether the optimization is supposed to stop or not. Returning `nothing` leads to an error about using a non-boolean in a boolean context. Is this documented somewhere? All I could find was the [Configurable options](https://julianlsolvers.github.io/Optim.jl/stable/user/config/#Configurable-options) page saying returning `true` stops `optimize`, which seems incomplete to me.

I can happily make a PR if I indeed didn't miss anything to expand the documentation a bit.

tuncbkose:
Closes #1111, feedback is welcome.


tuncbkose:
Not sure why it was complaining about a merge conflict, but it should be ok now

docs/src/user/config.md
```diff
@@ -52,7 +52,7 @@ In addition to the solver, you can alter the behavior of the Optim package by us
...
 * `show_every`: Trace output is printed every `show_every`th iteration.
-* `callback`: A function to be called during tracing. A return value of `true` stops the `optimize` call. The callback function is called every `show_every`th iteration. If `store_trace` is false, the argument to the callback is of the type  [`OptimizationState`](https://github.com/JuliaNLSolvers/Optim.jl/blob/a1035134ca1f3ebe855f1cde034e32683178225a/src/types.jl#L155), describing the state of the current iteration. If `store_trace` is true, the argument is a list of all the states from the first iteration to the current.
+* `callback`: A function to be called during tracing. The return value should be a boolean, where `true` will stop the `optimize` call early. The callback function is called every `show_every`th iteration. If `store_trace` is false, the argument to the callback is of the type  [`OptimizationState`](https://github.com/JuliaNLSolvers/Optim.jl/blob/a1035134ca1f3ebe855f1cde034e32683178225a/src/types.jl#L155), describing the state of the current iteration. If `store_trace` is true, the argument is a list of all the states from the first iteration to the current.
```
pkofod, line 55:
```suggestion
* `callback`: A function to be called during tracing. The return value should be a boolean, where `true` will stop the `optimize` call early. The callback function is called every `show_every`th iteration. If `store_trace` is false, the argument to the callback is of the type  [`OptimizationState`](https://github.com/JuliaNLSolvers/Optim.jl/blob/a1035134ca1f3ebe855f1cde034e32683178225a/src/types.jl#L155), describing the state of the current iteration. If `store_trace` is true, the argument is a list of all the states from the first iteration to the current.
```

pkofod:
Thank you 👍

Issue #1124, opened 2025-01-18
Parallel ParticleSwarm?


sgaure:
I have a problem which benefits greatly by evaluating the function on the swarm in parallel. I have a local version of `Optim` which simply adds a `parallel` field in the `ParticleSwarm` struct. If `parallel` is set, I call a `compute_cost_parallel!` instead of `compute_cost`. 

```
function compute_cost!(f,
                       n_particles::Int,
                       X::Matrix,
                       score::Vector)

    for i in 1:n_particles
        score[i] = value(f, X[:, i])
    end
    nothing
end

function compute_cost_parallel!(f,
                       n_particles::Int,
                       X::Matrix,
                       score::Vector)

    Polyester.@batch for i in 1:n_particles
        score[i] = value(f, X[:, i])
    end
    nothing
end
```

This is all fine, though I haven't actually verified that the `value` call is thread safe. I might create a PR which includes this enhancement.  However, it's the wrong way. It should be up to the user how to parallelize. Some might use `Distributed`, some might use `OhMyThreads.jl`,  `@threads`, `@spawn`, or `MPI.jl`, and so on. So the loop should really be replaced by just a matrix call:

```
function compute_cost_parallel!(f, etc...)
    value!!!(score, f, X)
    return nothing
end
```
where the objective function `f` accepts a matrix `X`, and fills in the vector `score`.  Is this doable with the current API?

miguelmaso:
Hi everyone,

First of all, thanks to the maintainers for this package, and to @sgaure for pushing this forward.

I want to chim in because I am currently working on a high-performance optimization problem (calibration of parameters for coupled constitutive models) where every evaluation is very expensive. I’ve performed benchmarks comparing Optim.ParticleSwarm with other global optimization libraries (like Metaheuristics.jl).

Interestingly, Optim's implementation proved to be significantly in terms of memory allocations, execution time and accuracy. However, the current lack of a native batch evaluation (like proposed in #1125) is the only bottleneck preventing us from fully adopting Optim for large-scale simulations.

Having merged this contribution would be a major change for those of us working with "expensive" objective functions. Is there anything the community can do to help merging #1125, or any guidance on the preferred API to address this issue?

Thanks in advance

ChrisRackauckas:
[...]

PR #1125, opened 2025-01-20
Parallel ParticleSwarm

sgaure:
#1124 

This PR enhances `ParticleSwarm` with a new keyword `parallel::Bool = false`. If set to `true`, the objective function will be called with a vector to fill in, and a matrix of column vectors as inputs. Thus, the user may choose to evaluate the swarm in parallel with whatever means.


sgaure:
Btw, I forgot to mention. The commits also contain a change which allows some upper bounds to be equal to the lower bounds. This can be used as a simple way to keep some variables constant.

pkofod:
I see, yes I also implemented something similar in NLSolvers.jl. Let me review it. Thanks

sgaure:
The test failures seem to be unrelated. I came to think of it, should the new `parallel` keyword be called `batched` instead? It's not implementing anything parallel.

PR #1127, opened 2025-01-24
Fix default of `allow_f_increases` and `allow_outer_f_increases` in docs

JoshuaLampert:
Looking at the [source code](https://github.com/JuliaNLSolvers/Optim.jl/blob/b041bd63373cf2403180c6993eff235b5c010670/src/types.jl#L97) the default value for `allow_f_increases`  seems to be `true` and not `false` as indicated in the docs. Same for `allow_outer_f_increases`.
Looks like it was missed to update in #847.


pkofod:
Thank you!

PR #1128, opened 2025-03-05
Fix lists in nelder_mead.md

abhro:


pkofod:
Does your other PR make this one superfluous or should I just merge this first?

abhro:
I suppose the references edit is pointless now, with respect to #1130, but the main reason behind this PR was to fix the list in the constructor.

pkofod:
Let's just get that in then.

PR #1130, opened 2025-03-19
Use DocumenterCitations.jl

abhro:
Currently the references are enumerated at the end of doc pages, but as more references are added (as seen by empty headers on some other doc pages), it'll be useful to have a citation management tool with Documenter.jl.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,6 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 NLSolversBase = "d41bc354-129a-5804-8e4c-c37616107c6c"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -9,3 +9,6 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 Documenter = "1"
 Literate = "2"
+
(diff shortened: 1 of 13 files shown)


docs/src/algo/samin.md
```diff
@@ -107,3 +107,8 @@ Results of Optimization Algorithm
 ## References
```
pkofod, line 107: Should the old ones not go here then?

pkofod: I mean, be removed

abhro: Yes. I think I forgot to remove them after comparison. Thank you!

pkofod:
can you rebase this?

abhro:
Done. I've updated the doc-citations branch to reflect JuliaNLSolvers/Optim.jl's master branch

docs/Project.toml
```diff
@@ -9,3 +10,6 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 Documenter = "1"
 Literate = "2"
+
+[sources.Optim]
```
pkofod, line 14: I was looking for this [sources.X] feature for Project.tomls but couldn't find it. Can you tell me where it's documented? Thanks

abhro: The official documentation is [here](https://pkgdocs.julialang.org/v1/toml-files/#The-%5Bsources%5D-section). I first saw it [in Lux.jl which is maintaining a docs site](https://github.com/LuxDL/Lux.jl/blob/v1.10.0/docs/Project.toml#L33) for a monorepo.

abhro: This is mainly to simplify building docs locally. It's not always obvious that the main package needs to be `]dev`-d, and having this line just lets someone run `Pkg.instantiate()` on the docs environment and have everything set up

PR #1131, opened 2025-03-21
Support autodiff with DifferentiationInterface

gdalle:
Preliminary PR to run tests with the new version of NLSolversBase (which includes DI)

How should we document and test this more properly?


blegat:
The test failureS might be fixed by https://github.com/JuliaNLSolvers/Optim.jl/pull/1073

gdalle:
Yeah I don't think they're related to this PR

pkofod:
I merged 1073

pkofod:
> The test failures might be fixed by #1073

I did merge them, but I'm wondering if those fixes are related to the `No nonlinear objective.` errors. Could it be that the NDifferentiable are not populating all the fields they should @gdalle ?

pkofod:
or, @blegat looking at
```
 exclude = String[
            # No objective
            "test_attribute_SolveTimeSec",
            "test_attribute_RawStatusString",
```
were more tests added that we need to exclude?

blegat:
Yes, there might just be new tests that need to be excluded

pkofod:
> Yes, there might just be new tests that need to be excluded

one of them is this https://github.com/jump-dev/MathOptInterface.jl/blob/6bfca94b728e48a7800caf26bcc749fae86e0bc7/src/Test/test_nonlinear.jl#L1847-L1870 does that need to be excluded?

gdalle:
> Could it be that the NDifferentiable are not populating all the fields they should @gdalle ?

Looking at the NLSolversBase PR again, there is [this comment](https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/153/files#r1873491660) that never got an answer about `con_jac!` being ignored, but I don't think it affects the objective.
Otherwise I can't see how we would fail to initialize some fields without errors popping up earlier in the test suite.

gdalle:
I inserted #1133 into this PR

pkofod:
Seems like one is saying that a nonlinear objective is not present but I cannot make out if it should be there or not. The other one is an IPNewton that fails and I'm not sure if that could be related to the constraint function you're referring to above.

gdalle:
Do those checks also fail without the new NLSolversBase, eg if we upper bound its compat to before the DI integration?

pkofod:
Ill check later, good question !

gdalle:
Having a baseline of failing tests is not great for adding new features 😅 if they have been failing for a while perhaps they can be marked as `@test_broken`

pkofod:
Let's see what we get from #1133

pkofod:
Also failed there. Let's try https://github.com/JuliaNLSolvers/Optim.jl/pull/1134

gdalle:
Note that #1133 failed on Julia 1.6, where the DI-upgraded NLSolversBase cannot even be installed

pkofod:
You can try to resolve the conflicts due to the bounds I added on master to see what happens here now

gdalle:
@pkofod I merged master into this branch, let's see what blows up

pkofod:
Okay, seems to pass. codecov complains about an unhit method, but it seems to have been like that before as well.

PR #1134, opened 2025-03-23
Update MOI_wrapper.jl to exclude two failing tests

pkofod:


gdalle:
Perhaps constrain NLSolversBase to stick to the non-DI version in this PR, that way we can compare all of the latest changes in the next?

pkofod:
I thought you wanted to do it to be safe. The new version was never added on ci in this pr i think.

pkofod:
Also, on master they fail https://github.com/JuliaNLSolvers/Optim.jl/actions/runs/14013061756/job/39235222331  and that also pulls in v7.8.3 so without your additions, so I don't see how it could be related.

gdalle:
On master the CI pulls in 7.8.3 for Julia 1.6 and the new 7.9.0 for Julis 1.10.

gdalle:
Sorry for the confusion, basically I wanted to disentangle the changes in #1131 from the switch to NLSolversBase v7.9.0, which happens automatically for Julia 1.10 from now on even without further changes in Optim.

Issue #1093, opened 2024-04-26
Expose more optimisation parameters to users


rafaqz:
In simulated annealing the start temperature, rate of expansion, and acceptance ratio are all fixed constants defined in code. 

With expensive objective functions it can make a lot of sense to adjust these parameters, but currently that requires editing the code directly.

pkofod:
I agree 👍

pkofod:
Start temperature is controlled by the temperature input. For example `SimulatedAnnealing(; temperature=t->30/t)` gives you a temperature of 30 in the first iteration, then 15, then 7.5, .... The temperature controls the rate of cooling (is that what you call rate of expansion?) and the formulate for the acceptance probability depends on the temperature. Can you maybe expand on your thoughts? Is this maybe a documentation issue?

rafaqz:
Sorry I mean bounded, SAMIN. But really they should be the same in this regard.

In SAMIN the start temperature is just set to `2.0`:
https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/multivariate/solvers/constrained/samin.jl#L76

Acceptance thresholds are hard coded:
https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/multivariate/solvers/constrained/samin.jl#L218-L219

And the temperature is multiplied by 10 in the increase phase:
https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/multivariate/solvers/constrained/samin.jl#L312

pkofod:
I see, let me review that then :)

PR #1136, opened 2025-03-24
Expose SAMIN options

pkofod:
#1093


pkofod:
@rafaqz not sure if it's still relevant to you, but here goes an attempt :)

PR #1142, opened 2025-03-27
Implement EnumX termination codes

pkofod:
An alternative to calling all the individual functions to check why it stopped.

### DIFF
--- a/src/Optim.jl
+++ b/src/Optim.jl
@@ -127,8 +127,8 @@ export optimize, maximize, # main function
        ## Non-linear constraints
        IPNewton
 
-
 include("types.jl") # types used throughout
+include("termination_code.jl")
 include("Manifolds.jl") # code to handle manifold constraints
 include("multivariate/precon.jl") # preconditioning functionality
 
@@ -222,14 +222,8 @@ include("multivariate/solvers/constrained/ipnewton/utilities/trace.jl")
 # Maximization convenience wrapper
 include("maximize.jl")
 
-@static if !isdefined(Base, :get_extension)
-    include("../ext/OptimMOIExt.jl")
-    using .OptimMOIExt
-    const Optimizer = OptimMOIExt.Optimizer
-else
-    # declare this upfront so that the MathOptInterface extension can assign it
-    # without creating a new global
-    global Optimizer
-end
+# declare this upfront so that the MathOptInterface extension can assign it
+# without creating a new global
+global Optimizer
 
 end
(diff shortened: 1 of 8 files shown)


src/Optim.jl
```diff
@@ -25,9 +25,10 @@ using LineSearches           # for globalization strategy in Quasi-Newton algs
 
 import NaNMath               # for functions that ignore NaNs (no poisoning)
 
-import Parameters: @with_kw, # for types where constructors are simply defined
-                   @unpack   # by their default values, and simple unpacking
-                             # of fields
+import Parameters:
+    @with_kw, # for types where constructors are simply defined
```
devmotion, line 29: Unrelated comment: I assume you can just replace this with `@kwdef` nowadays

src/Optim.jl
```diff
@@ -25,9 +25,10 @@ using LineSearches           # for globalization strategy in Quasi-Newton algs
...
+    @unpack   # by their default values, and simple unpacking
```
devmotion, line 30: Unrelated comment: `Parameters.@unpack` (which is a reexported `UnPack.@unpack` IIRC) specializes on the names that are extracted (using `Val`s), which was a major source of compilation time in OrdinaryDiffEq at some point. I made SimpleUnPack that avoids this, but on Julia >= 1.7 (IIRC) you can just use the native destructuring syntax (`(; a, ...) = ...`) instead (which is what SimpleUnPack actually does for these newer Julia versions).

src/types.jl
```diff
@@ -170,9 +170,48 @@ end
...
+    "Hessian was not finite"
+    HessianNotFinite
+    "For algorithms where the TerminationCode is not yet implemented."
+    NotImplemented
+end
+
 abstract type OptimizationResults end
 
-mutable struct MultivariateOptimizationResults{O, Tx, Tc, Tf, M, Tls, Tsb} <: OptimizationResults
+mutable struct MultivariateOptimizationResults{O, Tx, Tc, Tf, M, Tls, Tsb,TE<:Enum} <: OptimizationResults
```
devmotion, line 214: AFAICT there's no need for a new type parameter:
```suggestion
mutable struct MultivariateOptimizationResults{O, Tx, Tc, Tf, M, Tls, Tsb} <: OptimizationResults
```

src/types.jl
```diff
@@ -201,9 +240,10 @@ mutable struct MultivariateOptimizationResults{O, Tx, Tc, Tf, M, Tls, Tsb} <: Op
     time_limit::Float64
     time_run::Float64
     stopped_by::Tsb
+    termination_code::TE
```
devmotion, line 243:
```suggestion
    termination_code::TerminationCode.T
```

Issue #1163, opened 2025-04-03
How does the L-BFGS preconditioning work? Improve documentation


kellertuer:
I have a question about optim and its preconditioner: How does that actually work for L-BFGS?

 The docs at https://julianlsolvers.github.io/Optim.jl/stable/algo/lbfgs/ just have a formula that looks a bit like it is a copy paste error from Gradient descent? The code I have not yet fully understood - there is a setup (https://github.com/JuliaNLSolvers/Optim.jl/blob/8a55bc3e4b4c1ccf86d5ea860de10d772c005171/src/multivariate/solvers/first_order/l_bfgs.jl#L198) and a apply (https://github.com/JuliaNLSolvers/Optim.jl/blob/8a55bc3e4b4c1ccf86d5ea860de10d772c005171/src/multivariate/solvers/first_order/l_bfgs.jl#L62) ? 

And why is that done “in the middle” of the two-loop, I would have expected it afterwards? Is theremaybe any documentation on why this was don as it is now?

If I understand it better, I would also love to improve the docs of course.

/cc @antoine-levitt

antoine-levitt:
In the docs I think P refers to the low rank approximation to the hessian, not to preconditioning, preconditioning is just bolted on afterwards. The precondprep thing is because several preconditioners depend on the current iterate, so precondprep tells them to update their local state in preparation for the apply. Re the implementation of preconditioning that dates back, I think @cortner is the one who did it?

kellertuer:
Hm, I do not understand the local state part here.

But then, still, literature would help, and that P is something else than the preconditioned is a bit unfortunate in the docs as well – but my goal is, after having understood this, to update the docs.

antoine-levitt:
> Hm, I do not understand the local state part here.

That depends on the application, but eg in DFT the preconditioner is something like (-alpha - Delta)^-1, where the constant alpha sets the length scale of the problem. That length scale is a priori unknown, so we compute it according to the current iterate. See https://github.com/JuliaMolSim/DFTK.jl/blob/b5873044316f04986c65567f79eb15cf8346f160/src/eigen/preconditioners.jl#L75. 

I think the docs predate the addition of preconditioner and just describes LBFGS generically.

kellertuer:
DFT is one reason I look into this ;)

The I still do not completely see why we have to have that in two steps, to me it looks like both the preconprep and the preconceived apply are only called once; just one before the for loop and the other one in between.

[...]

PR #1164, opened 2025-04-04
Improve (L-)BFGS docs

kellertuer:
This is an approach to #1163

Since even after very many tries I did not get beyond a few cross reference errors on the examples – they do render but something seems still wrong, I will try this PR to maybe render and look at the docs.


kellertuer:
Got it to work locally and also fixed the literature section to print the literature correctly. One could also replace that in the doc strings (wherever they are rendered).

So this is the form I would propose for the L-BFGS docs then.

kellertuer:
I think I am finished with this and if it is ok it could be merged, @pkofod – if your time allows.

Issue #1171, opened 2025-06-06
Modify `Optim.Options`


thchr:
I had hoped to be able to change an individual option in `Optim.Option`, inheriting all other options, but could not find a way. Is it already implemented?

I.e., something like
```jl
o = Optim.Options(; f_abstol = 1e-4, g_abstol=1e-3)
modified_o = Optim.Options(o; show_trace = true) # now with f_abstol = 1e-4, g_abstol=1e-3, show_trace = true
```

pkofod:
It is not possible, but may be a good idea to implement. Would you be up for giving it a go?

PR #1172, opened 2025-06-09
add `Options(options; kws...)` for inheriting from an existing options set

thchr:
This implements the feature described in #1171. I did not add the deprecated keyword arguments that are possible to pass to the `Optim(; kws..)` constructor; since this would be new API, it seemed unnecessary.

~~The implementation is not very DRY, since it needs to repeat the valid keyword arguments - but cleaner solutions would probably require a dependency like SetFields.jl or Accessors.jl, which doesn't feel warranted for a small feature like this.~~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Optim"
 uuid = "429524aa-4258-5aef-a3af-852621145aeb"
-version = "1.13.3"
+version = "1.14.0"
 
 
 [deps]
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Optim"
 uuid = "429524aa-4258-5aef-a3af-852621145aeb"
-version = "1.13.0"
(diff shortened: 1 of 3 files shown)


thchr:
Gentle bump.

pkofod:
I suppose users could just use https://github.com/JuliaObjects/Accessors.jl and set them themselves....

However, if we expose it as a feature directly of `Optim.Options` maybe we should not hardcode the list? I would think you could just get the property names and loop over that? Then, if we add a new option it should automatically be covered.

Project.toml
```diff
@@ -1,6 +1,7 @@
 name = "Optim"
 uuid = "429524aa-4258-5aef-a3af-852621145aeb"
-version = "1.13.2"
+version = "1.13.3"
```
pkofod, line 3:
```suggestion
version = "1.14.0"
```

pkofod:
Thanks

PR #1173, opened 2025-06-11
Fix SimulatedAnnealing docs

timholy:
`default_neighbor!` uses an argument-ordering that's the opposite of
what's in the docs.

Changing the docs makes this non-breaking, but if the docs are the preferred behavior, then perhaps a breaking change would be in order.


pkofod:
This looks good.

Issue #1170, opened 2025-06-03
Support for specialized matrix types (e.g., BlockBandedMatrices) with optimized solvers


rsenne:
Hello, and thank you for the excellent package.

I was wondering if there are plans to support additional matrix types that have specialized solving routines or optimized numerical algorithms. For instance, in my current workflow I am using a Hessian matrix that is block-tridiagonal. I pass it as a SparseMatrixCSC for performance, but I realized that I could instead represent it using [BlockBandedMatrices.jl](https://github.com/JuliaLinearAlgebra/BlockBandedMatrices.jl), which provides efficient linear solvers for such structures.

This would be particularly helpful in Newton-based optimization routines involving expressions like:
$\theta^{new} = \theta^{old} - H \setminus \nabla$

Having native support for block-banded matrices—or even a hook to dispatch to a custom solver—would be a great enhancement.

Thanks again for your work!

rsenne:
i poked around a bit and I think this is as easy as modifying `update_state!` like follows:

```
function update_state!(d, state::NewtonState, method::Newton)
    # Search direction is always the negative gradient divided by
    # a matrix encoding the absolute values of the curvatures
    # represented by H. It deviates from the usual "add a scaled
    # identity matrix" version of the modified Newton method. More
    # information can be found in the discussion at issue #153.
    T = eltype(state.x)
    H = NLSolversBase.hessian(d)
    g = gradient(d)

    if H isa AbstractSparseMatrix || H isa AbstractBlockBandedMatrix # Added BlockBandedMatrix support here
        # General fallback to backslash (dense RHS is fine)
        state.s .= -(H \ convert(Vector{T}, g))
    else
        state.F = cholesky!(Positive, H)
        if g isa Array
            ldiv!(state.s, state.F, -g)
        else
            gv = Vector{T}(undef, length(g))
            copyto!(gv, -g)
            copyto!(state.s, state.F \ gv)
        end
    end
```

I then ran the following MWRE:

```
using Optim, BlockBandedMatrices, LinearAlgebra

# test for BlockBandedMatrix Newton method
function test_blockbanded_newton()
    println("Testing Newton method with BlockBandedMatrices...")
    
    # Create a simple 4x4 block-banded matrix with 2x2 blocks
    n = 4
    λ = [2, 2]  # row block sizes
    μ = [2, 2]  # column block sizes
    l, u = 1, 1  # lower and upper bandwidth
    
[...]

PR #1176, opened 2025-07-10
Allow generic solver to be passed to Newton method

rsenne:
This PR is in reference to issue #1170. To solve this I have edited the interface to allow a user to pass a generic solver. This in theory would allow one to use any matrix type given the user has that package loaded and there exists a solver designed for that matrix type on board. This also allows one to write a custom solver (i.e., problem specific factorization, strategies for solving) to circumvent the default behavior. I have included some examples of this in testing including:

1. Different factorizations (would allow users to circumvent the default PositiveFactorizations.jl implementation of cholesky)
2. Usage of a non-originally supported matrix type (StaticArray)
4. Test type preservation of Hessian with TwiceDifferentiable object using BlockArrays.jl
3. A more realistic example of a non-trivial use case (block-tridiagonal hessian) this implementation is not totally optimized as it is a proof of principle

Let me know if there are any other tests you would like me to include or changes to the documentation.


src/multivariate/solvers/second_order/newton.jl
```diff
@@ -1,90 +1,88 @@
...
 ```
-
 ## Description
-The `Newton` method implements Newton's method for optimizing a function. We use
-a special factorization from the package `PositiveFactorizations.jl` to ensure
+The `Newton` method implements Newton's method for optimizing a function. 
+
+The `solve` function should take (H, g) and return s such that H*s = -g.
+Defaults to a robust solver that handles dense or sparse matrices.
+To do this, we use a special factorization from the package `PositiveFactorizations.jl` to ensure
```
pkofod, line 21:
```suggestion
If the matrix is not an `AbstractSparseMatrix`, we use a special factorization from the package `PositiveFactorizations.jl` to ensure
```

src/multivariate/solvers/second_order/newton.jl
```diff
@@ -1,90 +1,88 @@
...
+)
+    Newton(_alphaguess(alphaguess), linesearch, solve)
+end
+
+# Default solver that handles common matrix types intelligently
+function default_newton_solve(H, g)
+    if H isa AbstractSparseMatrix
+        return -(H \ g)
+    else
+        # Use PositiveFactorizations for robustness on dense matrices
```
pkofod, line 41:
```suggestion
        # Use PositiveFactorizations for robustness on dense matrices
         # Search direction is always the negative gradient divided by
         # a matrix encoding the absolute values of the curvatures
         # represented by H. It deviates from the usual "add a scaled
         # identity matrix" version of the modified Newton method. More
         # information can be found in the discussion at issue #153.
```

rsenne:
Hi @devmotion. You're right that's my mistake. In honesty I somewhat forgot what I was doing on this PR so my apologies for the late reply. I think i have fixed the regression I introduced. Let me know if there are any other issues

src/multivariate/solvers/second_order/newton.jl
```diff
@@ -40,51 +44,55 @@ end
...
-            copyto!(gv, -gradient(d))
+        # Use PositiveFactorizations for robustness on dense matrices
+         # Search direction is always the negative gradient divided by
+         # a matrix encoding the absolute values of the curvatures
+         # represented by H. It deviates from the usual "add a scaled
+         # identity matrix" version of the modified Newton method. More
+         # information can be found in the discussion at issue #153.
+         state.F = cholesky!(Positive, H)
+         if g isa Array
+            ldiv!(state.s, state.F, -g)
```
devmotion: Use
```suggestion
            ldiv!(state.s, state.F, g)
            state.s .= .-state.s
```
to reduce allocations?

src/multivariate/solvers/second_order/newton.jl
```diff
@@ -40,51 +44,55 @@ end
...
-            gv = Vector{T}(undef, length(gradient(d)))
-            copyto!(gv, -gradient(d))
+        # Use PositiveFactorizations for robustness on dense matrices
+         # Search direction is always the negative gradient divided by
+         # a matrix encoding the absolute values of the curvatures
+         # represented by H. It deviates from the usual "add a scaled
+         # identity matrix" version of the modified Newton method. More
+         # information can be found in the discussion at issue #153.
+         state.F = cholesky!(Positive, H)
+         if g isa Array
```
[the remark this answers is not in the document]
rsenne: There was an original comment in the code that said:

```
# is this actually StridedArray?
```
I think this is true, but didn't want to deviate from the source. But i will commit this change.

src/multivariate/solvers/second_order/newton.jl
rsenne: From the original code it looks like the gradient was negated?

```
ldiv!(state.s, state.F, -gradient(d))
```

Otherwise yes to the suggestion for reduced allocations.

src/multivariate/solvers/second_order/newton.jl
```diff
@@ -40,51 +44,55 @@ end
...
+            ldiv!(state.s, state.F, -g)
+         else
+            gv = convert(Vector{T}, length(g))
+            copyto!(gv, -g)
```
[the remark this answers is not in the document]
rsenne: I agree.

PR #1177, opened 2025-07-14
BUGFIX: Update preconditioner in cg

IvanBioli:
After the formatting of the repository in #1150 , a small error was introduced in the `update_state!` function in the `src/multivariate/solvers/first_order/cg.jl` file: the preconditioner is not updated at every iteration, but only at the first one. The change ensures that the `_apply_precondprep` method is called. Changes:
* [`src/multivariate/solvers/first_order/cg.jl`](diffhunk://#diff-0a73ae73cf3162a4ba402251714e7921cc7521eae370fae3ff1539b469bcd1d1R187): Added a call to `_apply_precondprep(method, state.x)`


IvanBioli:
> Thanks for noticing, can you link to the line in question? I don't see any changes to cg.jl in that pr :)

I added ```_apply_precondprep(method, state.x)``` at line 187

PR #1192, opened 2025-11-01
Update benchmark workflow based on AirspeedVelocity docs

devmotion:
The purpose of this PR is two-fold:
- Simplify the definition of the benchmarking GHA job based on the docs of AirspeedVelocity
- Thereby support benchmarking of PRs from forks

I'm a bit unsure about the second point, but based on eg https://github.com/JuliaPhysics/DynamicQuantities.jl/pull/183 my impression is that the action should support PRs from forks.

### DIFF
--- a/.github/workflows/benchmark_pr.yml
+++ b/.github/workflows/benchmark_pr.yml
@@ -4,75 +4,15 @@ on:
   pull_request_target:
     branches:
       - master
-
 permissions:
-  pull-requests: write
+  pull-requests: write    # needed to post comments
 
 jobs:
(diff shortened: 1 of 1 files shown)


devmotion:
Ah I think I might have to update/refresh my fork such that it contains the latest commit on master in the original repo. I wonder if this could avoided by upstream changes in AirspeedVelocity

pkofod:
Did you update your fork? If so, it seems like it has the same issue still

devmotion:
At least it can be tested once it is merged.

PR #1195, opened 2025-11-02
Use `autodiff` API based on `ADTypes` instead of symbols

devmotion:
Fixes https://github.com/JuliaNLSolvers/Optim.jl/issues/1187

### DIFF
--- a/src/multivariate/optimize/interface.jl
+++ b/src/multivariate/optimize/interface.jl
@@ -4,6 +4,9 @@ fallback_method(f) = NelderMead()
 fallback_method(f, g!) = LBFGS()
 fallback_method(f, g!, h!) = Newton()
 
+# By default, use central finite difference method
+const DEFAULT_AD_TYPE = ADTypes.AutoFiniteDiff(; fdtype = Val(:central))
+
 function fallback_method(f::InplaceObjective)
     if !(f.fdf isa Nothing)
         if !(f.hv isa Nothing)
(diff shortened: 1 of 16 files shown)


src/multivariate/optimize/interface.jl
```diff
@@ -4,6 +4,9 @@ fallback_method(f) = NelderMead()
 fallback_method(f, g!) = LBFGS()
 fallback_method(f, g!, h!) = Newton()
 
+# By default, use central finite difference method
+const DEFAULT_AD_TYPE = ADTypes.AutoFiniteDiff(; fdtype = Val(:central))
```
devmotion: One could consider switching to a different default AD backend (maybe ForwardDiff for univariate optimization and e.g. Mooncake - if at least all tests pass - for multivariate optimization problems? - but the choice for when to switch to which backend is likely also problem/dimension dependent, see also https://docs.sciml.ai/Optimization/stable/API/ad/#ad). But I think it would be better to make such more fundamental changes in a separate PR.

pkofod: I agree both to switch to reverse mode for multivariate and forward for univariate, but also that it's probably a separate PR

pkofod:
JET complains

devmotion:
I checked it locally, and none of the things JET complains are introduced or changed by the PR. It seems the PR just helps JET to perform its type-signature-based analysis more thoroughly, and hence more problems are revealed 😄 

I could fix a few of them in this PR but arguably it might be better to do that in a separate PR.

PR #1196, opened 2025-11-02
Update CI: Test min + LTS and test macOS with native binaries

devmotion:


pkofod:
the tests that fail are quite sensitive / unstable

devmotion:
@pkofod You could release the bug fix in https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/163 (I bumped the version in that PR and the changes up until that commit should be non-breaking) to see if it fixes the test failures here.

pkofod:
given the length of the skip list it may be time to find a better way to test this method :)

PR #1198, opened 2025-11-02
Add tests with Aqua, ExplicitImports and JET

devmotion:
For now tests with JET are just marked as broken; once they are fixed one could switch to `JET.test_package`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,9 +2,7 @@ name = "Optim"
 uuid = "429524aa-4258-5aef-a3af-852621145aeb"
 version = "1.14.0"
 
-
 [deps]
-Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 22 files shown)


pkofod:
wrt the conflicts I removed those methods in another PR. MethodError() is an error in itself and it makes no sense to add a method just to throw a method error :)

pkofod:
I suppose maximum in src/maximize.jl  needs a `Base.` qualifier?

pkofod:
With maximum and minimum now based on the Base function I just wanted to say that I would probably be okay with that. I originally chose not to do it, because I was unsure if `minimum` fit the "intention" of Base. minimum. The latter gives you the smallest value in collection, whereas minimum here gives you a number that is the result of some algorithm. Though in a sense it does (try) to return the smallest function value amongst the function values the feasible set so I would be okay to do it, I just wanted to specify why it was not exported in the past.

PR #1201, opened 2025-11-03
Work around upstream issue by running benchmarks on Julia LTS

devmotion:
Workaround for https://github.com/MilesCranmer/AirspeedVelocity.jl/issues/113 which seems to be caused by a bug in Pkg on Julia 1.12.

Note that due to the restrictions of pull_request_target in this PR the action runs with the existing config on the master branch (ie on Julia 1.12) and not the updated config in this PR (on Julia LTS).


pkofod:
The idea is just to catch any obvious mistakes, so I don't think the version of Julia should matter too much. 👍

PR #1207, opened 2025-11-13
Use gradient!(d, x). in perform_linesearch

pkofod:
This shouldn't matter, but *if* somehow happened to be calling the objective function in the callback or some other way this would now produce the correct result. Previously, it used the most recent evaluation. Now, it will try to evaluate at state.x and if that was the last point evaluated we'll just use the cached result. This should always be the case unless someone messed with the objective outside. We could even assert that g_calls is constant around the call to be sure to catch it but then users wouldn't be able to call the gradient outside at their own expense of runtime.

### DIFF
--- a/src/utilities/perform_linesearch.jl
+++ b/src/utilities/perform_linesearch.jl
@@ -40,13 +40,14 @@ end
 
 function perform_linesearch!(state, method, d)
     # Calculate search direction dphi0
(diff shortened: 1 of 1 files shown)


pkofod:
Interesting that fminbox fails here.. that would indicate that it was doing some funny business in the resets

PR #1210, opened 2025-11-24
Compute JVP in line searches

devmotion:
Based on https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/168 and https://github.com/JuliaNLSolvers/LineSearches.jl/pull/187.

~~The diff would be cleaner if #1195 (which depends on #1209) would be available on the master branch.~~

Based on #1212.

---

Currently, tests are failing due to missing definitions of `NLSolversBase.value_jvp!` etc. for `ManifoldObjective`.

### DIFF
--- a/src/Manifolds.jl
+++ b/src/Manifolds.jl
@@ -29,6 +29,10 @@ function NLSolversBase.value!(obj::ManifoldObjective, x)
     xin = retract(obj.manifold, x)
     return value!(obj.inner_obj, xin)
 end
+function NLSolversBase.value(obj::ManifoldObjective, x)
+    xin = retract(obj.manifold, x)
+    return value(obj.inner_obj, xin)
+end
 function NLSolversBase.gradient!(obj::ManifoldObjective, x)
     xin = retract(obj.manifold, x)
     g_x = gradient!(obj.inner_obj, xin)
--- a/src/multivariate/solvers/constrained/ipnewton/interior.jl
+++ b/src/multivariate/solvers/constrained/ipnewton/interior.jl
@@ -526,7 +526,7 @@ constraints do not contribute to `h`, because the hessian of `x_i` is
 zero. (They contribute indirectly via their slack variables.)
 """
 hessianI(x, constraints, λcI, μ) =
-    hessianI!(zeros(eltype(x), length(x), length(x)), x, constraints, λcI, μ)
+    hessianI!(zeros(eltype(x), (length(x), length(x))), x, constraints, λcI, μ)
 
 """
     userλ(λcI, bounds) -> λ
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -357,7 +357,7 @@ function initial_state(method::NewtonTrustRegion, options, d, x0)
 end
 
 
-function update_state!(d, state::NewtonTrustRegionState, method::NewtonTrustRegion)
+function update_state!(d::TwiceDifferentiable, state::NewtonTrustRegionState, method::NewtonTrustRegion)
     # Find the next step direction.
     m, state.interior, state.lambda, state.hard_case, state.reached_subproblem_solution =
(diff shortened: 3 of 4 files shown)


src/multivariate/solvers/first_order/accelerated_gradient_descent.jl
```diff
@@ -42,7 +42,7 @@ function initial_state(
     initial_x = copy(initial_x)
     retract!(method.manifold, initial_x)
 
-    value_gradient!!(d, initial_x)
```
devmotion: @pkofod why would we ever want to call the `!!` methods explicitly? If the input is different from the one that the value and gradient were evaluated with the value/gradient will be recomputed anyway.

pkofod: An alternative is to create new instances of OnceDifferentiable, but the "problem" is Fminbox that has outer and inner loops. In the previous inner loop you may evaluate the objective at x but then Fminbox updates a parameter that changes the function essentially and then the stored value is not correct. 

Without getting too specific here, you can thing of it as the EBEs in Laplace's method. If you have an outer objective that performs an EBE optimization on the inside then a OnceDifferentiable for the outside objective of the marginal loglikelihood cannot know if we changes the inner EBEs or not. 

As I said above, maybe a better approach is to construct a new OnceDifferentiable per outer iteration.

I believe other users have relied on it in the past as well, because they are doing the same. Allocating / constructing the OnceDifferntiable once and then they're solving a sequence of optimize with some parameter that's updated between optimize calls. I'm not completely sure if it's documented or not, but the idea is that each initial evaluation and each reset! for Fminbox forces an evaluation even if `x` is not updated.

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -64,72 +31,109 @@ function _barrier_term_gradient(x::T, l, u) where {T}
...
+end
+
+in_box(bw::BarrierWrapper, x) = in_box(bw.b, x)
+
 # Wrappers
-function value!!(bw::BarrierWrapper, x)
-    bw.Fb = value(bw.b, x)
-    bw.Ftotal = bw.mu * bw.Fb
+NLSolversBase.value(obj::BarrierWrapper) = obj.Ftotal
+function NLSolversBase.value!(bw::BarrierWrapper, x)
```
pkofod: I didn't check CI yet, but I think these may call failures because the multiplier could have been updated

pkofod: I guess you can rewrite these to check if mu was updated in the same manner...

src/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -536,12 +597,13 @@ function optimize(
     end
     results = optimize(dfbox, x, _optimizer, options, state)
     stopped_by_callback = results.stopped_by.callback
-    dfbox.obj.f_calls[1] = 0
+    # TODO: Reset everything? Add upstream API for resetting call counters?
```
pkofod: As discussed above, `clear!` should suffice

devmotion:
Yes, these test issues are architecture specific and have persisted for a long time in this PR: https://github.com/JuliaNLSolvers/Optim.jl/pull/1210#issuecomment-3598918787

devmotion:
> I suppose the JET errors only shows up after the combined change to initial_state as well as upgrading to v8 of NLSolvesrBase?

No, I don't think the `initial_state` change is relevant at all. It's caused by rebasing on master which contains #1223. JET@0.11 (requires Julia >= 1.12) is just much better at detecting issues. Currently we only use JETs to detect "typos" (undefined references) and perform a package-wide analysis that is based on the method signatures (and hence generally not very targeted). Hence the latter is affected quite a bit by the major code changes and signature changes in the PR.

I pushed https://github.com/JuliaNLSolvers/Optim.jl/pull/1210/commits/9fa280e83cda949ce506ac5bb930e050d7ca91d5, a minimal set of changes that fixed the JET errors locally.

PR #1211, opened 2025-11-26
Add Julia 1.12 to benchmarks

devmotion:
This Pkg bug might have been fixed in 1.12.


devmotion:
@pkofod shall we try it?

PR #1218, opened 2025-12-03
Remove `delta_min` setting of `NewtonTrustRegion`

devmotion:
I propose to remove the `delta_min` setting, and in particular its non-zero default value, from the `NewtonTrustRegion` algorithm.

I observed in https://github.com/JuliaStats/Survival.jl/pull/69, that with the default `delta_min` tests fail (termination not due to gradient norm or objective (function) value) whereas with `delta_min = 0` they pass. IMO generally, if the function is not differentiable and optimization fails, it will fail anyway; however, by exiting early due to `delta_min` there'll always be cases where non-optimal values are returned in an otherwise fine optimization.

### DIFF
--- a/src/multivariate/optimize/optimize.jl
+++ b/src/multivariate/optimize/optimize.jl
@@ -116,15 +116,6 @@ function optimize(
             stopped = true
         end
 
-        if method isa NewtonTrustRegion
-            # If the trust region radius keeps on reducing we need to stop
-            # because something is wrong. Wrong gradients or a non-differentiability
-            # at the solution could be explanations.
-            if state.delta ≤ method.delta_min
-                stopped = true
-            end
-        end
-
         if hasproperty(state, :g_x) && !all(isfinite, state.g_x)
             options.show_warnings && @warn "Terminated early due to NaN in gradient."
             break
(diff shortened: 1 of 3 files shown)


devmotion:
My argument for removing `delta_min` completely is that there are already other, algorithm independent, ways to stop early in problematic cases, such as `f_reltol` or `iterations`. As the scaling example shows, if the values are not scaled accordingly, sometimes `delta_min` has no effect at all whereas e.g. `f_reltol` would still allow you to exit early.

pkofod:
> My argument for removing `delta_min` completely is that there are already other, algorithm independent, ways to stop early in problematic cases, such as `f_reltol` or `iterations`. As the scaling example shows, if the values are not scaled accordingly, sometimes `delta_min` has no effect at all whereas e.g. `f_reltol` would still allow you to exit early.

I don't think you can use those settings because these steps are probably rejected so we don't consider the "iteration" for convergence.

pkofod:
> I'm still questioning this option, but I close this PR for now given that #1219 was merged.

I understand the concern. You can set a maxiter instead. What should that then be. I can agree that any number you set should be set so small that there can be no doubt that your number is indeed small in the domain you're looking at.

The nonlinear nature of nonlinear optimization does pose a difficulty, because we have no nice properties of guaranteed decrease of some measure (say gradient sup norm) towards the minimum, so there is no way to say it any run won't have to have a step that is in a 1e-32 radius ball only to then start taking larger steps. 🤷

pkofod:
I'm open to have an issue. It fits in with the gradient norm as I said earlier and as you linked. The question of dimensionality seems central to the practically stopping at "optima" but it does seem like it's a fundamentally unsolved problem in practical numerical optimization.

PR #1222, opened 2025-12-06
Fix bounds errors in default `x_abschange` and `x_relchange`

devmotion:
I noticed that currently the default `x_abschange` and `x_relchange` assume that the state is non-empty when running parameter estimation tests with an artificial example without any parameters. Clearly, that case is not useful in practice but nevertheless I think it would be good to fix the current unsafe behaviour and use the opportunity to simplify the existing code.

One consideration in this PR is whether to use `StatsBase.Linfdist` or whether only to copy this arguably very simple function. Given that StatsBase is not needed in this package otherwise (see #1221), I opted for copying and adapting the implementation.

### DIFF
--- a/src/utilities/assess_convergence.jl
+++ b/src/utilities/assess_convergence.jl
@@ -9,6 +9,11 @@ x_relchange(state::AbstractOptimizerState) = x_relchange(state.x, state.x_previo
 x_relchange(x::AbstractArray{<:Number}, x_previous::AbstractArray{<:Number}) = Linfdist(x, x_previous) / Base.maximum(abs, x) # Base.maximum !== maximum
 
 # Copied and adapted from https://github.com/JuliaStats/StatsBase.jl/blob/d70c4a203177c79d851c1cdca450bc6dbd2a4683/src/deviation.jl#L100-L110
+#    Linfdist(a, b)
+#
+# Compute the L∞ distance, also called the Chebyshev distance, between
+# two arrays: ``\\max_{1≤i≤n} |a_i - b_i|``.
+# Efficient equivalent of `maximum(abs, a - b)`.
 function Linfdist(x::AbstractArray{<:Number}, y::AbstractArray{<:Number})
     n = length(x)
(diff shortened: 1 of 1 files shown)


src/utilities/assess_convergence.jl
```diff
@@ -4,9 +4,21 @@ f_relchange(state::AbstractOptimizerState) = f_relchange(state.f_x, state.f_x_pr
...
 
 x_abschange(state::AbstractOptimizerState) = x_abschange(state.x, state.x_previous)
-x_abschange(x, x_previous) = maxdiff(x, x_previous)
+x_abschange(x::AbstractArray{<:Number}, x_previous::AbstractArray{<:Number}) = Linfdist(x, x_previous)
 x_relchange(state::AbstractOptimizerState) = x_relchange(state.x, state.x_previous)
-x_relchange(x, x_previous) = maxdiff(x, x_previous) / Base.maximum(abs, x) # Base.maximum !== maximum
+x_relchange(x::AbstractArray{<:Number}, x_previous::AbstractArray{<:Number}) = Linfdist(x, x_previous) / Base.maximum(abs, x) # Base.maximum !== maximum
+
+# Copied and adapted from https://github.com/JuliaStats/StatsBase.jl/blob/d70c4a203177c79d851c1cdca450bc6dbd2a4683/src/deviation.jl#L100-L110
+function Linfdist(x::AbstractArray{<:Number}, y::AbstractArray{<:Number})
```
pkofod:
```suggestion
#    Linfdist(a, b)
#
# Compute the L∞ distance, also called the Chebyshev distance, between
# two arrays: ``\\max_{1≤i≤n} |a_i - b_i|``.
# Efficient equivalent of `maxabs(a - b)`.
function Linfdist(x::AbstractArray{<:Number}, y::AbstractArray{<:Number})
```

pkofod: no need to make it a docstring, but I figured it would be nice to have the description here for developers

PR #1226, opened 2025-12-21
Document optimization state and callback functionality. Change initia…

### LINKED ISSUE #962: No API docs for the callback arguments
Hi! Maybe I was simply unable to find it anywhere, but it seems there is no documentation on what is/are the arguments for the `callback` passed to `Options`.
From the code, it seems to be a `struct` of some sort.
Maybe one could improve the docs/docstring on this side?

pkofod:
…l_state to initial_x in the user-facing API and export initial_state.

Fixes #962

### DIFF
--- a/test/general/api.jl
+++ b/test/general/api.jl
@@ -254,7 +254,7 @@ end
     @test Optim.upper_bound(res) == 1.0
     @test Optim.rel_tol(res) ≈ 1.4901161193847656e-8
     @test Optim.abs_tol(res) ≈ 2.220446049250313e-16
-    @test_throws ErrorException Optim.initial_state(res)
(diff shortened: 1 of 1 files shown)


docs/src/user/callbacks.md
```diff
@@ -0,0 +1,27 @@
...
+function my_callback(state)
+    print(" Objective Value: ", state.f_x)
+    println(" at state x: ", state.x)
+    return false  # Return true to stop the optimization
+end
+function objective(x)
+    return (x[1]-2)^2 + (x[2]-3)^2
+end
+
+initial_x = [0.0, 0.0]
```
devmotion: Slightly confusing to use the same name as for the API function here. At least it confused me initially 😄

PR #1230, opened 2026-01-20
Fix manifold value_gradient! call such that it returns both things.

pkofod:
Fixes 1.9 for LineSearches v7.5.1. This fix is also appropriate for older versions of LineSearches as it should be the returned thing, but the ManifoldObjective wasnt used in the problematic manner so I'm not updating the bounds.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -23,7 +23,7 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - name: Cache artifacts
-        uses: actions/cache@v1
+        uses: actions/cache@v4
         env:
           cache-name: cache-artifacts
         with:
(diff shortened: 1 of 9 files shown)


devmotion:
I don't understand this PR. This issue was already fixed in Optim@1.13.3 by https://github.com/JuliaNLSolvers/Optim.jl/pull/1213. Even if you would backport this fix to Optim@1.9, then still all other existing releases < Optim@1.13.3 would suffer from this problem: https://github.com/JuliaNLSolvers/LineSearches.jl/issues/192

Either we just close https://github.com/JuliaNLSolvers/LineSearches.jl/issues/192 and tell users to update to Optim@1.13.3. Or - since this might also not be an option for all users due to restrictions imposed by other dependencies - we update the registry to only allow LineSearches < 7.5.1 on Optim < 1.13.3. I don't think yanking LineSearches@7.5.1 is an option, as that release is completely fine per se (the bug is an Optim bug in older releases) and development has moved on, using the same pattern in LineSearches, in subsequent releases.

pkofod:
for some reason I got 1.9.4 when running something locally, but clearly that is not the last version prior to v2. My bad!

PR #1235, opened 2026-03-16
Various fixes to the MOI wrapper

odow:


pkofod:
Yes, seems better to handle the problem type directly rather than removing the tests in test/

PR #1242, opened 2026-04-10
Try to restart the optimization if the line search fails.

pkofod:


pkofod:
I was able to relax some tests with these changes. The Trigonometric problem clearly converges to a local minimum so the optimizers generally *work* however it seems that the starting point leads to a minimum that is not the optimal, so it's kind of only suited for "global optimization" workflows I think. Edit: wrong, it's just that the trigonometric objective has periodic solutions so x should not be compared to the zero vector

PR #1244, opened 2026-04-12
Add support for the user defined rng in ParticleSwarm

ParfenovS:
This PR adds support for the user defined rng in ParticleSwarm procedure. This is useful in the case when, for example, there are several optimizers running in parallel using Threads and one wants to achieve reproducibility.

### DIFF
--- a/src/multivariate/solvers/zeroth_order/particle_swarm.jl
+++ b/src/multivariate/solvers/zeroth_order/particle_swarm.jl
@@ -2,6 +2,7 @@ struct ParticleSwarm{Tl,Tu} <: ZerothOrderOptimizer
     lower::Tl
     upper::Tu
     n_particles::Int
+    rng::Random.AbstractRNG
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/multivariate/solvers/zeroth_order/particle_swarm.jl
```diff
@@ -217,7 +222,7 @@ function update_state!(f, state::ParticleSwarmState{T}, method::ParticleSwarm) w
     for k = 1:n
         state.x_learn[k] = state.x[k]
     end
-    random_index = rand(1:n)
+    random_index = rand(rng,1:n)
     random_value = randn()
```
pkofod, line 226: was this an oversight? I think all random number generators should receive the rng

src/multivariate/solvers/zeroth_order/particle_swarm.jl
pkofod, line 229: also here

ParfenovS: Yes, this is an oversight. There should be randn(rng). My bad.

PR #1246, opened 2026-04-12
Update documentation

abhro:
Please see each commit and associated message for a more atomic description of each change. This PR somewhat conflicts with #1245, but if that's merged, a clean merge conflict resolution is simple.

PR #1247, opened 2026-04-13
Add/update language syntax highlight tags for fenced code block examples

abhro:
Commit cherry-picked from #1246, and removed from there since it's a big enough diff to barely make it worth it to have its own PR.


pkofod:
Thanks, appreciate it!

PR #1257, opened 2026-06-05
Guard against gbarrier_norm == 0 in computation of initial mu (fixes #1233)

### LINKED ISSUE #1233: Default value of `Fminbox.mu0 = NaN` causing optimisation failures on v2.0.1.
Upgrading from v2.0.0 to v2.0.1 causes failures in previously working optimisations. I've taken to setting a finite value for mu0, but it appears the automatic tuning might be broken. I think it can be traced to PR #1232 . I've checked the initial values of the gradient and function and there's nothing egregious.

RafaelArutjunjan:
Fixes #1233 

If both `gnorm == 0` and `gbarrier_norm == 0`, then `mu = 0/0` which evaluates to `NaN`. I.e., this error is produced if all parameters are unbounded. Bug was introduced by removing safety catch in commit 9b664c2f "Re-instate old behavior for Fminbox..." leading up to v2.0.1.

I added back the old safety check and a dedicated test for good measure.

### DIFF
--- a/test/multivariate/solvers/constrained/fminbox.jl
+++ b/test/multivariate/solvers/constrained/fminbox.jl
@@ -181,6 +181,9 @@
                 Optim.Options(),
             )
         end
+        @testset "Guard against gbarrier_norm == 0 in computation of initial mu #1233" begin
+            optimize(exponential, fill(-Inf,2), fill(Inf,2), initial_x, Fminbox(), Optim.Options())
+        end
     end
 end
 
--- a/test/multivariate/solvers/constrained/fminbox.jl
+++ b/test/multivariate/solvers/constrained/fminbox.jl
@@ -182,7 +182,31 @@
             )
         end
         @testset "Guard against gbarrier_norm == 0 in computation of initial mu #1233" begin
-            optimize(exponential, fill(-Inf,2), fill(Inf,2), initial_x, Fminbox(), Optim.Options())
+            l_inf, u_inf = fill(-Inf, 2), fill(Inf, 2)
+        
+            # gnorm > 0, gbarrier_norm == 0: previously mu = gnorm/0 = Inf
+            res = optimize(exponential, l_inf, u_inf, initial_x, Fminbox(), Optim.Options())
+            @test Optim.converged(res)
+            @test Optim.minimizer(res) ≈ [2.0, 3.0] atol = 1e-3
+                                
+            # gnorm == 0 and gbarrier_norm == 0: previously mu = 0/0 = NaN
+            # ([2.0, 3.0] is the stationary point of `exponential`)
+            res = optimize(exponential, l_inf, u_inf, [2.0, 3.0], Fminbox(), Optim.Options())
+            @test Optim.converged(res)
+            @test Optim.minimizer(res) ≈ [2.0, 3.0]
+        
+            # initial_mu unit tests
+            box_inf = Optim.BoxBarrier(l_inf, u_inf)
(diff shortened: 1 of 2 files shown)


pkofod:
Could you find a test example that actually has bounds? I spent a lot of trying to figure this out and I believe there was a reason I removed it...

RafaelArutjunjan:
I cannot think of an example off the top of my head to be honest but what could be the benefit for not having this simple check? It's not like this would affect performance in any relevant way? 

Of course there could be more elaborate ways to have fminbox gracefully handle unbounded parameters but this seems like the simplest, most straightforward fix? The alternative is users receiving a cryptic error message they have to spend time and energy on to diagnose what the actual problem is.

pkofod:
> I cannot think of an example off the top of my head to be honest but what could be the benefit for not having this simple check? It's not like this would affect performance in any relevant way?

I see it now, sorry I was mixing this with a different case where it the barrier terms would cancel out and the zero would just be due to the centering, but we added a heuristic initial mu for that. Within initial_mu, gbarrier_norm is computed with the abs transform (precisely so centered starting points don't produce zero), so gbarrier_norm == 0 can only mean no coordinate has a finite bound, and mu = 0 is the right answer there since the barrier is identically zero anyway.

The test you added does not test for NaN though I think, because the gradient is finite here, so it's only testing the Inf case.

test/multivariate/solvers/constrained/fminbox.jl
```diff
@@ -181,6 +181,9 @@
                 Optim.Options(),
             )
         end
+        @testset "Guard against gbarrier_norm == 0 in computation of initial mu #1233" begin
+            optimize(exponential, fill(-Inf,2), fill(Inf,2), initial_x, Fminbox(), Optim.Options())
```
pkofod:
```suggestion
            l_inf, u_inf = fill(-Inf, 2), fill(Inf, 2)
        
            # gnorm > 0, gbarrier_norm == 0: previously mu = gnorm/0 = Inf
            res = optimize(exponential, l_inf, u_inf, initial_x, Fminbox(), Optim.Options())
            @test Optim.converged(res)
            @test Optim.minimizer(res) ≈ [2.0, 3.0] atol = 1e-3
                                
            # gnorm == 0 and gbarrier_norm == 0: previously mu = 0/0 = NaN
            # ([2.0, 3.0] is the stationary point of `exponential`)
            res = optimize(exponential, l_inf, u_inf, [2.0, 3.0], Fminbox(), Optim.Options())
            @test Optim.converged(res)
            @test Optim.minimizer(res) ≈ [2.0, 3.0]
        
            # initial_mu unit tests
            box_inf = Optim.BoxBarrier(l_inf, u_inf)
            @test Optim.initial_mu(box_inf, [0.0, 0.0], [1.0, 1.0], Fminbox()) == 0  # Inf case
            @test Optim.initial_mu(box_inf, [0.0, 0.0], [0.0, 0.0], Fminbox()) == 0  # NaN case
        
            # half-bounded box: a single finite bound must still give mu > 0
            box_half = Optim.BoxBarrier([-Inf, 0.0], [Inf, Inf])
            @test Optim.initial_mu(box_half, [0.0, 1.0], [1.0, 1.0], Fminbox()) > 0
        
            # explicitly non-finite user-supplied mu0 should still throw
            @test_throws ArgumentError Optim.initial_mu(box_inf, [0.0, 0.0], [1.0, 1.0],
                                                        Fminbox(LBFGS(); mu0 = Inf))
```

RafaelArutjunjan: Sure, makes sense to test `initial_mu` more directly to check that other possible cases are also handled well.

pkofod:
Thanks!

PR #1259, opened 2026-06-09
update CG with theta parameter from HZ2013 and betamax to bound the beta value

nmbader:
Add two new parameters to the ConjugateGradient solver
1) The missing `theta` as it appears in the Hager-Zhang 2013 paper, Equation 2.2. The default value is `theta=1` which ensures back compatibility
2) `betamax` to bound the `beta` value (in the absolute sense) which is used in determining the step direction. This is to allow the user to control the magnitude of `beta` and prevent it from overwhelming the gradient at any given iteration. The default value is `betamax=Inf` which has no effect and ensures back compatibility.

### DIFF
--- a/src/multivariate/solvers/first_order/cg.jl
+++ b/src/multivariate/solvers/first_order/cg.jl
@@ -44,6 +44,8 @@
 
 struct ConjugateGradient{Tf,T,Tprep,IL,L} <: FirstOrderOptimizer
     eta::Tf
+    theta::Tf
+    betamax::Tf
     P::T
     precondprep!::Tprep
     alphaguess!::IL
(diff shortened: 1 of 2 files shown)


src/multivariate/solvers/first_order/cg.jl
```diff
@@ -235,10 +244,11 @@ function accept_step!(d, state::ConjugateGradientState, method::ConjugateGradien
     betak =
         (
             real(dot(state.y, state.pg)) -
-            real(dot(state.y, state.py)) * real(dot(state.g_x, state.s)) / ydots
+            real(dot(state.y, state.py)) * real(dot(state.g_x, state.s)) / ydots * method.theta
         ) / ydots
     # betak may be undefined if ydots is zero (may due to f not strongly convex or non-Wolfe linesearch)
     beta = NaNMath.max(betak, etak) # TODO: Set to zero if betak is NaN?
+    beta = sign(beta) * Base.min(abs(beta), method.betamax) # Ensure beta is not too large in absolute value
```
pkofod: or 
```
beta = clamp(beta, -method.betamax, method.betamax)
```
but either one is fine I suppose

PR #1275, opened 2026-08-18
Guard the trust region radius against rejected runaway steps

pkofod:
Two related gaps around `solve_tr_subproblem!`.

**Unassigned step on early returns.** The non-finite-Hessian and empty-eigenvalue returns left `s` untouched, so the caller read the NaN fill from `initial_state` on iteration 1, or a stale step later. Those paths now zero `s`: a zero step is rejected downstream and shrinks the radius instead of poisoning it.

**Rejection can enlarge the radius.** `state.delta = norm(state.s) / 4` on a rejected step is only a shrink when `norm(s) ≤ delta`. An unconverged root-finding solve can hand back a step orders of magnitude longer than the radius: on Extended Powell with a finite-difference Hessian, a solve at `delta = 0.25` returned a step of norm 1542, the "shrink" reset the radius to 385, and the solver locked into a reject-explode cycle until the iteration cap. The update is now `min(delta, norm(s)) / 4`, falling back to `delta / 4` when `norm(s)` is not finite, so a rejection always shrinks.

Adds a subproblem-level test that a non-finite Hessian leaves a well-defined zero step. The reject-explode cycle itself is exercised end to end by the standard `run_optim_tests` suite once the interior-classification follow-up lands on top of this (that change makes Extended Powell with an approximated Hessian reach the cycle deterministically without this guard).

Verified: `Subproblems I/II` and the PSD/interior sweep at 0 fails, the NewtonTrustRegion test file green, full suite green.

### DIFF
--- a/src/multivariate/solvers/second_order/newton_trust_region.jl
+++ b/src/multivariate/solvers/second_order/newton_trust_region.jl
@@ -98,6 +98,9 @@ function solve_tr_subproblem!(gr, H, delta, s; tolerance = 1e-10, max_iters = 5)
     # symmetric.  (Julia issue #17093)
     Hsym = Symmetric(H)
     if any(!isfinite, Hsym)
+        # Leave a well-defined (zero) step behind: callers read s after this
+        # returns, and a stale or NaN-filled s poisons the radius update.
+        fill!(s, zero(T))
         return T(Inf), false, zero(T), false, false
     end
     H_eig = eigen(Hsym)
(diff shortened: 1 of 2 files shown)


pkofod:
The safe-guard `min` here may seem surprising because `s` should not be longer than the radius but is a fallback for situations where the inner solution is not solved to completion even if the flag states it is. This could happen due to the heuristic that is used to stop. Future changes could perhaps improve this for example with the bounds derived here https://cran.r-project.org/web/packages/trust/vignettes/trust.pdf

PR #1277, opened 2026-08-18
Give Cholesky retries their own budget in the trust region root find

pkofod:
A failed factorization took the same path as a root-finding iteration: the retry sat inside `for iter = 1:max_iters`, so each failure burned one of the five iterations. With `H = [1 1; 1 1]`, `g = [1, 0]`, the first Cholesky at `lambda ≈ lambda_lb` fails, and the solve reported `reached_solution = false` at the default `max_iters = 5` but `true` at 6, with an identical step either way.

Root-finding iterations are now counted only on successful factorizations, and retries are bounded separately. The bound is derived, not picked: failures occur only while `lambda ≤ -min_H_ev ≤ H_scale`, and each failure at least doubles lambda from `sqrt(eps(T))·H_scale` upward, so about half the significand width in doublings suffices (`4 + ceil(Int, precision(T) / 2)`). The default `max_iters = 5` is unchanged; whether to raise it is a separate question.

Adds the `g = [1, 0]` case as a regression test asserting `reached_solution` at the default `max_iters`, with the boundary step and model value it should return.

Verified: `Subproblems I/II` and the PSD/interior sweep at 0 fails; the NewtonTrustRegion test file green.


pkofod:
irrelevant after #1288

PR #1286, opened 2026-09-13
Compute the line search's alpha = 0 slope from the gradient again

### LINKED ISSUE #1284: perform_linesearch! computes dphi_0 from the gradient even when the objective supplies a JVP

andreasnoack:
# Compute the line search's alpha = 0 slope from the gradient again

Reverts #1285, which fixed #1284. It restores `dphi_0 = dot(state.g_x, state.s)` in `perform_linesearch!` and replaces the test with one asserting that the objective's `fjvp` is not called for `dphi_0`.

## Why #1285 does not hold up

#1285 took `dphi_0` from `NLSolversBase.jvp!(d, x, s)` so that the slope at `alpha = 0` and the slopes the line search samples come from one derivative implementation. For an objective that supplies `fjvp`, that is a full extra evaluation on every iteration. The argument for paying it was that `dot(g, s)` and the JVP can differ by up to 20% relative near a stationary point because "`dot(g, s)` cancels across the components where a pushforward does not". Measured more carefully on the same kind of objective (a marginal likelihood through an adaptive ODE solver, BFGS + HagerZhang(rho = 2.0)), that explanation is wrong and the consistency buys nothing:

| ODE reltol / abstol | inner optimisation g_tol | median relative gap between the routes | max |
|---|---|---|---|
| 1e-8 / 1e-12 (defaults) | 1e-5 | 2.2e-8 | 8.9e-8 |
| 1e-12 / 1e-14 | 1e-5 | 4.6e-13 | 3.6e-12 |
| 1e-8 / 1e-12 | 1e-10 | 2.3e-8 | 8.7e-8 |
| 1e-12 / 1e-14 | 1e-10 | 8.0e-13 | 2.8e-12 |

15 line searches per run; both routes bit-reproducible in every case.

- The gap tracks the ODE tolerance one for one and is unaffected by the inner tolerance. It is the objective's own derivative noise, not cancellation: normalising by `sum(abs, g .* s)` gives the same figure, and roundoff in a dot product of O(1) terms would sit at 1e-16.
- The 20% relative figure in #1284 is that same 1e-8 absolute noise divided by a slope that has become tiny at convergence. No choice of route changes it.
- Every sample the line search takes at any other `alpha`, however small, is a fresh solve with the same noise relative to `dphi_0`, whichever route produced `dphi_0`. Making the `alpha = 0` slope come from the JVP does not make the sampled slope function any more self-consistent; it only makes an exact re-evaluation at `alpha = 0` bit-identical to the stored value.
- That exact re-evaluation was the HagerZhang failure in JuliaNLSolvers/LineSearches.jl#211, already fixed in LineSearches 7.8.1 (#212) before #1285 merged, and JuliaNLSolvers/LineSearches.jl#213 stops HagerZhang from evaluating at an endpoint at all by checking the secant point before evaluating it.

## Cost

On a 30-subject FOCE fit in Pumas with the default `BFGS(linesearch = HagerZhang(rho = 2.0), initial_stepnorm = 1.0)`, #1285 raises the JVP count over 10 iterations from 11 to 21, one extra full-population pushforward per iteration, and the fit time by about 15%. The gradient-only cost is unchanged.

## Test

`test/multivariate/linesearch_handoff.jl` gains a testset with an `InplaceObjective(; fdf, fjvp)` whose `fjvp` counts its calls. A recording line search checks that `dphi_0 == dot(NLSolversBase.gradient(df), s)` and, after `update_state!`, that `fjvp` was called zero times.


devmotion:
> explanation is wrong

That explanation is not wrong. The difference between both methods can indeed be arbitrarily large - eg due to cancellations - even if both gradient and jvp are implemented in the numerically best way. This does not require any ODEs or complex operations. That the problem doesn't show up in some example doesn't prove that the difference is generally negligible.

andreasnoack:
If the error was due to cancellation in the dot product then it wouldn't be reduced by a stricter ODE solver tolerance. Most likely, the difference is a reflection of the rounding error in either of the two derivative calculations. So when there is a 20% relative difference, it probably means that they are both roughly 20% off the true gradient in relative terms. Regardless, I don't think the relative error is the right quantitive close to a stationary point. The main point is that there is no reason to expect the errors at different step sizes should be related such that you'd get a better line search by using the jvp for all the derivatives compared to one where the derivative for `alpha = 0` is evaluated with the dot product. For each value of `alpha`, there will be a rounding error and I'll expect it to be mostly unsystematic across alpha values and choice of code path.

devmotion:
I'm not sure why one would focus on or single out ODEs here - the problem is not specific to ODEs at all. It's not about the ODE solver tolerances or eg complex linear algebra operations. It's simply about floating point errors that can be completely different depending on whether calculations are based on the JVP or the gradient, e.g. due to problematic floating point cancellations in the gradient dot product that do not show up in the JVP and vice versa. 

A concrete example:

$$
f(x) = \exp{(x_1 - 3 x_2)} + x_3
$$

has gradient

$$
\nabla f(x) = \begin{bmatrix}
 \exp{(x_1 - 3 x_2)} \\
 -3 \exp{(x_1 - 3 x_2)} \\
 1
 \end{bmatrix}
$$

and hence

$$
\nabla f(x)^T v = \exp{(x_1 - 3 x_2)} \left(v_1 - 3 v_2\right) + v_3
$$

Consider the following implementations of the JVP and the gradient:

```julia
julia> jvp(x, v) = exp(x[1] - 3 * x[2]) * (v[1] - 3 * v[2]) + v[3];

julia> g(x) = [exp(x[1] - 3 * x[2]), -3 * exp(x[1] - 3 * x[2]), 1];
```

Fix e.g.

```julia
julia> x = [200.0, 0.0, 0.0];
```

The gradient at `x` does not suffer from an accuracy problem:

```julia
julia> g(x) == Float64.(g(big.(x)))
true
```

However, if we consider `v = [z, 1.0, 1.0]` with `z` around 3, the JVP is accurate but the dot-product with the gradient can be completely off:

```julia
julia> v = [3.0, 1.0, 1.0];

julia> jvp(x, v)
1.0

julia> jvp(x, v) == Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
1.0
```

```julia
julia> v = [3.0 - 10 * eps(), 1.0, 1.0];

julia> jvp(x, v)
-1.6044884905421217e72

julia> jvp(x, v) ≈ Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
-1.7668470647783843e72
```

```julia
julia> v = [3.0 - 5 * eps(), 1.0, 1.0];

julia> jvp(x, v)
-6.417953962168487e71

julia> jvp(x, v) ≈ Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
-8.834235323891922e71
```

```julia
julia> v = [prevfloat(3.0), 1.0, 1.0];

julia> jvp(x, v)
-3.2089769810842434e71

julia> jvp(x, v) ≈ Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
-4.417117661945961e71
```

```julia
julia> v = [nextfloat(3.0), 1.0, 1.0];

julia> jvp(x, v)
3.2089769810842434e71

julia> jvp(x, v) ≈ Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
1.0
```

```julia
julia> v = [3.0 + 5 * eps(), 1.0, 1.0];

julia> jvp(x, v)
6.417953962168487e71

julia> jvp(x, v) ≈ Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
4.417117661945961e71
```

```julia
julia> v = [3.0 + 10 * eps(), 1.0, 1.0];

julia> jvp(x, v)
1.6044884905421217e72

julia> jvp(x, v) ≈ Float64(jvp(big.(x), big.(v)))
true

julia> dot(g(x), v)
1.3251352985837882e72
```

In this - arguably constructed - case it seems clear that it would make a difference and could break any linesearch algorithm if `alpha = 0` in the linesearch algorithm is evaluated using the gradient and subsequent non-zero `alpha`s are evaluated using the JVP.

andreasnoack:
> I'm not sure why one would focus on or single out ODEs here

Because it demonstrates the main source of the error here. It provides a nob to adjust the error source that we can then use to compare the two ways of computing the derivative. The error from computing the objective function and the gradient is then amplified by the Hessian approximation. The error of the dot product between the gradient and search direction is bounded by the condition number of the Hessian approximation. So the error is large when the Hessian approximation is poorly conditioned. However, the jvp most likely won't help you in the quasi-Newton setting because the direction is computed with the same Hessian approximation, so the direction itself has the error. The reason why your example works is that `v` is chosen freely.

devmotion:
I still don't understand - why should a change of the linesearch algorithm/Optim integration be motivated by a specific ODE-based function? That doesn't seem reasonable at all.

The whole purpose of the example is to show concretely that I'm not speculating or merely worried that JVP and gradient-based evaluations can differ basically arbitrarily even in simple cases under ideal conditions (no ODEs, no error tolerances to tweak, accurate JVPs and gradients). There's no way for Optim to know upfront and to prevent a user from working with a function where such a difference occurs and whether the current step is in a problematic region. The only thing Optim can - well, in my opinion must do - is to ensure that the linesearch algorithm does not sometimes uses one, sometimes the other approach. There's just no way to upper bound the difference upfront - the whole purpose of this example is to show it can be arbitrarily large and thereby break the linesearch completely.

test/multivariate/linesearch_handoff.jl
```diff
@@ -1,18 +1,19 @@
-# `dphi_0` has to be the slope the line search samples itself. `fjvp` here is the gradient
-# version off by one ulp, so the two routes are distinguishable without either being wrong.
-@testset "dphi_0 comes from the objective's JVP" begin
+# The slope at alpha = 0 is `dot(g, s)` with the gradient `update_fgh!` already computed.
+# An objective that supplies its own `fjvp` must not be asked for a JVP there: that is a
+# full extra evaluation per iteration, and its result is not more accurate than the
+# gradient's, only differently rounded.
```
devmotion: That's not true, as the example shows.

```suggestion
# full extra evaluation per iteration.
```

src/utilities/perform_linesearch.jl
```diff
@@ -37,12 +37,11 @@ function reset_search_direction!(state::ConjugateGradientState, ::ConjugateGradi
 end
 
 function perform_linesearch!(state, method, d)
-    # The line search takes every later slope from `value_jvp!`, so take the one at
-    # alpha = 0 from the objective's JVP too.
-    dphi_0 = real(NLSolversBase.jvp!(d, state.x, state.s))
+    # Calculate search direction dphi0
```
devmotion: The variable is not actually called `dphi0` and the name seems a bit irrelevant for the comment anyways:

```suggestion
    # Calculate search direction
```

devmotion (commented):
As the example shows, the error of the JVP and the gradient approach are not always on the same order of magnitude - and hence it *might* be a problem sometimes. Since it's not completely clear *when* and *how often* it's a problem though, let us make this change for now and reconsider and debug if it pops up again.

pkofod:
Fwiw, the argument about choosing v freely is put into words as the following by claude which matches our discussion earlier I think

> The example needs a freely chosen v. In the line search, s comes from the same gradient through the Hessian approximation, and dphi_0 = -g'Hg. To get his level of cancellation there, |g'Hg| has to sit at roughly eps * sum(abs, g .* s), which means the slope is at the noise floor of the gradient itself. In that regime every sampled slope at alpha > 0 is equally meaningless on either route, so the JVP route does not save the line search.

pkofod (approved):

PR #1287, opened 2026-09-14
Bound how far a trust-region step may fall outside the radius

pkofod:
`solve_tr_subproblem!` can stop before the boundary root is found, either because `max_iters` is reached or because the safeguards stagnate at a bound. It records that in `reached_solution`, but `update_state!` never reads the flag and takes the step regardless.

## What is and is not a problem

Landing a little outside the radius is not a defect. The radius is not a constraint of the user's problem; it is the method's own running estimate of how far the quadratic model can be believed, and `rho` retests that estimate at every step. Inexact sub-problem solvers depend on the slack: Moré–Sorensen's `kappa_easy` rule accepts any step within a relative tolerance of the boundary, and NLSolvers' `NTR` ships that tolerance at 1/10.

Measured over the unconstrained `OptimizationProblems.jl` set, by re-solving every sub-problem along the trajectories from an extended trace (`solve_tr_subproblem!` is deterministic, so this reproduces exactly what each iteration solved, with no changes to Optim):

| | |
|---|---|
| sub-problems solved | 9334 |
| returning a step outside the radius | 718 (7.69%) |
| **median excess** | **5.8e-05** |
| exceeding twice the radius | 42 |
| exceeding a million times it | 35 |
| worst | **2.69e23x** (`gulf`, on 139 of its 491 sub-problems) |

The distribution is the whole story: the bulk is noise and a thin tail is unbounded. At the tail, `lambda` is pinned near `lambda_lb`, `H + lambda*I` is numerically singular, and the step has nothing to do with the model the radius was chosen for.

So this bounds the excess rather than removing it, at the same 1/10 the inexact termination rules use.

## Why not just project every step back

Because it costs more than it buys. Of the 718, projecting all of them leaves 184 worse than the constrained optimum and drops 26 below the Cauchy decrease that global convergence rests on. The raw steps, by contrast, are never worse than the constrained optimum (they beat it precisely by being outside) and achieve Cauchy decrease 718/718.

## Choosing the tolerance

Swept on two independent problem sets, both with every setting varied inside a single process. Counts are exact and reproduce run to run.

`OptimizationProblems.jl`, 72 unconstrained problems, iterations over the 55 that converge under every setting:

| kappa | net iterations | better | worse | converged |
|---|---|---|---|---|
| 0 (clip everything) | +17 | 7 | 7 | 56 |
| 0.01 | −138 | 7 | 0 | **57** |
| 0.05 | −3 | 3 | 1 | 56 |
| **0.1** | **−4** | **2** | **0** | 56 |
| 0.25 | −1 | 1 | 0 | 56 |
| 1.0 | −1 | 1 | 0 | 56 |
| Inf (as shipped) | 0 | — | — | 56 |

Hand-written families to n = 3000 with several starting points, 180 cells:

| kappa | net iterations | better | worse | converged |
|---|---|---|---|---|
| 0 (clip everything) | +5 | 3 | 12 | 179 |
| 0.01 | +14 | 1 | 9 | 179 |
| **0.1** | **+7** | **1** | **6** | 179 |
| 0.25 | +1 | 0 | 1 | 179 |
| 1.0 | 0 | 0 | 0 | 179 |
| Inf (as shipped) | 0 | — | — | 179 |

`kappa = 0.01` is the best value on the first set and among the worst on the second, which is why both were run. At 1/10 the cost is 7 iterations across 180 cells, five of the six regressions in a single family, with no change in convergence or iteration-cap hits on either set. It clips 83 of the 9334 sub-problems.

## Two alternatives, measured and rejected

**Raising `max_iters` from 5.** It drives `rat43` into the iteration cap and costs 12% more iterations on the first set. The cap is hit on a fifth of all sub-problems and the trajectories depend on that inexactness, so 5 is a tuning constant rather than a bug. Supporting evidence that it is close to right: an independent implementation of the same safeguarded Newton iteration (NLSolvers' `NTR`) needs at most 6 Cholesky factorizations, never more, across 180 constructed sub-problems.

**Falling back to the Cauchy point** when the clipped step does not clear it. It restores a guarantee the clipping gives up, and across both sets it changes exactly one cell — for the worse.

Also tried: replacing the ridged-Cholesky root find with a bracketed root find in the eigenbasis, which is where the step is built from anyway and cannot return an unbounded step at all. It is exact and feasible everywhere on the conformance suite, and it costs 16% more iterations, more than raising `max_iters` does. Noted here because it is the structural fix and someone will think of it; the inexactness is load-bearing.

## Relation to #1277

The two compose and neither subsumes the other. #1277 stops failed factorizations from consuming root-finding iterations and leaves `max_iters` alone; this bounds what a step can be when the budget is exhausted anyway. One figure here is qualified by it: the same trace shows `reached_solution` false on 20.5% of sub-problems, measured on v2.3.1 and therefore without #1277, so some of that share is the mis-accounting #1277 fixes. The 7.69% is measured on the step norm directly and is unaffected.

## Tests

Full suite green on this branch rebased on master.


pkofod:
This is not exactly the right approach. The usual thought is to accept a step length that's within some fraction of the trust region radius (kappaeasy in NLSolvers for example)

pkofod:
#1288 makes this obsolete

PR #1288, opened 2026-09-17
Solve the trust region subproblem in the eigenbasis and do not recompute the eigen decomposition when shrinking the radius

devmotion:
I came across [SciML/NonlinearSolve.jl#1287](https://github.com/SciML/NonlinearSolve.jl/pull/1287) and wanted to check whether there was anything in it we could adopt to improve Optim.

Its headline ("use Moré–Sorensen") turned out not to be the transferable part — `solve_tr_subproblem!` has been a Moré–Sorensen solver all along. What it actually buys is a *cheap* λ step, so the safeguarded iteration can run to convergence, plus the safeguarding itself. Optim can do better than copy it: `eigen(Symmetric(H))` is already computed before the root finder starts, and in that basis both quantities the Newton step needs are closed-form, so a λ step is O(n) with no factorization at all.

### Solving the subproblem in the eigenbasis

With the step cheap, the iteration can afford to run. The bracket is now tightened on the sign of φ, an out-of-bracket iterate retreats by the same geometric mean `initial_safeguards` already uses, and the loop stops on the boundary condition rather than the λ increment. The step is formed once from the λ that is returned, so the pair actually solves `(H + λI)s = -gr`. The model value is evaluated in a form that cannot cancel and avoids the `eps*‖H‖*‖s‖²` error term.

Over 3894 random boundary subproblems, at each solver's own default budget:

|  | before | after |
|---|---|---|
| `\|‖s‖-δ\|/δ` (p99 / max) | 1.31 / 235 | 2.9e-07 / 7.5e-06 |
| KKT residual (p99 / max) | 0.68 / 34 | 5.0e-13 / 1.3e-12 |
| `reached_solution` | 2602/3894 | 3893/3894 |

### Reusing the eigenbasis across rejected steps

Adopted from [#1304](https://github.com/SciML/NonlinearSolve.jl/pull/1304), though not its code: upstream warm-starts λ because each λ trial is a factorization, which is exactly the cost the eigenbasis removes. What Optim was repeating is the decomposition itself. A rejected step restores `x`, `g` and `H` bit for bit and only shrinks `delta`, so the next solve is the same subproblem at a smaller radius, and it was paying a fresh `eigen` on a bitwise-identical matrix.

`TRSubproblemCache` holds the eigenbasis and the gradient in it. It is refreshed at the top of a step that follows an accepted one, immediately above the solve that consumes it, so a rejected step simply skips the decomposition. No refresh is wasted — one only happens when a next step begins, so the final accepted step never triggers one.

The arithmetic is untouched, and the iterate path with it: over every twice-differentiable stock problem plus rosenbrock at n = 2, 20, 100 and 200, the minimum, iteration count, minimizer and full `delta` trace are bitwise unchanged. Rosenbrock, derivatives supplied, min of 200 samples:

| | n=20 | n=100 |
|---|---|---|
| before | 211 µs | 6119 µs |
| after | 188 µs | 5357 µs |

Allocations at n = 20 / 100 / 200: 411 → 363, 5454 → 4915, 19171 → 18003 KiB.

### KrylovTrustRegion

Two defects share the same shape and are fixed here too: `cg_steihaug!` broke out on a near-zero `dHd`, which on the first iteration left the step at zero, and the resulting `m_diff == 0` slipped past an `@assert` to give `rho = 0/0`. NaN loses every comparison, so the radius never shrank and the solver span for the whole budget on a zero step — `f(x, y) = x*y` from `[1, 0]` reproduces it, and is now a test. Its `assess_convergence` also honours all five tolerances now, like `NewtonTrustRegion` since #1278.

The same defect class is [#1301](https://github.com/SciML/NonlinearSolve.jl/pull/1301) upstream. `NewtonTrustRegion` already survives it, because it branches the radius update on the accept decision rather than on `rho`, but nothing covered that — there is now a regression test.

Deliberately left out, as they reshape the iterate path on problems that already converge and belong in a separate opt-in change: Moré's step-following radius update and diagonal scaling `‖Ds‖ ≤ Δ`.


pkofod:
I'll review it, but I am also looking at some other improvements for the trust region solver already so I will have to check for overlap.

devmotion:
### Benchmarks for this PR

The suite had no second-order solvers at all, so the CI benchmark comment on this PR reports exclusively on code the PR does not touch. 160ec339 adds them:

* the three second-order solvers on the existing Gabor fit, and
* an extended Rosenbrock at `n = 2, 20, 100` with analytic gradient and Hessian, started at `repeat([-1.2, 1.0], n ÷ 2)` where the Hessian is indefinite — so the trust region solvers begin on the boundary and actually run the safeguarded λ iteration this PR rewrites.

**The CI comment above will not pick this up.** `pull_request_target` always evaluates the workflow from the base branch, and `benchpkg` resolves a *single* benchmark script and reuses it for every revision — which the action defaults to the default branch. So master's script is applied to both sides and a benchmark added in a PR never runs, not even in the PR column. 160ec339 also sets `bench-on: ${{ github.event.pull_request.head.sha }}`, which fixes that for PRs opened after it merges. Hence the local run below.

Rerun at 79a39466. The numbers previously in this comment covered 96c18646 and 160ec339 only — that is, before the eigenbasis is reused across rejected steps — and the comparison between the two runs is the subject of the second point below.

<details>
<summary><code>benchpkg Optim --rev=master,dirty --exeflags="-O3 --threads=auto" --script=benchmark/benchmarks.jl</code> (Julia 1.12.7, M-series macOS)</summary>

|                                                          | master             | this PR            | master / PR    |
|:---------------------------------------------------------|:------------------:|:------------------:|:--------------:|
| multivariate/problems/rosenbrock/2/Newton                | 21.4 ± 2 μs        | 20.9 ± 0.93 μs     | 1.02 ± 0.1     |
| multivariate/problems/rosenbrock/2/NewtonTrustRegion     | 28.7 ± 3.5 μs      | 22 ± 2.3 μs        | 1.3 ± 0.21     |
| multivariate/problems/rosenbrock/2/KrylovTrustRegion     | 0.145 ± 0.011 ms   | 16.2 ± 1.6 μs      | 8.95 ± 1.1 ⚠️  |
| multivariate/problems/rosenbrock/20/Newton               | 0.0537 ± 0.0055 ms | 0.0543 ± 0.0051 ms | 0.99 ± 0.14    |
| multivariate/problems/rosenbrock/20/NewtonTrustRegion    | 0.276 ± 0.023 ms   | 0.198 ± 0.012 ms   | 1.39 ± 0.14    |
| multivariate/problems/rosenbrock/20/KrylovTrustRegion    | 0.0323 ± 0.0047 ms | 0.032 ± 0.003 ms   | 1.01 ± 0.17    |
| multivariate/problems/rosenbrock/100/Newton              | 0.744 ± 0.059 ms   | 0.71 ± 0.021 ms    | 1.05 ± 0.088   |
| multivariate/problems/rosenbrock/100/NewtonTrustRegion   | 8.46 ± 0.84 ms     | 5.59 ± 0.35 ms     | 1.51 ± 0.18    |
| multivariate/problems/rosenbrock/100/KrylovTrustRegion   | 0.199 ± 0.017 ms   | 0.193 ± 0.016 ms   | 1.03 ± 0.12    |
| multivariate/solvers/second_order/Newton                 | 0.127 ± 0.0066 ms  | 0.128 ± 0.0058 ms  | 0.993 ± 0.069  |
| multivariate/solvers/second_order/NewtonTrustRegion      | 0.154 ± 0.0078 ms  | 0.153 ± 0.011 ms   | 1 ± 0.088      |
| multivariate/solvers/second_order/KrylovTrustRegion      | 0.207 ± 0.0099 ms  | 0.208 ± 0.0073 ms  | 0.994 ± 0.059  |
| multivariate/solvers/first_order/Adam                    | 0.452 ± 0.025 ms   | 0.431 ± 0.014 ms   | 1.05 ± 0.067   |
| multivariate/solvers/first_order/AdaMax                  | 0.435 ± 0.028 ms   | 0.434 ± 0.017 ms   | 1 ± 0.076      |
| multivariate/solvers/first_order/BFGS                    | 0.0826 ± 0.007 ms  | 0.0788 ± 0.0036 ms | 1.05 ± 0.1     |
| multivariate/solvers/first_order/LBFGS                   | 0.0889 ± 0.003 ms  | 0.0895 ± 0.0041 ms | 0.994 ± 0.056  |
| multivariate/solvers/first_order/NGMRES                  | 0.649 ± 0.038 ms   | 0.635 ± 0.022 ms   | 1.02 ± 0.069   |
| multivariate/solvers/first_order/ConjugateGradient       | 0.238 ± 0.013 ms   | 0.24 ± 0.012 ms    | 0.994 ± 0.074  |
| multivariate/solvers/first_order/GradientDescent         | 0.56 ± 0.022 ms    | 0.562 ± 0.02 ms    | 0.998 ± 0.053  |
| multivariate/solvers/first_order/MomentumGradientDescent | 1.11 ± 0.055 ms    | 1.1 ± 0.032 ms     | 1.01 ± 0.058   |

`time_to_load` is omitted: it reads 0.425 ± 0.043 s vs 0.498 ± 0.11 s, but that is `Pkg.add` vs `Pkg.develop`, not this PR. Timing `using Optim` five times in three scratch environments — master via `add`, master via `develop`, this PR via `develop` — gives 0.46–0.51 s in all three, with no separation between them. CI compares two git revisions and will not show this.

</details>

#### Reading the numbers

**Calibration first.** The eight first-order solvers are untouched by this PR, and all eight land in 0.99–1.05, straddling 1. That band is this run's noise floor, so treat anything inside it as no change. (The run previously reported here sat at 0.95–0.99 instead — a systematic offset that has to be subtracted before its ratios can be compared with these.)

**`NewtonTrustRegion` is now measurably faster: 1.30 / 1.39 / 1.51 at n = 2 / 20 / 100.** That is 79a39466, and it is well clear of the calibration band at all three sizes. The eigenbasis rewrite on its own (96c18646) was meant to be more robust rather than faster and measured that way — the same three rows previously read 0.968 / 1.12 / 1.04 against a 0.95–0.99 baseline, i.e. roughly 1.01 / 1.16 / 1.08 once that offset is removed. The win arrives when the basis stops being recomputed: a rejected step restores `x`, `g` and `H` bit for bit and only shrinks delta, so the `eigen(Symmetric(H))` it used to redo was on a bitwise-identical matrix, and after 96c18646 that decomposition is the entire cost of the solve. The gain grows with n, as an O(n³) decomposition skipped should.

**The Gabor rows stay at 1.00**, and that is the same effect seen from the other side rather than a contradiction. That fit is n = 2 with a 64-term loss, so an iteration is dominated by the objective and there is nothing to save on a 2×2 `eigen`. Rosenbrock at n = 2 is the same matrix size with a trivially cheap objective, and it shows 1.30.

**The n=2 `KrylovTrustRegion` row is not a speedup.** ⚠️ It compares a stuck run against a converged one, so the 8.95 is meaningless as a timing ratio. Running each benchmark to completion under a 1000-iteration budget and checking `Optim.converged`:

| | master | this PR |
|---|---|---|
| n=2 `Newton` | ✅ 24 iters | ✅ 24 iters |
| n=2 `NewtonTrustRegion` | ✅ 26 iters | ✅ 26 iters |
| **n=2 `KrylovTrustRegion`** | ❌ **hits the 1000-iteration cap, minimum 1.0e-15** | ✅ **26 iters, minimum 1.2e-28** |
| n=20 and n=100, all three | ✅ 24–27 iters | ✅ 24–27 iters |

That is the `cg_steihaug!` fix, reproduced from a standard starting point on a textbook problem rather than a constructed one. At `[-1.2, 1.0]` the Hessian is indefinite and the first CG direction can hit non-positive curvature; on master that path could return a zero step, making `rho = (f_old - f_new) / (m_old - m_new)` a `0/0` NaN. A NaN `rho` fails every radius-update comparison, the radius freezes, and the solver spins out its entire budget.

It only shows at n=2 — at n=20 and n=100 the initial direction has enough room that the degenerate case does not trigger, which is why the three sizes are worth keeping.

Every other row is converged-vs-converged and is a fair comparison. The iteration counts are unchanged between master and this PR everywhere except that n=2 `KrylovTrustRegion` row, so the `NewtonTrustRegion` speedups above are per-iteration, not a shorter run. The Gabor rows are identical on both revisions (`Newton` 7 iters → 27.802, `NewtonTrustRegion` 10 → 24.944, `KrylovTrustRegion` 8 → 24.944; `Newton` converges to a different, worse local minimum, which is legitimate on a nonconvex fit).

pkofod:
kind of confusing that the benchmark table has different units for time...

